PR #139, opened 2017-05-22
LCM CMake-based build script

ibaned:
@lxmota


ibaned:
Two requests from @lxmota:

 - [ ] build directory names should not be capitalized
 - [ ] output should go to a log file, named by the old `$PREFIX`

lxmota:
@ibaned I have been testing this branch and committing a few changes here and there.

I notice that with the old scripts, the tests in release mode with both gcc and clang in our Fedora systems take around 770 seconds to complete, whereas with the new cmake-based scripts they take around 4500 seconds.

Any thoughts?

ibaned:
The two most likely causes I can think of are:
1. The `-j` flag is being passed to `ctest` in one but not the other. I looked at the old one and didn't see a `-j` flag, so maybe thats not it.
2. The optimization flags are not being passed to the compiler. I'm not in the office, but will try to confirm whether they get passed correctly when I can. The new script output should at some point mention `CMAKE_CXX_FLAGS` and `CMAKE_C_FLAGS` in its output.

ibaned:
Hmm... it is possible that commit https://github.com/gahansen/Albany/commit/ca05a421fd1f06b058a5eb5b83b1879beaaf1cda caused a regression... the `BUILD_STRING` environment variable is supposed to be used, it is set by the modules, for example here:
https://github.com/gahansen/Albany/blob/lcm-cdash/doc/LCM/modulefiles/lcm/release#L6

If the changes to the `.cmake` files from ca05a42 are reverted, perhaps the situation will improve. The output of the script should be checked to confirm that `CMAKE_BUILD_TYPE` is being set properly.

lxmota:
I see. I wasn't working for me for some reason. In fact I got some configuration errors. I'll double check.

lxmota:
As for the -j flag for the tests, my understanding is that tests manually set the number of processors that they want to run on. Other than that the tests run in serial.

In fact, as some point in the past I tried to run the tests by setting the -j flag manually for the tests and this resulted in a significant slow down. So in the old scripts I ended not passing the -j flag for the test phase.

Maybe that is happening here?

ibaned:
I'm pretty sure the new scripts do not pass `-j`. You can always use `top` to confirm the number of processes is reasonable while `ctest` is running. ca05a42 still looks like the problem to me. The change to `build.sh` is okay, but not the others.

lxmota:
Ok, I think I fixed the problem that I introduced.

I'm also trying to disable the serial mesh tests by detecting whether  TPL_Netcdf_PARALLEL is set to ON or OFF in the Trilinos configuration, but I'm having a hard time. I'm not able to find a way to read that variable in Albany's CMakeLists.txt. And I'm trying to detect whether Netcdf is parallel by searching the NEtcdf headers themselves, but that also is proving hugely complicated.

And I still have to see why the tests are slow.

ibaned:
@lxmota 

1. I have an idea for detecting parallel NetCDF, see trilinos/Trilinos#1359

2. I am running tests on `algol` and I think the reason it is slower is that some tests are timing out:
   ```
   DomainTear2D_CopyRemesh ......................................***Timeout 1500.02 sec
   ```
   probably what is different between the scripts is that the old scripts set a smaller timeout, and the new ones do not.

ibaned:
Just for book-keeping, #140 is tracking the NetCDF parallel problem.

lxmota:
@ibaned I think I'm about ready to merge this branch. I'm doing some more testing and I need to update the LCM install doc.

I notice that you added an 'update' option to the mix, besides 'clean', 'config', etc. What does it do?

Anything else that would be worth mentioning in the install doc?

ibaned:
@lxmota sound good.

`update` will execute `git pull` in the package repository, and if combined with `dash` it will send a report of changed files to CDash. This can be a useful feature when the script is running as a `cron` job for nightly tests.

lxmota:
It appears that there is something wrong with trying to push the ctest results to the cdash site. This is what I get:

```
SetCTestConfiguration:BuildDirectory:/home/amota/LCM/albany-build-serial-clang-release
SetCTestConfiguration:SourceDirectory:/home/amota/LCM/Albany/doc/LCM/build
SetCTestConfiguration:ProjectName:Albany
SetCTestConfiguration:DropMethod:http
SetCTestConfiguration:DropSite:cdash.sandia.gov
SetCTestConfiguration:DropLocation:/CDash-2-3-0/submit.php?project=Albany
SetCTestConfiguration:IsCDash:TRUE
   Use HTTP Proxy: http://wwwproxy.ca.sandia.gov:80
Submit files (using http)
   Using HTTP submit method
   Drop site:http://cdash.sandia.gov/CDash-2-3-0/submit.php?project=Albany
   Upload file: /home/amota/LCM/albany-build-serial-clang-release/Testing/20170524-0600/Test.xml to http://cdash.sandia.gov/CDash-2-3-0/submit.php?project=Albany&FileName=algol.ca.sandia.gov___algol.ca.sandia.gov___20170524-0600-Nightly___XML___Test.xml&MD5=040ec67a7ea599ca837b62c31755c965 Size: 2847170
   Submit failed, waiting 10 seconds...
   Retry submission: Attempt 1 of 3
   Submit failed, waiting 10 seconds...
   Retry submission: Attempt 2 of 3
   Submit failed, waiting 10 seconds...
   Retry submission: Attempt 3 of 3
   Error when uploading file: /home/amota/LCM/albany-build-serial-clang-release/Testing/20170524-0600/Test.xml
   Error message was: The requested URL returned error: 417 Expectation Failed
   Problems when submitting via HTTP
CMake Warning at Albany/doc/LCM/build/snl_helpers.cmake:24 (message):
  Cannot submit albany_Fedora_algol.ca.sandia.gov_serial-clang-release Test
  results!
Call Stack (most recent call first):
  Albany/doc/LCM/build/snl_helpers.cmake:124 (snl_submit)
  Albany/doc/LCM/build/snl_helpers.cmake:249 (snl_test)
  Albany/doc/LCM/build/lcm_do_albany.cmake:64 (snl_do_subproject)
  Albany/doc/LCM/build/lcm_do_package.cmake:47 (lcm_do_albany)
  Albany/doc/LCM/build/lcm_build.cmake:67 (lcm_do_package)


CTEST_DROP_METHOD "http"
lcm_do_package returned "0"

```

Maybe my machine does not have permission? Everything else seems to be working fine.

ibaned:
I would try without the proxy

ibaned:
Also, I've only submitted from CEE, not `algol`

PR #169, opened 2017-08-10
Fix LCM modules for Skybridge

ibaned:
@lxmota @jwfoulk 

This pull request refactors the modules so that they work on Skybridge.

Important points:

1. All the `lcm/foo` modules have been renamed to `lcm-foo` to workaround Lmod's assumption that everything has a version number. I applied this change to both sets of modules for consistency

2. Lmod doesn't let people change `MODULEPATH` directly, so on Skybridge your setup has to be a bit different:
   ```bash
   export LCM_DIR=$PWD
   module use --append $LCM_DIR/Albany/doc/LCM/modulefiles #<- Lmod specific
   module load lcm-cluster
   module load serial-intel-release
   ```
   Using this, I was able to compile Trilinos and Albany on Skybridge.

3. I have *not* tested that the refactored old modules still work on SEMS or Fedora


lxmota:
@ibaned Thank you so much! I'm testing right now on SEMS and Fedora just to make sure that there are no glitches.

If everything looks good, I'll merge the pull request and announce that we now have modules for the clusters. This is really helpful to our users, thanks.

PR #203, opened 2017-10-27
Replace StrongDBC algorithm with one that scales

ibaned:
This PR changes the way StrongDBC functions.
The old code had a runtime on each MPI rank with was `O((#global DOFs)^2)` and used at least `O(#global DOFs)` memory.
The new code has a runtime on each MPI rank which is `O(#local nonzeros)`.


ibaned:
@lxmota @ikalash

ibaned:
@bgranzow

Issue #204, opened 2017-11-01
Remove CUbit-TRilinos CUTR capability


agsalin:
The capability to do shape optimization via library-mode Cubit has been lost over time. We can remove the associated code and tests, which have been walled off with ifdef ALBANY_CUTR.

ibaned:
I think you accomplished this? Also the nightlies look okay. Closing, please reopen if theres anything else to this.

PR #205, opened 2017-11-01
Remove CUTR

agsalin:
Addresses issue #204 

Remove Cubit Trilinos code for shape optimization. Has been dead code behind ifdefs for a while.


ibaned:
@mperego hasn't commented but we should get this in today so we can see what the nightly tests have to say about it. I'll go ahead and merge.

Issue #209, opened 2017-11-02
Subfolders in evaluators


bartgol:
The folder 'evaluator' is getting big, and with a too large font size listing its content takes more than one screen. Besides, there are evaluators that do conceptually different things. So I'm thinking we should probably create subfolders in it. I came to that conclusion by looking at the evaluators folder in LCM.

My suggestion is to add these subfolders:

- evaluators/interpolation (PHAL_*Interpolation*)
- evaluators/bc (Dirichlet*, Neumann)
- evaluators/load_save (for states)
- evaluators/gather_scatter (for solution/residual/params/egiendata,coords)
- evaluators/response (all evaluators with 'Response' in it)
- evaluators/pdes (CahnHIll*, NS*, Heat*,...)
- evaluators/utils (AddNoise,Field2Norm, ConvertFieldType,DOFCell2Side,Constant...)

Better names are always welcome. Perhaps the load_save and gather_scatter could be merged into 'inout'. Or we could differentiate load/gather that are from the mesh from those that are from Tpetra vectors (and same for save/scatter). This would put gathering coordinates in the same folder of loading states.

Thoughts?

lxmota:
I'm all for it. But then again, I'm biased because that is what LCM does.

mperego:
I also like it.

ikalash:
Likewise.

bartgol:
Any suggestion/comments on how to regroup the evaluators and/or the subfolders names? Otherwise I can go ahead and open a PR, so everyone can see how it would look like...

lxmota:
Your proposal sounds good to me. So if you make a PR I'll take a look at it.

ikalash:
I think this issue can be closed now.  Any objections?

bartgol:
Agreed. Closed.

PR #212, opened 2017-11-06
Evaluators subfolders

bartgol:
This PR addresses issue #209 .

Creating subfolders in evaluators and moving all evaluators to them.

The subfolders are tentative, and try to group the evaluators by their functionality. Currently, there are 8 subfolders:
    
- bc: dirichlet and neumann bc handling
- gather: gathering data (solution, coordinates, etc)
- interpolation: all the interpolation evaluators
- pde: evaluators specific to some type of PDE (or ODE)
- response: response evaluators (except for scatter)
- scatter: scattering of residual, parameter or responses
- stat: load/save states
- utility: general purpose utility evaluators that do not fit in any of the above

Suggestions/modifications to improve the names verbosity and/or the number of categories (subfolders) and/or the placement of the evaluators is more than welcome. There are a few evaluators that I was debated on where to put. For instance, the scatter response ones, could have gone in either scatter or response, but I opted for scatter. Similarly, IdentityCoodrinateFunctionTraits, which is in bc, could have been placed in utility.

Please, take a look and see if some is misplaced or not optimally placed.

On a side note, the reorganization made me notice some code duplication, and we may think about removing something. Namely, SeparableScatterScalarResponse and SeparableScatterScalarResponseT seem very very similar, since they both use Tpetra, so I guess we could eliminate one. QCAD_ResponseSaveField also seems to overlap with the standard SaveStateField, although it adds a few extra saving options (like magnitude for vectors), but perhaps could be merged into a single evaluator.

Finally, I would argue for a promotion of Adapt_, QCAD_, PNP_ evaluators to PHAL_ evaluators, since they are in the main PHAL folders. Otherwise, I would suggest to move them in their proper module (at least QCAD_ and PNP_), to leave PHAL as clean as possible.


bartgol:
Note: I did not run the tests. However, since this is only a reorganization of files, the fact that the code compiles should be enough to guarantee nothing gets broken.

ibaned:
@bartgol could you fix the merge conflicts?

bartgol:
I'll fix it, yes.

bartgol:
Ok, I fixed the conflict. If you think it's ok, go ahead and merge.

ibaned:
Yes, I think I will.

lxmota:
I get a configuration error after this merge:

```
And CMake says...:
-- Configuring done
CMake Error at src/CMakeLists.txt:909 (add_library):
  Cannot find source file:

    evaluators/pde/PHAL_Constant.cpp

  Tried extensions .c .C .c++ .cc .cpp .cxx .m .M .mm .h .hh .h++ .hm .hpp
  .hxx .in .txx


CMake Error: CMake can not determine linker language for target: albanyPDEs
CMake Error: Cannot determine link language for target "albanyPDEs".

```

lxmota:
Looks like there are more. Another one missing:

`src/evaluators/pde/PHAL_LangevinNoiseTerm.cpp
`

bartgol:
My bad. I thought the conflict fix was innocuous, but should have tested better. I will redo this tomorrow and check that nothing breaks, then will resubmit the PR.

lxmota:
Sounds good. Not really a problem on my end. I just didn't want the dashboard to be all red tonight.

Issue #213, opened 2017-11-06
detect 64-bit mode based on Tpetra


ibaned:
We should just look at what `GO` types Tpetra has enabled and base our choice of `GO` on that. The rule can be, we go with the largest `GO` type that Tpetra has instantiated.

mhoemmen:
See discussion here: https://github.com/trilinos/Trilinos/issues/1959

Summary: It would make sense to decouple the app's "global key" type choice, from Tpetra's global index type choice.  That would avoid unnecessary Tpetra instantiations, while decoupling the app's global index decision from Tpetra's.  Zoltan already does this.

ibaned:
Hmm there is an argument for having them be closely coupled, which is that the main reason to use 64 bit globals is to solve problems with over 2 billion global unknowns, and that use case only works if all the relevant packages are using 64-bit `GO`. Enforcing consistency protects users and developers from compiling the code with a mismatching combination of `GO`s and seeing issues down the line resulting from overflows in narrowing conversions.

bartgol:
We could have our own GO type, which could default, say, to 64 bits. Then, check that Tpetra has indeed that type enabled, and, if not, error out at config time, asking to either rebuild Trilinos, or select one of the enabled GO types in Tpetra. This way we always enforce compatibility, while also keeping the two packages choices somewhat independent. We also increase transparency, the default GO type is always the same. Selecting the largest available in Tpetra has the disadvantage that, if one has two different builds of Tpetra on two different machines, the Albany build will be different, despite his/her Albany configure scripts being identical.

ibaned:
I think it should be well understood that if you link Albany to a differently-configured Trilinos, you will get different behavior even if the Albany CMake flags are the same. In fact I think thats a good thing, I prefer to specify options only once instead of having to repeat them in both Trilinos and Albany and code error checks to ensure they're the same.

mhoemmen:
@bartgol wrote:
> We could have our own GO type, which could default, say, to 64 bits. Then, check that Tpetra has indeed that type enabled, and, if not, error out at config time, asking to either rebuild Trilinos, or select one of the enabled GO types in Tpetra.

[...]

PR #219, opened 2017-11-08
First pass at making everything 64-bit

ibaned:
These are changes to bring us to always-64-bit mode.
Per issue #213, it separates Albany's `GO` type from `Tpetra_GO`, they just have to be both 64 bits.
I haven't compiled this yet, just putting it up for now.


src/Albany_DataTypes.hpp
```diff
@@ -23,12 +24,11 @@
...
 #include "Phalanx_KokkosDeviceTypes.hpp"
 
-//amb Need to move to configuration.
-//#define ALBANY_SFAD_SIZE 27
-//#define ALBANY_SLFAD_SIZE 27
+#include "TpetraCore_config.hpp"
 
-//#define ALBANY_FAST_FELIX
-// Typedef AD types to standard names
+#ifndef HAVE_TPETRA_INST_DOUBLE
```
bartgol: I would use HAVE_TPETRA_DOUBLE, in case someone does not use (?) ETI in Tpetra.

bartgol: Also for all the int types below.

src/Albany_DataTypes.hpp
```diff
@@ -59,36 +59,48 @@ typedef Sacado::Fad::DFad<RealType> TanFadType;
...
+typedef std::int32_t                                         LO;
+typedef Kokkos::Compat::KokkosDeviceWrapperNode<PHX::Device> KokkosNode;
+
+typedef int Tpetra_LO;
+#if defined( HAVE_TPETRA_INST_INT_LONG_LONG )
+typedef long long Tpetra_GO;
+#elif defined( HAVE_TPETRA_INST_INT_LONG )
+static_assert(sizeof(long) == sizeof(GO),
+    "Tpetra's biggest enabled GlobalOrdinal is long but thats not 64 bit");
+typedef long Tpetra_GO;
```
bartgol: Why bothering with two types names if we enforce them to have the same size? Can't we just have `GO`, so that people don't get confused? If we allowed Albany to be built with 32 bits global ordinals, then it would make sense (but I would check that `sizeof(Tpetra_GO)>=sizeof(GO)` then).

ibaned: Because `Tpetra_GO` is used as a template parameter to defined `Tpetra_Vector`, it has to be _exactly_ the type that Tpetra supports, not just one that is compatible.

bartgol: Oh, like `GO=long` and `Tpetra_GO=unsigned long`? But if we allow this, aren't we running into a lot of warning when, e.g., later on we compare a `GO` with a `Tpetra_GO`? Or, even worse, if we get an implicit conversion from one to the other when it is not a positive `GO`? I gave a quick look and we only convert a `GO` into a `Tpetra_GO`, which is fine if the `GO` is non-negative (and maybe even -1 is fine cause it may be caught by Tpetra as an Invalid ordinal). But aren't we walking on thin ice, with the risk of a bug sneaking in at any time? Can't we just ask for the same type, not just same size at this point? I'm really not sure, maybe I'm just overreacting.

ibaned: See @mhoemmen 's comments in #213 where he advocates that we not require exactly the same type. Yes, we should use appropriate conversions, and safe ones if possible. Note that ordinals are not actually supposed to ever be negative, there are other reasons for using signed integers even if most of the time stored values are positive. Also, this is is something we'll have to do for DTK anyways. We need to get into the habit of knowing which type each package uses and converting as needed. Its actually more painful (in fact impossible, e.g. DTK) to constantly reconfigure packages until they all have exactly the same types.

ibaned: I'm actually expecting that our most common conversion will be `long` <-> `long long`, which are the same for all practical purposes on Linux but technically different per the compiler. I think `int64_t` is `long`, and Tpetra usually goes with `long long`.

src/Albany_DataTypes.hpp
mhoemmen: `HAVE_TPETRA_INST_DOUBLE` is actually independent of ETI.  Tpetra "enables" a set of Scalar types, regardless of whether ETI is enabled.  It is incorrect to use Scalar (GlobalOrdinal, etc.) types as template parameters if they have not been enabled.

src/Albany_DataTypes.hpp
mhoemmen: I've seen cases where `int64_t` is `long`, and other cases where it is `long long`.  (Super annoying!  I really wish `int64_t` were a proper type instead of an alias (typedef).)

mhoemmen: Albany could consider using a "tagged integer" type for `GO`.  This would prevent arithmetic that implicitly mixes global and local indices, for example.

ibaned: > Albany could consider using a "tagged integer" type for `GO`.

Interesting idea... can you point to an implementation / example / description of these things so we can consider it?

src/Albany_DataTypes.hpp
bartgol: So what's the difference between HAVE_TPETRA_DOUBLE and HAVE_TPETRA_INST_DOUBLE?

src/Albany_DataTypes.hpp
mhoemmen: @ibaned Sierra's Aria product uses this extensively.  We've wanted this for years in Tpetra, but it would break backwards compatibility.  If an app decouples its global index type from Tpetra, it's free to add this feature.

ibaned: some quick googling turned [this](http://www.ilikebigbits.com/blog/2014/5/6/type-safe-identifiers-in-c) up... seems pretty compelling... I can think of possible unwanted side effects, but it may be worth a try.

mhoemmen: @ibaned Yup!  I don't mean to tell Albany that they should do this, just that if Albany decouples its global index type from Tpetra's, then Albany has the opportunity to consider things like this.

bartgol: [this](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2013/n3515.pdf) would make all of this much easier...

ibaned: @bartgol yes, tagged integers seem to be a template trick to basically implement opaque typedefs, as much as possible in the current language.

src/Albany_DataTypes.hpp
mhoemmen: @bartgol None :-)

bartgol: Ah, ok. Good to know. Any plan to remove one of the two and avoid the confusion in the future? :-)

mhoemmen: @bartgol Just saw your question now for some reason.  The answer is "possibly yes."  Note that "instantiation" does not necessary imply "explicit instantiation," though :-) .

ibaned:
@ikalash and I have pushed several fixes to this branch, and I just merged master into it to bring in @bartgol 's changes.

@lxmota would you like to try this branch out before we merge it?

ikalash:
Thanks for letting me know, @ibaned .  The CUDA nightly build uses the always-64-bit branch so it'll get tested tonight regardless of whether it gets merged into master.  

Once this branch gets merged into master, I should probably consolidate my nightlies to just have one (64-bit) build rather than separate 64-bit and 32-bit builds (right?).  Also I think I'll need to change my Trilinos configure to set: 

 "-DTpetra_INST_INT_LONG_LONG:BOOL=ON"
 "-DTpetra_INST_INT_INT:BOOL=OFF"

lxmota:
I get compile errors in this branch. I attach the compile log.

[albany-serial-gcc-release.log](https://github.com/gahansen/Albany/files/1515626/albany-serial-gcc-release.log)

lxmota:
Condensed version of the above (just the errors):

```
/home/amota/LCM/Albany/src/Petra_Converters_64.cpp:28:80: error: conversion from ‘Teuchos::ArrayView<const long long int>’ to non-scalar type ‘const Teuchos::ArrayView<const long int>’ requested
/home/amota/LCM/Albany/src/Petra_Converters_64.cpp:53:97: error: no matching function for call to ‘Tpetra::Map<int, long long int, Kokkos::Compat::KokkosDeviceWrapperNode<Kokkos::Serial> >::Map(const global_size_t&, Teuchos::Array<long int>&, const GO&, const Teuchos::RCP<const Teuchos::Comm<int> >&, const Teuchos::RCP<Kokkos::Compat::KokkosDeviceWrapperNode<Kokkos::Serial> >&)’
/home/amota/LCM/Albany/src/Petra_Converters_64.cpp:120:64: error: no matching function for call to ‘Tpetra::CrsMatrix<double, int, long long int, Kokkos::Compat::KokkosDeviceWrapperNode<Kokkos::Serial> >::getGlobalRowView(GO&, Teuchos::ArrayView<const long int>&, Teuchos::ArrayView<const double>&)’
/home/amota/LCM/Albany/src/Petra_Converters_64.cpp:121:70: error: ‘int Epetra_CrsMatrix::ExtractGlobalRowView(int_type, int&, double*&) const [with int_type = long int]’ is private within this context
/home/amota/LCM/Albany/src/Petra_Converters_64.cpp:238:127: error: no matching function for call to ‘createNonContigMapWithNode<LO, Tpetra_GO, KokkosNode>(Teuchos::Array<long int>&, const Teuchos::RCP<const Teuchos::Comm<int> >&, const Teuchos::RCP<Kokkos::Compat::KokkosDeviceWrapperNode<Kokkos::Serial> >&)’
/home/amota/LCM/Albany/src/Petra_Converters_64.cpp:257:133: error: no matching function for call to ‘createNonContigMapWithNode<LO, Tpetra_GO, KokkosNode>(Teuchos::Array<long int>&, const Teuchos::RCP<const Teuchos::Comm<int> >&, const Teuchos::RCP<Kokkos::Compat::KokkosDeviceWrapperNode<Kokkos::Serial> >&)’
/home/amota/LCM/Albany/src/Petra_Converters_64.cpp:280:127: error: no matching function for call to ‘createNonContigMapWithNode<LO, Tpetra_GO, KokkosNode>(Teuchos::Array<long int>&, const Teuchos::RCP<const Teuchos::Comm<int> >&, const Teuchos::RCP<Kokkos::Compat::KokkosDeviceWrapperNode<Kokkos::Serial> >&)’
/home/amota/LCM/Albany/src/Petra_Converters_64.cpp:300:135: error: no matching function for call to ‘createNonContigMapWithNode<LO, Tpetra_GO, KokkosNode>(Teuchos::Array<long int>&, const Teuchos::RCP<const Teuchos::Comm<int> >&, const Teuchos::RCP<Kokkos::Compat::KokkosDeviceWrapperNode<Kokkos::Serial> >&)’
/home/amota/LCM/Albany/src/Petra_Converters_64.cpp:314:63: error: no matching function for call to ‘Tpetra::CrsGraph<int, long long int, Kokkos::Compat::KokkosDeviceWrapperNode<Kokkos::Serial> >::insertGlobalIndices(GO&, Teuchos::Array<long int>&)’
/home/amota/LCM/Albany/src/disc/stk/Albany_GenericSTKFieldContainer_Def.hpp:291:49: error: call of overloaded ‘LID(const long unsigned int&)’ is ambiguous
/home/amota/LCM/Albany/src/responses/Albany_SolutionResponseFunction.cpp:77:52: error: no matching function for call to ‘Tpetra::CrsGraph<int, long long int, Kokkos::Compat::KokkosDeviceWrapperNode<Kokkos::Serial> >::insertGlobalIndices(GO&, Teuchos::ArrayView<long int>&)’
/home/amota/LCM/Albany/src/responses/Albany_SolutionResponseFunction.cpp:365:65: error: conversion from ‘Teuchos::ArrayView<const long long int>’ to non-scalar type ‘Teuchos::ArrayView<const long int>’ requested
/home/amota/LCM/Albany/src/responses/Albany_SolutionResponseFunction.cpp:374:154: error: no matching function for call to ‘createNonContigMapWithNode<LO, Tpetra_GO, KokkosNode>(Teuchos::Array<long int>&, Teuchos::RCP<const Teuchos::Comm<int> >, Teuchos::RCP<Kokkos::Compat::KokkosDeviceWrapperNode<Kokkos::Serial> >)’
g++: error: CMakeFiles/albanySTK.dir/Albany_GenericSTKFieldContainer.cpp.o: No such file or directory
g++: error: CMakeFiles/albanyLib.dir/Petra_Converters_64.cpp.o: No such file or directory
g++: error: CMakeFiles/albanyLib.dir/responses/Albany_SolutionResponseFunction.cpp.o: No such file or directory
Scanning dependencies of target DTK_Interp_and_Error
[ 88%] Building CXX object src/LCM/CMakeFiles/DTK_Interp_Volume_to_NS.dir/utils/dtk_interp_and_error/interpolation_volume_to_ns.cpp.o
gmake[2]: *** No rule to make target 'src/libalbanyLib.so', needed by 'src/LCM/DTK_Interp_and_Error'.  Stop.
[ 88%] Building CXX object src/LCM/CMakeFiles/DTK_Interp_and_Error.dir/utils/dtk_interp_and_error/dtk_interp_and_error.cpp.o
[ 99%] Built target DTK_Interp_and_Error
   39 Compiler errors

```

ibaned:
@lxmota thanks for checking! I'll try to fix these soon.

ibaned:
@lxmota I just tried to fix these. Let me know if you get a chance to try it again.

lxmota:
Still some errors (and a warning) left:

```
In file included from /home/amota/LCM/Albany/src/disc/stk/Albany_TmplSTKMeshStruct.cpp:7:0:
/home/amota/LCM/Albany/src/disc/stk/Albany_TmplSTKMeshStruct_Def.hpp:28:16: warning: invalid suffix on literal; C++11 requires a space between literal and string macro [-Wliteral-suffix]
 #define ST_LLU "%"PRId64

/home/amota/LCM/Albany/src/responses/Albany_SolutionResponseFunction.cpp: In member function ‘virtual void Albany::SolutionResponseFunction::setupT()’:
/home/amota/LCM/Albany/src/responses/Albany_SolutionResponseFunction.cpp:76:39: error: no match for ‘operator=’ (operand types are ‘Teuchos::ArrayView<long long int>’ and ‘Teuchos::ArrayView<long int>’)
     rowAV = Teuchos::arrayView(&row, 1);
                                       ^
In file included from /home/amota/LCM/trilinos-install-serial-gcc-release/include/Teuchos_ArrayRCP.hpp:47:0,
                 from /home/amota/LCM/trilinos-install-serial-gcc-release/include/Teuchos_Array.hpp:52,
                 from /home/amota/LCM/Albany/src/responses/Albany_AbstractResponseFunction.hpp:12,
                 from /home/amota/LCM/Albany/src/responses/Albany_DistributedResponseFunction.hpp:12,
                 from /home/amota/LCM/Albany/src/responses/Albany_SolutionResponseFunction.hpp:12,
                 from /home/amota/LCM/Albany/src/responses/Albany_SolutionResponseFunction.cpp:7:
/home/amota/LCM/trilinos-install-serial-gcc-release/include/Teuchos_ArrayView.hpp:158:15: note: candidate: Teuchos::ArrayView<T>& Teuchos::ArrayView<T>::operator=(const Teuchos::ArrayView<T>&) [with T = long long int]
 ArrayView<T>& ArrayView<T>::operator=(const ArrayView<T>& array)
               ^~~~~~~~~~~~
/home/amota/LCM/trilinos-install-serial-gcc-release/include/Teuchos_ArrayView.hpp:158:15: note:   no known conversion for argument 1 from ‘Teuchos::ArrayView<long int>’ to ‘const Teuchos::ArrayView<long long int>&’

```

ibaned:
I think I was able to fix these as well...

PR #228, opened 2017-11-28
Evaluators subfolders

bartgol:
Re-creating the PR, after it got merged and reverted due to some errors. This addresses issue #209


bartgol:
I'm testing the branch, but I get some compilation errors with ATO enabled. I am disabling ATO, since I get the same errors in master. Is ATO no longer supported?

ibaned:
@bartgol for future reference, the trick is to "revert the revert", i.e. do `git revert <first-revert-commit>`, and then add commits with fixes.

ikalash:
@ibaned , @bartgol : ATO should work unless you are building with CUDA.  I have ATO turned on in all my nightlies running on my workstation.  What are the errors you get in ATO, @bartgol ?

bartgol:
@ikalash Here is the error

`/home/lbertag/workdir/albany/albany-copy-src/src/ATO/evaluators/ATO_TensorPNormResponse_Def.hpp: In instantiation of ‘void ATO::TensorPNormResponse<EvalT, Traits>::evaluateFields(typename Traits::EvalData) [with EvalT = PHAL::AlbanyTraits::Tangent; Traits = PHAL::AlbanyTraits; typename Traits::EvalData = PHAL::Workset&]’:
/home/lbertag/workdir/albany/albany-copy-src/src/ATO/evaluators/ATO_TensorPNormResponse.cpp:15:1:   required from here
`/home/lbertag/workdir/albany/albany-copy-src/src/ATO/evaluators/ATO_TensorPNormResponse_Def.hpp:326:19: error: no match for ‘operator+=’ (operand types are ‘RealType {aka double}’ and ‘PHX::MDFieldTypeTraits<Kokkos::View<const Sacado::Fad::DFad<double>**, Kokkos::OpenMP> >::return_type {aka Sacado::Fad::ViewFad<const double, 0u, 1u, Sacado::Fad::DFad<double> >}’)`

I think it is because I am building with

`-D ENABLE_MESH_DEPENDS_ON_PARAMETERS:BOOL=ON`

I think ATO has a few lines where it uses a `RealType` where it should use a `MeshScalarT` instead. It all works fine in general, but when the mesh depends on parameters, it fails. Perhaps it's no big deal for ATO, but do you think it's better to be consistent and use MeshScalarT for mesh quantities that depend on mesh, in case in the future someone starts using ATO for meshes that depend on parameters/solution (I don't really know what ATO does, so I don't even know if this will/could ever happen).

mperego:
@bartgol Oh yeah. `MESH_DEPENDS_ON_PARAMETERS` works only for the simple PDE problems and for FELIX.

bartgol:
Btw, also LCM gives erros when `ENABLE_MESH_DEPENDS_ON_PARAMETERS` is on, since it tries to pass a mesh-dependent quantity to some AAdapt routine that expects a scalar field, and instead a Fad type is passed.

I'm not sure if it is easy/correct to change those routines to work with any scalar type. Perhaps it is just easier to not allow one to build LCM (and ATO?) when `ENABLE_MESH_DEPENDS_ON_PARAMETERS` is on...

bartgol:
@mperego Yeah, that's what I was starting to wonder. Should we prevent people from trying to build packages other than FELIX with that option on?

mperego:
It might not be too hard to fix LCM and other packages, but since they are not using this capability I'm not sure it is worth. I t would be nice to prevent people to have that option on when building packages other than FELIX or DEMO_PDES. Sorry I did not add that check.

bartgol:
Great! Should we go ahead and merge and see tomorrow what the nightly tells us?

ibaned:
Yea, lets try that.

Issue #233, opened 2017-12-07
Compiling Albany without STK


bgranzow:
I'm attempting to compile Albany without STK enabled. This was a previously supported mode of Albany compilation, enabled in the top-level Albany CMake file by setting the CMake variable `ALBANY_HAVE_STK` if STK was not found in the list of Trilinos packages.

The first three errors I get are posted below:

```
In file included from /gpfs/u/home/PGES/PGESgnzw/scratch/bgq/albany/src/evaluators/state/PHAL_SaveSideSetStateField.cpp:11:
In file included from /gpfs/u/home/PGES/PGESgnzw/scratch/bgq/albany/src/./evaluators/state/PHAL_SaveSideSetStateField_Def.hpp:13:
In file included from /gpfs/u/home/PGES/PGESgnzw/scratch/bgq/albany/src/./disc/stk/Albany_AbstractSTKMeshStruct.hpp:16:
/gpfs/u/home/PGES/PGESgnzw/scratch/bgq/albany/src/./disc/stk/Albany_AbstractSTKFieldContainer.hpp:27:10: fatal error:
      'stk_mesh/base/Field.hpp' file not found
#include <stk_mesh/base/Field.hpp>
         ^
```

```
In file included from /gpfs/u/home/PGES/PGESgnzw/scratch/bgq/albany/src/LCM/solvers/Schwarz_Alternating.cpp:8:
In file included from /gpfs/u/home/PGES/PGESgnzw/scratch/bgq/albany/src/./disc/stk/Albany_STKDiscretization.hpp:14:
In file included from /gpfs/u/home/PGES/PGESgnzw/scratch/bgq/albany/src/./disc/stk/Albany_AbstractSTKMeshStruct.hpp:16:
/gpfs/u/home/PGES/PGESgnzw/scratch/bgq/albany/src/./disc/stk/Albany_AbstractSTKFieldContainer.hpp:27:10: fatal error: 'stk_mesh/base/Field.hpp' file not found
#include <stk_mesh/base/Field.hpp>
```

```
In file included from /gpfs/u/home/PGES/PGESgnzw/scratch/bgq/albany/src/evaluators/state/PHAL_SaveStateField.cpp:11:
In file included from /gpfs/u/home/PGES/PGESgnzw/scratch/bgq/albany/src/./evaluators/state/PHAL_SaveStateField_Def.hpp:14:
[...]

PR #235, opened 2017-12-09
first attempt at issue [#233]

bgranzow:
This is a first attempt at restoring the ability to build Albany without STK enabled in Trilinos.

### DIFF
--- a/src/evaluators/state/PHAL_SaveStateField_Def.hpp
+++ b/src/evaluators/state/PHAL_SaveStateField_Def.hpp
@@ -11,8 +11,11 @@
 #include "Phalanx_DataLayout.hpp"
 
 #include "Albany_StateInfoStruct.hpp"
+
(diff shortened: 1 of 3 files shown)


src/evaluators/state/PHAL_SaveStateField_Def.hpp
```diff
@@ -251,6 +255,9 @@ saveNodeState(typename Traits::EvalData workset)
     default:  // error!
       TEUCHOS_TEST_FOR_EXCEPTION (true, std::runtime_error, "Error! Unexpected field dimension (only node_scalar/node_vector for now).\n");
   }
+#else
```
bgranzow: Maybe a run-time error should be thrown here?

bartgol: I guess you're right.

Issue #242, opened 2018-01-10
Potential issue with current optimized implementation of Interpolation evaluators


mperego:
These interpolation evaluators:
PHAL::DOFInterpolation, 
PHAL::DOFVecInterpolation, 
PHAL::DOFTensorInterpolation,
PHAL::DOFGradInterpolation, 
PHAL::DOFVecGradInterpolation, 
PHAL::DOFTensorGradInterpolation
are specialized for Jacobian evaluations and their implementation is optimized exploiting the fact that the FAD derivatives of the interpolation of the solution (or of the solution gradient) are sparse (for each quadrature point, all the derivatives are 0 except one, so one can compute the derivative only for that nonzero component). @agsalin originally introduced this optimization to significantly speed up the Jacobian computation.
There are two potential issues with this optimized implementation:
1. In order for this to work correctly one has to provide the offset of the solution component (or of the vector/tensor components) considered. Some constructors in the Albany::EvaluatorUtils , set the offset to zero by default when not provided. I think we should avoid providing a default for the offset so that the user is aware that such offset is needed.
2. More importantly, if instead of the solution one is interpolating a derived field (e.g. say the solution is a tensor and one wants to interpolate a field that's the trace of the solution), the interpolation derivatives won't be computed correctly because the sparsity pattern of the derivatives will be different than that of the solution.

These issues are only affecting the Jacobian computation and, in turn, the convergence of the nonlinear solver or the sensitivities, it is not affecting the final solution.

I propose to fix this issue by:
1. removing, for the evaluators above,  the (optimized) Jacobian specialization and let the Sacado AD magic to do the work for us for the Jacobian evaluation. These will be slower but correct also when interpolating a generic field derived from the solution.
2. Introducing new interpolation evaluators, possibly derived from the ones above,  that implement an optimized specialization of the Jacobian that works only for primitive (solution) variables. We can call them something like PHAL::FastSolutionInterpolation, PHAL::FastSolutionVecGradInterpolation.. etc.
3. Modify the corresponding evaluators constructors in the Albany::EvaluatorUtils class so that they accept a boolean flag (e.g. doFastSolutionInterpolation) that tells whether to use the optimized or the non-optimized (safe) version of the interpolation evaluators. The flag will have a default value that will select the safe version of the interpolation evaluators. When the fast interpolation evaluators are requested the offset  will be requested as well. 
4. (I would leave this to the application teams) Modify the problems to use the fast evaluators when possible.

[...]

PR #248, opened 2018-01-17
Interp evaluators

mperego:
This addresses issue #242.


mperego:
@ikalash would you mind checking whether this branch work fine with CUDA builds?
B.t.w. The speed-up I've noticed using the fast interpolation evaluators over the simpler and safer ones is modest. 5 to 12% for the evaluator which is almost not noticeable for the overall test. I'm wondering if it makes sense to keep the fast implementation around. I guess that with the vectorization of FAD types the speed-up will be reduced even further.

ikalash:
@mperego : sure I'll test this with CUDA and report back the result.

ikalash:
I will test on cuda today.

mperego:
@bartgol @ikalash Thanks!

ikalash:
Sorry for the delay.  I tested this branch on Ride with a CUDA KokkosNode and everything looks good.  In terms of total time for the tests, there is a little bit of a speedup but not much (~4%):

master Total Test time (real) = 1740.86 sec
interpEvaluators Total Test time (real) = 1668.37 sec

bartgol:
Uhm, the master branch should already use the fast interpolation, right? I'm surprised this branch is faster than master...

ikalash:
Doesn't it depend on enabling FAST_FELIX whether the optimizations are used?  Maybe I'm mixing it up with something else.  Anyway, a 4% difference is not significant I would say.  You could run the same branch several times and get these sorts of numbers, depending on, e.g., what else is running.

mperego:
@ikalash Thanks a lot! 
@bartgol Yes, the master branch was already using the fast interpolation. Anyway, let's keep it as is for now, but in the future we might decide to get rid of the fast interpolation evaluators if the speed-up small.

PR #261, opened 2018-02-06
Removes global TpetraBuild and replaces with Albany::build_type function

tjfulle:
@ibaned, here is a, perhaps, inelegant solution.

### DIFF
--- a/src/Albany_Utils.cpp
+++ b/src/Albany_Utils.cpp
@@ -564,3 +564,19 @@ void Albany::assert_fail(std::string const& msg) {
   std::cerr << msg;
   abort();
 }
+
+namespace Albany {
+
+BuildType build_type(const BuildType value)
+{
+  static bool initialized_ = false;
+  static BuildType value_ = value;
+  if (!initialized_ && (value != BuildType::None)) {
+    value_ = value;
+    initialized_ = true;
+  }
+  return value_;
+}
+
+} // namespace Albany
+
--- a/src/Albany_Application.cpp
+++ b/src/Albany_Application.cpp
@@ -75,8 +75,6 @@ int countScale;
 int previous_app;
 int current_app;
 
-extern bool TpetraBuild;
-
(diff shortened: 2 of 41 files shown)


tjfulle:
The method uses a similar technique as [Tpetra_Details_Behavior.hpp](https://github.com/trilinos/Trilinos/blob/develop/packages/tpetra/core/src/Tpetra_Details_Behavior.hpp) to get/set "global" flags.

I built locally on my workstation and have only run a few of the tests (all of which pass - both Tpetra and Epetra builds).  I would like to document the function and run all the tests on my blade before merging.

src/Albany_Utils.cpp
```diff
@@ -564,3 +564,19 @@ void Albany::assert_fail(std::string const& msg) {
...
   abort();
 }
+
+namespace Albany {
+
+BuildType build_type(const BuildType value)
+{
+  static bool initialized_ = false;
+  static BuildType value_ = value;
+  if (!initialized_ && (value != BuildType::None)) {
```
bartgol: Should we throw an exception if no build type is selected?

tjfulle:
The build type is only used in a handful of places to differentiate Tpetra/Epetra builds.  I only explicitly set the build type in those executables that previously set TpetraBuild = true.  The only logic that uses the build type is to check if it is a Tpetra build, so what is there is currently sufficient.  I can set the build type to Epetra everywhere TpetraBuild was previously set to false if that is deemed necessary.

bartgol:
Oh, I see. So the 'None' case is kind of like a 'whatever, i don't care' case? If so, then it's fine.

tjfulle:
@bartgol exactly.  And it’s an optional argument so that the function can set/get the build type.  It’ll all be obsolete when Epetra is fully removed.

bartgol:
Well, hold on. May be we keep epetra after all... :)

Issue #260, opened 2018-02-02
floating point exceptions in thermo-mechanical tests


ibaned:
RPI's thermo-mechanical tests started failing today:

http://cdash.sandia.gov/CDash-2-3-0/testDetails.php?test=3437452&build=66342

@lxmota does this have to do with your bulk modulus fix?

gahansen:
It also looks like the thermomechanical can test is failing. I changed the line with "kappa" in J2Model_Def.hpp back to what it was before @lxmota 's commit 26958caacc9362ee6de9c938bb971956c9e634bc and the  ThermoMechanicalCan_thermomech test passes. The FPE is generated down inside the Ifpack2 relaxation solve in MueLu. If I change the preconditioner to Ifpack2, I get the error: 

Ifpack2::ILUT::Compute: zero pivot encountered! Replacing with rownorm and continuing...

@lxmota - any thoughts on what might be broken?

lxmota:
Sorry about this. All those tests will need to be rebased.

The lack of kappa is a serious error in the material models that Max Bloomfield pointed out. The results, even if they pass the tests, are wrong.

ibaned:
@lxmota I agree that more correct physics should always take precedence, and we should do that in this case too. However, these tests are failing to converge and/or getting floating point exceptions, so they will require more than just rebasing to start passing again.

gahansen:
Yes - we expect that we will need to rebaseline the tests, but that will not fix the FPE I don't believe. The addition of kappa in the thermal term appears to create a zero pivot in the preconditioner. Maybe we have the problem setup wrong?

lxmota:
Hmm, that could happen if the material is nearly incompressible, as the bulk modulus would be a very large number, of even infinity if the Poisson's ratio is 0.5.

Are all these RPI tests?

gahansen:
Do we have any other tests that look at thermal expansion other than these?

lxmota:
It appears that we don't other tests for this. Thus the error went undetected for so long.

gahansen:
It looks like the Poisson's ratio is 0.32. Here is the materials file.
[MechanicsWithTemperatureMaterials.yaml.txt](https://github.com/gahansen/Albany/files/1699912/MechanicsWithTemperatureMaterials.yaml.txt)

gahansen:
The tests SCOREC_* are tests that RPI folks committed, but they run on the internal dashboard:

[...]

PR #266, opened 2018-02-21
Consistent treatment of thermomechanics in some LCM material models

tjfulle:
Comments:
Rather than subtract the thermal contribution from the updated stress at the
end of the update, remove the thermal contribution to the deformation gradient
at the beginning of the stress update.  Use only the mechanical portion in
the stress update.  See #260 for motivation.

Addresses: #260
Review: @lxmota

Test Summary:
RHEL6, GCC 6.1, OPENMPI 1.10.1
100% tests passed, 0 tests failed out of 192


tjfulle:
@lxmota

PR #289, opened 2018-04-25
Felix hydrology updates

bartgol:
Fixing hydrology framework, plus one tiny addition in PHAL.

Mostly, this makes the hydrology problem work. There is no test yet, it will come in the next few day. However, I thought to put up this PR to avoid branching of the code.


bartgol:
@mperego Can I merge this?

PR #294, opened 2018-05-16
Stk mesh load field function form

bartgol:
This PR adds PanzerExprEval as optional trilinos package for Albany. If present, STK mesh structs can load a field from file by parsing a mathematical expression. For instance, one can specify

<Parameter name="Field Expression" type="Array(string)" value="{a=2.0,a*x*x}"/>

The branch also adds a tiny test in the new folder tests/small/Utils, for the problem PopulateMesh, which has been moved from FELIX to the main Albany folders. The test simply populates an exodus mesh with some fields, whose info are provided via xml file. Fields can be either loaded from ascii file, set to constant or random values, or computed from mathematical expressions (if PanzerExprEval is available).


src/disc/stk/Albany_GenericSTKMeshStruct.cpp
```diff
@@ -1393,21 +1398,119 @@ void Albany::GenericSTKMeshStruct::loadField (const std::string& field_name, con
...
+    double* xyz;
+    stk::topology::rank_t entity_rank = stk::topology::NODE_RANK;
+    if (scalar) {
+      AbstractSTKFieldContainer::ScalarFieldType* field = metaData->get_field<AbstractSTKFieldContainer::ScalarFieldType> (entity_rank, field_name);
+      for (int i(0); i<entities.size(); ++i)
+      {
+        values = stk::mesh::field_data(*field, entities[i]);
+        xyz    = stk::mesh::field_data(coordinates, entities[i]);
+
+        eval.set("x",xyz[0]);
```
ibaned: This is a slow way to process the data. It would be better to create a Kokkos::View for all the X values for all entities and then evaluate the expression for all entities and get the resulting Kokkos::View. that is how the system was designed to be used.

bartgol: I agree, but I couldn't find a way to extract all the xyz values at once. The only way to do it would be to loop, fill the xyz view, then call the eval method once. I guess that can be done. I didn't bother with that since this is a setup phase only, but I guess for complex math expressions it may be worth it.

I'll fix this.

ibaned: It should be possible to get array data out of STK somehow... might be very complicated though, maybe the fill loop is the easier approach, I don't know.

bartgol:
@ibaned I made the evaluation parallel, and I threw in also a break up of the loadField method into three parts (loadField, fillField and computeField), so hopefully it is less unreadable now.

bartgol:
Something is still off, trying to track down the problem.

bartgol:
Ok, fixed.

Issue #311, opened 2018-07-03
Lack of Albany_config.h


bartgol:
Is there a reason why we shouldn't have something like an Albany_config.h file? I find it useful to have a file in the build directory to check whether a compile definition was enabled or not. Sure, I can compile with VERBOSE=ON, and see all the -D stuff passed to the compiler, but a header file is imho a nicer solution, which allows me to see what's enabled and also what was _enablable_ but not enabled...

Any thought about this?

jewatkins:
That sounds useful to me.

bartgol:
I'll make a PR for this then.

bartgol:
PR #317 took care of this.

PR #317, opened 2018-07-09
Albany config h

bartgol:
Addresses issue #311, plus fixes some minor things here and there.

I am currently testing on my machine with gcc 6.1.0 and a few packages enabled (FELIX, ATO, QCAD, LCM, AERAS), and mesh depending on parameters. However, there are quite a few options in Albany, and I can't test all of them. I will take care of fixing things that make tests not pass once this ends up in master (assuming it does), possibly with some help from packages' developers if I don't understand the logic of the file.

I suggested some reviewers that either showed interest or are responsible of builds, without obligation.

Conceptually, the PR is "small": remove all `ADD_DEFINITIONS(...)` from the main CMakeLists.txt, and add corresponding `#cmakedefine` directives in a config file. Complications arise from the fact that now I have to make sure `Albany_config.h` is included _before_ any `ALBANY_XYZ` macro is used. I think I took care of that for the packages above (I'm at 131/380 tests passed and running), but I may have missed some that are not covered by some tests. If you find some, ding me and I will take care of it.


bartgol:
Yes, I have the bad habit of doing some whitespace reformatting here and there as I go. I'll try to limit myself in the future.

PR #323, opened 2018-07-13
Data types header split

bartgol:
Splitting the data types into multiple headers, and removing some typedefs.

This has zero impact on the current code (all the headers are still included by Albany_DataTypes.hpp), but create a more organized set of types. Ideally, one should not include Albany_DataTypes.hpp, unless all types are needed (e.g., Albany_Application).

The idea behind the split is to separate conceptual areas of the library. Most notably, the model part (Thyra-based), the linear algebra part (Tpetra-based), and the FE-assembly part (Phalanx-Sacado-based).

I did not include a reorganization of the types into namespaces (most notably, wrapping all types (and all namespaces) inside Albany namespace), since it would be a very large set of changes (though, imho, more structured and organized).

### DIFF
--- a/src/ATO/ATOT_Solver.cpp
+++ b/src/ATO/ATOT_Solver.cpp
@@ -18,6 +18,7 @@ Please remove when issue is resolved
 
 #include "Teuchos_ParameterList.hpp"
 #include "Teuchos_XMLParameterListHelpers.hpp"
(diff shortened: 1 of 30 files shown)


bartgol:
Tests still ongoing. Will probably be ready to merge next week.

Issue #324, opened 2018-07-13
KokkosClassic::Details::getNode now deprecated


bartgol:
This free function is used by Petra_Converters to give a default value to the node argument in (almost) all the Epetra->Tpetra conversions. This input is used to to build Tpetra maps.

Tpetra doesn't really need an actual object of the Node type to create a map. Actually, I'm not sure why it is still part of (some of) the constructor(s) signature(s).

Anyhow, I suggest eliminating the use of the node argument in the Petra converters. I grepped through the whole library, and found no use of the Petra converters that passed an actual node object to the routine. Still, removing this will change the interface of the function, so I can't just do it, since someone _may_ actually be using this interface passing a node argument (though I _highly_ doubt it).

If people are cool with it, I will just push a commit directly to master. Otherwise I can include it in PR #323 .

bartgol:
Note: this is a consequence of trilinos/Trilinos#3060, which is addressing trilinos/Trilinos#57 (which, despite the number being way smaller than 3060, dates "only" less than 3 years ago).

bartgol:
The change in Trilinos is triggering a _large_ amount of warnings in Albany. I would suggest to fix this soon. I will prepare a mini-branch, probably on Monday.

bartgol:
I created a tiny branch for this. I will rebase it on master once the data type PR goes in, then put a tiny-PR up.

bartgol:
This was solved by PR #332.

PR #332, opened 2018-07-18
Removed calls to deprecated function in Trilinos

bartgol:
This PR addresses issue #324 (see the issue and commit comments for details).

The PR is very tiny, but I wanted it to go through a PR workflow (rather than merge in master) because it breaks backward compatibility, so it needs to be tracked.

The break is in the Petra conversion routines for maps: the last argument (the node) is no longer present. However, notice the the previous impl had a default for that argument, and nowhere in albany those functions were called passing an explicit last argument; insteady, all calls made use of the default value. Therefore, unless someone is using those functions passing an explicit node, the change should be unharmful.


bartgol:
Since tests pass on my machine, I'm going to merge this. If there are problems in the nightlies, I'll take care of them tomorrow.

PR #334, opened 2018-07-19
Thyra refactor phase 1

bartgol:
This PR will take care of replacing the use of (T/E)petra stuff with Thyra stuff, where possible. The idea is to procrastinate the casting to the actual type to the very last moment, leaving the upper part of the stack petra-free.

It is currently a work in progress. However, I'd like it to be here, so that people can i) see and track what's happening, ii) try out the branch, iii) give feedback. I would also like to break down the PR into multiple PR's, so that it doesn't fall much behind master. The drawback of the multiple PRs approach is that the intermediate ones will have multiple conversions of Thyra objects to/from Petra objects, and will feature a few duplicate interfaces (Tpetra and Thyra). If you have a preference in that regard, please, let me know. My null hypothesis is that it's better to keep the code duplication and number of conversions under control, so do few PRs. So unless someone convince me it is better to merge everytime the tests pass, I will stick with that.


bartgol:
I forgot to mention one warning: I am activating a very limited set of packages during development (FELIX and LCM), to keep compilation/tests times 'reasonable', and run a more full set of packages only once in a while. This means I sometimes find a compilation/runtime error in other packages that should have been dealt with in a previous commit. In such situations, I *rewrite* git's history, and force push to this branch.

If you are planning to work on this branch, be aware that pulls will sometimes fail, due to rewritten remote history. Also, let me know when you do commits, so I will make sure to incorporate them before force pushing. I am aware this is not the cleanest approach, but I'd like each commit to build cleanly with all packages, while also keep my day-to-day builds/tests at a reasonable time.

mperego:
@bartgol Looks good to me. Thanks for working on this!
As for breaking down the PRs in multiple PRs I think it is mostly up to you. 
I agree it's better not to have too many intermediate steps with replicated code.. but if this significantly improves your workflow, I'm fine with that.

bartgol:
The refactor of the evaluateXYZ methods in Application as well as all the responses is half done. So far all the _input_ (multi)vectors (x, xdot, xdotdot, Vx, Vxdot, Vxdotdot, Vp) have been converted to thyra, while the _output_ ones (f, J, JV, fp, g, gx, gxdot, gxdotdot, gp) are still Tpetra.

The refactor is going fairly smoothly. The main concern I have is the scatter/combine phase of distributed objects. Thyra has some support for distributed vectors, but it is basically limited to establishing whether a vector space is distributed or replicated, and distringuish between local and global dimensions. So the scatter/combine has to happen at the Petra level.

So far I am scattering x. xdot, xdotdot in the solution manager, but moving to residual, jacobian and other outputs, I think there will be the need for a Albany structure for scatter/combine, with a single Thyra interface, which dispatches to the correct Petra methods depending on the albany build type.

I can continue to work on this PR or we can merge it and I can open another one, to have a more continuous integration workflow. Any preference? It is the same for me.

bartgol:
Currently fixing tests failing.

bartgol:
All tests pass with OMP_NUM_THREADS=1. With more threads, I have some tests failing (see issue #339), but I think it is an orthogonal issue to this PR.

@mperego , if you are ok with this, I can merge, otherwise I can wait until I am done with the refactor of also the other workset (multi)vectors/operators. Perhaps the latter is the cleanest option.

mperego:
@bartgol Looks good to me, I think it makes sense to merge. Thanks for doing all this work! It must have been tedious..

Are you planning to convert also the Distributed Parameter library to use Thyra?

bartgol:
I just rebased on master, after the merge of the rename_felix_to_landice branch. Running the tests once more, then I will merge. The cahnge in the output vectors (residual/jacobian/responses) will come in a second PR.

Issue #330, opened 2018-07-18
Failing FELIX tests on CUDA Kokkos node


jewatkins:
This issue is a continuation of issue #326. The following FELIX tests are failing:
```
FO_GIS_GisUnstructuredTpetra | Failed | 2m 22s 100ms | Completed (Failed)
FO_GIS_GisAdjointSensitivityTpetra | Failed | 25m  10ms | Completed (Timeout)
FO_GIS_GisSensSMBwrtBetaTpetra | Failed | 25m  10ms | Completed (Timeout)
```
with the following error:
```
terminate called after throwing an instance of 'std::runtime_error'
  what():  cudaFree(m) error( cudaErrorCudartUnloading): driver shutting down /home/projects/albany/build/TrilinosInstall/include/Sacado_DynamicArrayTraits.hpp:233
Traceback functionality not available
```
It turns out, this error is very similar to the errors we were getting in issue #55 with the Epetra tests where we ended up just turning them off in issue #44. I think the issue wasn't with Epetra but with the sensitivities in the Epetra tests. The above tests have sensitivities turned on and that's what's causing the error. I think the issue is that the Fad type for the Tangent is being set to DFad where we would like to use SFad or SLFad for the CUDA tests. I'll see if I can push a fix tomorrow.

ikalash:
Can this issue be closed?

PR #336, opened 2018-07-20
[WIP] Add more flexibility to Fad type choice

jewatkins:
This commit removes the ENABLE_SLFAD and ALBANY_FAST_FELIX config
options and replaces them with the more generic ENABLE_FAD_TYPE
and ENABLE_TAN_FAD_TYPE options. This allows the user to set
the desired Fad type (SFad, SLFad, DFad) for all specializations
or the Tangent, repectively. Sizes are set with ALBANY_SFAD_SIZE,
ALBANY_SLFAD_SIZE, ALBANY_TAN_SFAD_SIZE or ALBANY_TAN_SLFAD_SIZE.

This allows us to deal with issue #330 although the issue still
remains for DFad Tangent on GPUs.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1004,21 +1004,56 @@ ENDIF()
 ## Other Albany Template Options here:
 MESSAGE("\nAlbany Templated Data Type Options:")
 
-# Set FAD data type to static SLFAD if requested.
-OPTION(ENABLE_FAST_FELIX "Flag to turn on Code Optimization for FELIX that
-may break other physics" OFF)
-OPTION(ENABLE_SLFAD "Flag to turn on Code Optimization for ALBANY that
-may break other physics" OFF)
-
-SET(SLFAD_SIZE 32 CACHE INT "set Sacado SLFad size")
-
-IF (ENABLE_SLFAD OR ENABLE_FAST_FELIX)
-  SET(ALBANY_FAST_FELIX ON)
-  SET(ALBANY_SLFAD_SIZE ${SLFAD_SIZE})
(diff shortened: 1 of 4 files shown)


jewatkins:
Although switching to SLFad seems to fix the GPU bug, @mperego pointed out that DFad should work on the GPU. In fact, the current nightly tests use DFad for all tests. I'll see if I can find a reason why Tangent might break the FELIX sensitivities.

agsalin:
Jerry, we have at least one user, Matt Hoffman at LANL, who builds Trilinos and Albany for ice sheets using the old config options. It gets understandable frustrating for him when Trilinos and/or Albany change options, because he rarely can rebuild without committing to a week of debugging. 

Can you add in some logic in the CMakeLists where it checks for the old config options and throws an error message that clearly tells how to alter the cmake config script to get the same behavior as before? Maybe put a date in the comment so we remember to delete it in a year. Thanks. And Thanks for working to improve the code.

ikalash:
@agsalin : we discussed this with @mperego yesterday at the ProSPect meeting.  I brought up Matt Hoffman as a user too.  Mauro did not think it would be a big deal to communicate to him the new options.  Just in case, my suggestion was to put in logic to throw an error when trying to build with FAST_FELIX on so the user does not accidentally try to use the option if it is removed.  I am fine with your suggestion too.

mperego:
I might be wrong but I don't think Matt is using FAST_FELIX. 
We are slowly try to streamline the process of building MALI, so hopefully this will reduce Matt's and Steve's frustration.
Anyway, I'm fine with your solutions, just thought they were not necessary this time.

agsalin:
Thanks for being on top of it already.

jewatkins:
Thanks for the suggestions everyone. I'll throw an error message for now. We can always clean it up later.

jewatkins:
I'm not sure why 32 is the default. @ikalash do you know?

PR #338, opened 2018-07-23
Renamed FELIX to LandIce

bartgol:
This PR is needed since the FELIX driver within MPAS is now the Albany-LandIce (ALI) driver, and the whole thing is Mpas-Albany-LandIce (MALI). It is simply a renaming PR, and can go in whenever.

@mperego , I know you have some outstanding branch with work on FELIX. This can go in after your work.


mperego:
@ikalash @agsalin @jewatkins just want to make sure you are fine with renaming FELIX into LandIce in Albany.
In the past we also discussed calling it simply LI, but we think LandIce is better. It's descriptive and short enough.

ikalash:
The name change is fine with me.  Please proceed.

jewatkins:
LandIce is okay with me too.

agsalin:
Yes, as you like.

bartgol:
Ok, merging this before the thyra refactor branch (which I will rebase on master)

bartgol:
Ah, I apparently didn't pushed a finished up branch. Not sure how. I'll fix it in 5 minutes.

bartgol:
Note: regarding the error for the use of ENABLE_FAST_FELIX, I left it as it is, rather than throwing for 'ENABLE_FAST_LANDICE'. Do you think I should throw in both cases, just in case some MALI user seds (can that be a verb?) FELIX to LANDICE?

jewatkins:
I was thinking about this as I was writing that error. I would say maybe just throw for both just to be safe? We should be able to remove it all completely soon.

mperego:
I would throw only with ENABLE_FAST_FELIX, and I'll get rid of it as well in a few months.. 
We made tons of changes in the input files and other options, so I think Matt or Steve will need our help anyway.
Also if an option is not used you'll get a warning message at the end of the configuration.

bartgol:
That makes sense. Ok, throw only for ENABLE_FAST_FELIX it is.

PR #356, opened 2018-08-15
Thyra refactor phase 2

bartgol:
This PR puts Thyra interfaces in the output parts of the workset (well, most of them): residual, jacobian, responses and derivatives.

It is _almost_ complete, not yet. But I wanted it to be up so you guys can see it.


bartgol:
The last commit should finally complete the PR. All tests (albany, pde, LCM, ATO, LandIce, Aeras) now pass (with gcc, omp, 1 thread), except the following ones:

- `Parallel_Dynamic_Cubes_Newmark_Piro`
- `CrystalPlasticity_OrientationOnMesh`
- `CrystalPlasticity_OrientationOnMesh_np4`
- `CrystalPlasticity_OrientationNotOnMesh`
- `CrystalPlasticity_OrientationNotOnMesh_np4`
- `CrystalPlasticity_MultiSlipHard_Implicit`

However I think they fail for some reasons unrelated to this pr; regression tests pass. Some fail due to a dangling time monitor:

```
*** Teuchos::StackedTimer::report() - Remainder for a block will be *** 
*** incorrect if a timer in the block does not exist on every rank  *** 
*** of the MPI Communicator.                    
```                    ***
Other fail because an exodiff test finds only 1 field in one file (the continuation does not save the solution at all time steps). These _may_ be related to the PR, I am investigating.

However, the PR seems almost ready.
It still needs fixing, but orthogonally to this PR, I think.

**Comment by bartgol** [comment]
I misinterpreted the output of the Teuchos timer. That was not an error, it was simply an info message. I found some bugs, which I fixed, and now I only have one test failing: SteadyHeatConstrainedOpt2D_Dirichlet_Mixed_ParamsT.

This failing test is bizarre: master and the branch give me a very different output at the end. In particular, the branch has a final linear system solve to evaluate a dist param derivative, while I don't see that linear system solve in master. Moreover, both master and branch say that the mean value of the final solution is nan. But that's not all: the branch passes the comparisons in release mode, and fails them in debug mode; in release mode one of the MPI ranks returns a non-zero exit code, so the test fails.

I will investigate this.

**Comment by bartgol** [comment]
Well, this gets more misterious. The test that failed in an odd way on my workstation, passes on my laptop, and there is no sign of nans.

**Comment by bartgol** [comment]
Ok, I give up understanding what's the deal with the test `SteadyHeatConstrainedOpt2D_Dirichlet_Mixed_ParamsT`. It passes when I run it individually (with `ctest -R Mixed_ParamsT`, but fails (with nans in the output) when I run the whole testsuite.

Can someone please run the test in this branch (individually and in the whole testsuite) and tell me what they see? It's the 13th test, and the first ones are fast, so running the whole testuite until that one gets run should not take long.

**Comment by bartgol** [comment]
Finally fixed the last test. I will rebase on master, then I believe this is ready to merge now.

File: src/Aeras/Aeras_HVDecorator.cpp [hunk 1]
```diff
@@ -91,50 +93,40 @@ Aeras::HVDecorator::HVDecorator(
...
-	  mass = createOperatorDiag(1.0, 0.0, 0.0, true);
-  if(Hydro_app)
-	  mass = createOperatorDiag(1.0, 0.0, 0.0, false);
-  Teuchos::RCP<Tpetra_CrsMatrix> laplace;
-  if(SW_app)
-      laplace = createOperator(0.0, 0.0, 1.0, true);
-  if(Hydro_app)
-      laplace = createOperator(0.0, 0.0, 1.0, false);
+  Teuchos::RCP<Tpetra_CrsMatrix> mass = Albany::getTpetraMatrix(createOperatorDiag(1.0, 0.0, 0.0));
+  Teuchos::RCP<Thyra_LinearOp> laplace = createOperator(0.0, 0.0, 1.0);
```
mperego: Why do we always create these?

ikalash:
@bartgol : do you have an estimate of how much effort would be required to port the MOR and QCAD stuff to the new interface?  My guess is it is likely substantial for the Epetra code, specifically if it’s based on epetraext model evaluator (all of MOR, some of QCAD).  If so, I think we need to have a discussion about removing those capabilities from master and moving them to a tag unless there is funding / time available for the porting, as we discussed this morning in CSRI.

bartgol:
Hard to say. I never checked the MOR code. As for the QCAD, it may get a bit ugly, but probably doable in a 1<=x<=3 days of work. However, I wonder whether that would be time well spent. The copious conversions Epetra->Tpetra->Thyra->Tpetra->Epetra will make QCAD hard to read, which means it won't be a great example for a new user. In that case, it might be more efficient to leave a working tag of Albany to be Thyra-free (in the sense that all thyra refactor is not present yet), and provide interested users with that tag. We can even add something to the README.md file, so it shows up in the github main page...

bartgol:
Hold on. I didn't realize two LCM tests failed: `MechanicsWithHydrogenBar_no_stabilization` and `MechanicsWithHydrogenBar_requires_stabilization`.

The first failes cause some fields are not found in the first file during an exodiff command:
`exodiff: DIFFERENCE .. Specified element variable "Cauchy_Stress_72" is not in the first file.`

The second test fails without a specific message, claiming there was an error parsing a cmake file.

HOWEVER: I re-run the failed tests, and they passed.  @ikalash is this related to the recent changes in STK? Was I supposed to do something to re-bless gold files?

ikalash:
Regarding the LCM tests failing: I don't see how you can get the error about the Cauchy_Stress_72 field not being found if (1) this branch is based off a fairly recent master branch of Albany, and (2) you're using a recent Trilinos.  You could double check things by copying the no_stabilization and requires_stabilization subdirs in Albany/tests/large/LCM/MechanicsWithHydrogenBar master to your branch, recompiling and re-running.  

One heads up: I am planning to change the Cauchy_Stress layout in Albany in the near future (see issue #389), which would require another rebaseline of all the LCM tests that run exodiff; but that has not been done that.

bartgol:
The puzzling thing is that the test passed the second time I ran it. Not sure what went wrong, but I don't have time to investigate now. If something goes wrong in the dashboard, I'll take care of it.

ikalash:
Hmmm, very weird.  You can grep the *exodiff files in your test dir for "Cauchy_Stress_72".  It shouldn't be there if your branch has a recent master merged into it.  You can also check if the *gold* exodus files in the dir have Cauchy_Stress_72, which also they shouldn't.  Not sure what was going on,

If this PR is merged in, the Aeras SW tests will be failing - is that right?  Should we look at these a bit before merging?

bartgol:
They will be failing _if you enable the funky mesh dependent on solution/param option(s)_.

To be fair, if you enable those, I _think_ master should fail to compile, since in several places Aeras assumes MeshScalarT = double. This branch would _only_ allow the code to compile, but It's probably best to inhibit the simultaneous enabling of those options with Aeras. What do you think?

bartgol:
I think only LandIce is using these features. I am not sure if they are ever tested. And the issue is that LandIce runs two different sets of tests, depending on whether they are enabled or not, so we should have two sets of builds. I'm not sure what your cism/mpas builds test. Perhaps we could enable it there (since there's no Aeras enalbed there, I think?).

I will add the inhibiting feature today, then if there are no objections, I will merge to master. I will also revert the changes that add 'ADValue(someMeshScalarT)' in Aeras, since they won't be needed anymore, and they may be misleading (suggesting the code works with MeshScalarT=Fad).

ikalash:
Sounds fine to me, thanks for doing this @bartgol!

mperego:
I also get the test SteadyHeatConstrainedOpt2D_Dirichlet_Mixed_ParamsT failing sort of randomly.
It's the only test that combines scalar parameters and distributed parameters.

Failure of FO_GIS_GisSensSMBwrtBetaTpetra might be related as well.

bartgol:
Let me work on those.

Issue #359, opened 2018-08-27
Remove unused/unsupported physics, provide legacy code last working git tag


bartgol:
This issue is to decide what packages are to be discontinued in Albany. By that I mean packages/physics that are no longer maintained, and that have become a maintenance burden, especially when Albany undergoes refactoring efforts.

Such packages should be removed from the current master (together with all their configurations options/macros that are specific to them). We should provide an Albany tag (together with a working trilinos tag) where such packages are still present and functional (= test passing).

Of course, if some user is willing to port such packages to the current master framework, they are more then welcome. We only do this because of the lack of manpower (and funding money) to maintain packages.

Here is a list of the current packages/physics/subfolders that are non-core. Please, add a comment below if you want the package to remain fully supported in the master branch. By requesting such maintenance, you should also commit some of your time to perform such maintenance. We cannot afford to keep packages that have no manpower to maintain them. For such packages, we can only provide a working git tag. I will update this message adding a checkmark next to the package, and one or more names for people responsible for maintaining it.

- [x] Aeras, @ikalash
- [ ] AFRL (waiting for a response, probably ok with a tag)
- [ ] AMP  (ok with tag)
- [ ] ANISO (ok with tag)
- [x] ATO
- [ ] CTM (ok with tag)
- [ ] Hydride (ok with tag)
- [x] LandIce, @bartgol, @mperego
- [x] LCM, @ikalash, @lxmota (Alejandro, @ikalash is volunteering you)
- [ ] MOR (ok with tag)
- [x] Peridigm (keep, owner?)
- [ ] QCAD (ok with tag)
- [x] SCOREC (keep, owner?)
- [x] Tsunami, @ikalash
- [ ] Dakota (ok with tag)

I would like to get a response by next week, so that we can take a decision at the next Albany conference call.

bartgol:
I should add that if a package has no issues or can be maintained with minimal effort, we should try to keep it in master. I am not trying to discontinue all packages that are not working but could with just a few minutes of coding. We are still _trying_ to make an effort to keep Albany a truly multiphysics library. However, there are some efforts ongoing that are already requiring a lot of time, and for which it would take considerable effort to maintain all packages.

ikalash:
Should we tag the last known owners of some of the packages to bring this issue to their attention and have them comment?

bartgol:
[...]

PR #363, opened 2018-09-13
Removing support for some Albany packages.

bartgol:
This addresses issue #359 .

The folder of discontinued packages are still present, containing only a single README file, explaining how to retrieve the discontinued code. For each discontinued package, I also created a tag `${PACKAGE_NAME}_support_end`. These tags all refer to the same commit in Albany (and in Trilinos/develop), but I still thought individual tags are safer (just in case some package gets updated in the future, and then discontinued again).

To revive a package, one needs to checkout the tag `${PACKAGE_NAME}_support_end`:

`git checkout tags/${PACKAGE_NAME}_support_end`

and then look at the tag comment with

`git tag -v ${PACKAGE_NAME}_support_end`

which will specify the Trilinos commit necessary to successfully build Albany and test the current package.

Additionally, the main README.md file in Albany has a paragraph (at the end) explaining what happened to the apparently missing packages.


bartgol:
Given the 4 approvals, unless some red flag is raised, tomorrow morning I'll merge this into master.

Issue #372, opened 2018-10-18
Remove depcrecated code warnings from Trilinos


bartgol:
There are two outstanding ones: Tpetra::[Mpi|Serial]Platform being deprecated, and STK deprecating the use of default init value in put_field. They generate an insane amount of warnings, and is better to move away from deprecated code anyways.

bartgol:
The recent changes in STK have deprecated 4+ dimensional fields (per entity). This means that `QPTensor3FieldType` in Albany is no longer a legit field type. It will have to be purged, in order to comply with future trilinos versions (where the stk deprecated methods will likely be removed).

bartgol:
@ikalash you interacted quite a bit with the stk people during the resolution of the errors introduced by their PR. Did they mention (or hint at) why they removed 4+ dimensional fields?

ikalash:
@bartgol I don't think this came up in my discussion on the relevant github issue - you can have a look here at the discsussion: https://github.com/trilinos/Trilinos/issues/3377 .  @alanw0 would know - Alan, have 4+ dimensional fields been removed from STK?

alanw0:
@ikalash and @bartgol, when we converted from put_field to put_field_on_mesh we didn't convert 4D and up, but only because we couldn't locate anyone who was using those. What is the use-case? I.e., what kind of field is this? If you need this I think we can go ahead and restore them.

alanw0:
And @bartgol if this is something you need, go ahead and re-open the issue you started about this, and we'll take care of it.

bartgol:
Well, there was _some_ code in Albany that stored a 3d tensor for each quadrature point. I'm not sure if this is really used. I tried to eliminate the 3d tensor fields alltogether from Albany's meshes, and I will check if anything fails in the nightlies. @ikalash I think 3rd order tensor is something that LCM could have used. Do you recall any test using them?

@alanw0 I think higher order tensors may still be interesting for _some_ application (I'm thinking about elasticity, for instance; or general relativity, but I am not sure how many people use stk for general relativity calculations). However, I am not going to reopen the issue, unless we realize the  3d tensor at quadrature points is indeed still used by some tests.

bartgol:
Fixed by #373 . Unless a real need for qptensor3d comes up again, I think we're fine.

bartgol:
@alanw0 Great, thanks! I will add a note in Albany that 4d and 5d are now again available in STK, if someone needs them. I don't see us ever needing 6+ dimensional fields...

PR #373, opened 2018-10-18
Purge deprecated calls

bartgol:
This PR takes care of issue #372 (plus it removes some leftover code that depended on `ALBANY_MOR` being defined).

Besides changing deprecated function names to non-deprecated ones, this PR also *removes* the field type `QPTensor3FieldType` from the STK discretizations structures. This is because the new implementation of the deprecated function in STK is only present for fields up to 3 dimensions (per entity), so the maximum we can add to the mesh is a 2d tensor per QP or Node (within an element) or a 3d tensor per Cell.

### DIFF
--- a/src/Albany_Application.cpp
+++ b/src/Albany_Application.cpp
@@ -234,21 +234,6 @@ void Albany::Application::initialSetUp(
     tangent_deriv_dim = 1;
   }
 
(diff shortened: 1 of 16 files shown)


bartgol:
All LandIce, LCM, Aeras and basic tests passed. I'm merging this.

Issue #374, opened 2018-10-19
MALI infrastructure work


mperego:
There is some infrastructure work needed in Albany LandIce to improve the overall MALI (MPAS-Albany LandIce) code:
1) Use the new boundary conditions instead of using PHAL::Neumann in the MPAS Interface
2) Use the ExtrudedSTKMeshStruct instead of the MpasSTKMeshStruct in the MPAS Interface
3) Make MALI compile with shared libraries. This might need a little bit of refactoring and avoid declaring some function as extern in the c++ interface (in MPAS repo).

We can split this in different issues if needed.

bartgol:
For 3., do you mean getting MPAS to link to a shared albanly library? If so, then we can also put the files that define the extern function into a third small library and link as `mpif90 main_mali.F90.o -lalbanyLib.so -llibrary_with_extern_definitions.so`. That should work too, no?

mperego:
@bartgol yeah I think you are right. But I'm not sure we really need to declare those functions as extern.

bartgol:
I think the mpas infrastructure now works. We did not fulfull point 2 in @mperego 's list, but the interface works, so I'm going to close this for now.

PR #380, opened 2018-11-02
LandIce: move away from PHAL Neumann, plus introduce StokesFOBase problem

### LINKED ISSUE #378: PHAL_Neumann: robin/radiate bcs
These two conditions are quite confusing and need to be refactored a bit, I think. A few observations:

- robin imposes the bc `du/dn = A*(u-u_0) + jump*scale*2`, where `A, u_0` are a given _constants_ (not functions);
- radiate imposes the bc `du/dn = A*(u^4 - u_0^4)`, with `A, u_0` as above;
- in literature, radiation and robin sometimes refer to the same boundary condition;
- radiate is quite specific, and could be easily generalized to accommodate any given exponent
- the class does not allow to impose robin/radiate on a vector field
- as it is, robin _requires_ to have a material database, which is used to retrieve the value for `scale` (while `jump` is passed in via input file).

My thoughts:

- the name 'robin' is _very_ common in literature, and should allow to _easily_ implement the classic `du/dn=A(u - u_0)`, without the need of a material database;
- there's no particular reason why one should not be able to impose a robin bc on a vector dof;
- there's a comment in `calc_dudn_robin` saying 'mult by 2 to emulate behavior of an internal side within a single material (element block) in which case usual Neumann would add contributions from both sides, giving factor of 2'. I get the explanation, but isn't this a corner case? The 2 is still confusing (especially if one does not know there is a 2, and thinks he's imposing an exact jump). One could still recover the 'emulated' effect by specifying a jump 'twice as large'...
- if one really wants to impose a robin condition with a jump, he/she should impose two conditions: robin _and_ jump. Grepping in the repo, there's only one test (two if you count E/T) that imposes a robin-jump combo bc. Keeping robin the way it is just to accommodate that test is imho short sighted;
- radiate should probably be called something like 'nonlinear robin' or 'power-law flux'.

If there are no objections, I want to change robin so that it is a _real_ robin bc (which is very reusable). It would really help with some LandIce test. Changing 'radiate' to a different name would be imho good for new users (less confusion, see my third comment), but I can live with it for a while...

bartgol:
This PR does a bunch of things.

1) Fixes #378 
2) Does some work towards part 1 of #374 . In particular, we now have a `LandIce BCs` sublist in the problem section, where we can handle basal, lateral, and syntetic-tests boundary conditions. They are longer sublists than the NBCs and DBCs ones. I am planning to work on a design similar to that in NBCs and DBCs to have shorter input files (something like `LIBC on SS blah for DOF u set basal friction`). There are a few details to solve though, so for now I just went with a more verbose input file.
3) Introduces a `StokesFOBase` class, meant to be used as base class for all `StokesFO*` problems. The idea is that these problems share a lot of common parts (the ice velocity equatiions, plus states registrations routines). Currently, there is a lot of repeated code which makes maintaining the multiplle `StokesFO*` problem hard (whenever you add a feature, you have to adjust all problems).


bartgol:
@jewatkins I haven't tested the branch on GPU yet. But even if tests pass, I would need your expertise to check if there is a performance regression. I expect a bit of regression on some tests, since I _think_ the generalization of having a `StokesFOBase` class _may_ add some extra evaluation steps. not 100% sure, but it's definitely a risk, so we should check for performance changes.

jewatkins:
@bartgol Thanks for the heads up. We can check the performance once it's pushed and the cuda build is working again.

PR #390, opened 2018-11-19
LandIce: Changes for enabling Regularized Coulomb Friction conditions

mperego:
Minor changes to enable Regularized Coulomb Friction conditions in MALI

### DIFF
--- a/src/LandIce/interface_with_mpas/Interface.cpp
+++ b/src/LandIce/interface_with_mpas/Interface.cpp
@@ -125,6 +125,7 @@ void velocity_solver_solve_fo(int nLayers, int nGlobalVertices,
   ScalarFieldType* stiffeningFactorField = meshStruct->metaData->get_field <ScalarFieldType> (stk::topology::NODE_RANK, "stiffening_factor");
   ScalarFieldType* effectivePressureField = meshStruct->metaData->get_field <ScalarFieldType> (stk::topology::NODE_RANK, "effective_pressure");
 
+  bool nonEmptyEffectivePressure = effectivePressureData.size()>0;
   for (UInt j = 0; j < numVertices3D; ++j) {
     int ib = (ordering == 0) * (j % lVertexColumnShift)
         + (ordering == 1) * (j / vertexLayerShift);
@@ -146,7 +147,7 @@ void velocity_solver_solve_fo(int nLayers, int nGlobalVertices,
     stiffeningFactor[0] = std::log(stiffeningFactorData[ib]);
     double* effectivePressure = stk::mesh::field_data(*effectivePressureField, node);
 
-    if(effectivePressure != NULL)
+    if(nonEmptyEffectivePressure && (effectivePressure != NULL))
       effectivePressure[0] = effectivePressureData[ib];
     
     if(smbField != NULL) {
@@ -347,8 +348,7 @@ void velocity_solver_extrude_3d_grid(int nLayers, int nGlobalTriangles,
     const std::vector<int>& dirichletNodesIds,
     const std::vector<int>& floating2dEdgesIds) {
 
-  slvrfctry = Teuchos::rcp(
-      new Albany::SolverFactory("albany_input.xml", mpiCommT));
+  slvrfctry = Teuchos::rcp(new Albany::SolverFactory("albany_input.yaml", mpiCommT));
   paramList = Teuchos::rcp(&slvrfctry->getParameters(), false);
 
   Teuchos::Array<std::string> arrayRequiredFields(9);
--- a/src/LandIce/interface_with_mpas/Interface.cpp
+++ b/src/LandIce/interface_with_mpas/Interface.cpp
@@ -69,6 +69,7 @@ void velocity_solver_solve_fo(int nLayers, int nGlobalVertices,
     const std::vector<double>& bedTopographyData,
     const std::vector<double>& smbData,
     const std::vector<double>& stiffeningFactorData,
+    const std::vector<double>& effectivePressureData,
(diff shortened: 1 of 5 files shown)


src/LandIce/interface_with_mpas/Interface.cpp
```diff
@@ -412,6 +416,16 @@ void velocity_solver_extrude_3d_grid(int nLayers, int nGlobalTriangles,
     std::cout<<"\nWARNING: Using Dirichlet BCs options provided in Albany input file. In order to use those provided by MPAS, remove \"Dirichlet BCs\" sublist from Albany input file.\n"<<std::endl;
   }
 
+  if(paramList->sublist("Problem").isSublist("LandIce Field Norm") && paramList->sublist("Problem").sublist("LandIce Field Norm").isSublist("sliding_velocity_basalside"))
```
mperego: @bartgol This is a bit ugly (you need to know how the field is named in the solver) but I think it is fine.

src/LandIce/problems/LandIce_StokesFOBase.hpp
```diff
@@ -546,9 +547,9 @@ constructVelocityEvaluators (PHX::FieldManager<PHAL::AlbanyTraits>& fm0,
   ev = evalUtils.getPSTUtils().constructNodesToCellInterpolationEvaluator ("temperature",false);
   fm0.template registerEvaluator<EvalT> (ev);
 
-  // Interpolate ice softness (aka, flow_factor) from nodes to cell
-  ev = evalUtils.getPSTUtils().constructNodesToCellInterpolationEvaluator ("flow_factor",false);
-  fm0.template registerEvaluator<EvalT> (ev);
+//  // Interpolate ice softness (aka, flow_factor) from nodes to cell
```
mperego: @bartgol I had to comment this because it conflicts with the FlowRate evaluator registered at line 646. Is that OK to comment this out, or should we put some logic to address this?

mperego:
@bartgol A line
https://github.com/gahansen/Albany/blob/master/src/LandIce/evaluators/LandIce_BasalFrictionCoefficient_Def.hpp#L120
I would choose the exponent to be `p-1` instead of `p`, to be somewhat consistent with the regularized Coulomb friction.

mperego:
@bartgol The `p` in the Regularized Coulomb coefficient, it is often associated with `1/n` where `n` is the Glen's Law exponent. I'm wondering if it is misleanding. Also it is a bit confusing if one thinks at `p` to be the exponent of the p-laplacian (which would be `1+1/n`). Anyway, I fine with this if you prefer this notation.

mperego:
@bartgol Would you mind adding the units of each parameter (I like to do that at the parameter declaration).

src/LandIce/problems/LandIce_StokesFOBase.hpp
bartgol: You may be right. I think I added that interpolation because at some point the flow_factor was loaded from file at nodes. Maybe I was wrong, but the scenario was never tested... Anyhow, the way it is now should be fine.

bartgol:
@mperego 

> Would you mind adding the units of each parameter (I like to do that at the parameter declaration).

Sure. You mean in the input xml (or yaml) file?

> At line
https://github.com/gahansen/Albany/blob/master/src/LandIce/evaluators/LandIce_BasalFrictionCoefficient_Def.hpp#L120
I would choose the exponent to be p-1 instead of p, to be somewhat consistent with the regularized Coulomb friction.

I chose it to do p instead of p-1 because that way I had the same power at both numerator and denominator, and could compute just one pow. If you do p-1, you need a pow for the numerator and a pow for the denominator. Besides, if you pick p=0, the term in parentheses goes away, and you are left with the usual Coulomb friction law. Of course, even with two different powers for numerator and denominator you end up with Coulomb for p=0, but it's harder to see that way (I think).

> The p in the Regularized Coulomb coefficient, it is often associated with 1/n where n is the Glen's Law exponent. I'm wondering if it is misleanding. Also it is a bit confusing if one thinks at p to be the exponent of the p-laplacian (which would be 1+1/n). Anyway, I fine with this if you prefer this notation.

I'm not sure I follow what you are suggesting here. Yes, p=1/n in most of the literature (although that would be a lucky coincidence, since sliding and ice rehology seem uncorrelated to me). I never put that explicitly since, back when I started coding this, I was interested in estimating the exponent, so I wanted to be untied to Glen's law. If you are suggesting that the symbol 'p' is misleading cause one thinks about the p-laplacian, then ok, let's change it to something else. I think I always call it 'power' in the code, but any name would do. I have no strong preference. 'q' seems to be quite popular for power exponents too.

src/LandIce/interface_with_mpas/Interface.cpp
mperego: The velocity is computed by Albany. That conditional is mainly for allowing to change the regularization (needed for schoof) in the input file. Eventually Mpas should provide the parameters

bartgol:
Ah, the power law. Well, I never really tested that in the last N+1 years, so I can't even guarantee it works. But i'm fine converting it to p-1.

bartgol:
As for the units, I think there are a lot of evaluators where we should add that information. Let's make it another PR? It would be nice to have units next to all parameters and fields.

mperego:
@bartgol I'm fine with having another PR for units, but I need to sort units out for the Regularized Friction now so that we know hot to use it from MPAS

mperego:
@bartgol can you check my last commit? If it is good for you I'll merge the branch

bartgol:
Looks great, thanks!

PR #397, opened 2018-11-29
Extruded sides

mperego:
Miscellaneous changes including 
- modifications do the ExtrudedMeshStruct to extrude side sets and node sets of the basal mesh
-  (LandIce) bug fix for wedge elements
-  (LandIce) changes to the gather of VerticallyAveragedVelocity so that the evaluated field is defined on sides
-  (LandIce) modifications to the UpdateZCoordinate

### DIFF
--- a/src/LandIce/problems/LandIce_StokesFO.hpp
+++ b/src/LandIce/problems/LandIce_StokesFO.hpp
@@ -141,32 +141,63 @@ StokesFO::constructEvaluators (PHX::FieldManager<PHAL::AlbanyTraits>& fm0,
     ev = evalUtils.constructGatherCoordinateVectorEvaluator();
     fm0.template registerEvaluator<EvalT> (ev);
 #else
+    bool adjustBedTopo = params->get("Adjust Bed Topography to Account for Thickness Changes", false);
+    bool adjustSurfaceHeight = params->get("Adjust Surface Height to Account for Thickness Changes", false);
 
-    //----- Gather Coordinate Vector (ad hoc parameters)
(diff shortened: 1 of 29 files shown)


src/LandIce/problems/LandIce_StokesFO.hpp
```diff
@@ -141,32 +141,63 @@ StokesFO::constructEvaluators (PHX::FieldManager<PHAL::AlbanyTraits>& fm0,
...
+      p->set<std::string>("Top Surface Name", "observed_surface_height");
+      p->set<std::string>("Updated Top Surface Name", "surface_height");
+      p->set<std::string>("Bed Topography Name", "observed_bed_topography");
+      p->set<std::string>("Updated Bed Topography Name", "bed_topography");
+      p->set<Teuchos::ParameterList*>("Physical Parameter List", &params->sublist("LandIce Physical Parameters"));
+
+      ev = Teuchos::rcp(new LandIce::UpdateZCoordinateMovingBed<EvalT,PHAL::AlbanyTraits>(*p, dl));
+      fm0.template registerEvaluator<EvalT>(ev);
+    }
+    else if(adjustSurfaceHeight && !adjustBedTopo) {
```
bartgol: This case seems to never be tested (all tests adjust bed topography). Can we change one test case so that it tests this?

mperego: It makes sense, I'll do that.

mperego: I added the test. I'm going to merge. Thanks.

Issue #399, opened 2018-12-07
Make sure no two evaluators evaluate the same field


bartgol:
Following trilinos/Trilinos#4005, we should now make sure that each field is evaluated by exactly one evaluator (1). This is currently a problem in at least Aeras and LandIce. It's going to take some work to clean these up, but I do think it's worth it. It was me who opened the issue requesting this feature, since I already spent 4h of debugging (twice!) to find out I had two evaluators evaluating the same field.

In most cases (and I think all the failing tests on the dashboard that fail for this reason), the two evaluators are in fact two copies of the same evaluator. In this case, it would be _safe_ to overlook the repetition, and allow PHX to discard one of the two. However, there are cases (as I said, it happened to me twice), where two _different_ evaluators evaluate the same field. Perhaps you refactored the code, but left a piece of the old one in, and now you have ev1 and ev2 both evaluating the same field. Or even worse: you _always_ had two evaluators evaluating almost the same field: the two fields just differ because of their dimensions, but changing mesh element type (or the vector dimension) will make them clash, and you would not notice it.

In short, I think that, despite being annoying to fix things that basically worked, this is after all a useful improvement to the library. I will take care of fixing all the LandIce tests in the next few days. @ikalash do you think you have time to look at the Aeras ones? If it is the same source of error as in LandIce, there is probably an interpolation evaluator that is registered multiple times.

(1) Note: _contributed_ fields are ok.

jewatkins:
It looks like this was the issue I was having in #392.

ikalash:
Yes I can look at Aeras.  Seeing what you did @bartgol for LandIce should be helpful.  

A more general comment: I think we need to fix the nightlies before we proceed with the Albany refactor.  I hope to make some progress on this next week.  I am sorry to keep holding things up with the refactor...  but the testing is important and the more code we push w/o fixing lingering issues, the more difficult it will be to fix things later.

bartgol:
I'm working at fixing `LandIce_StokesFOBase.hpp`, which is where the evaluators that cause problems are registered. I hope to put a mini-PR up in the next few days.

ikalash:
Thanks @bartgol .  I am looking at this now.  Can you please expand on what you think is the problem when the error messages such as the following, where ev1 = ev2 in your notation:
```
[...]

PR #412, opened 2018-12-18
Fix 399

bartgol:
This PR addresses #399 . After the change in Trilinos, the default behavior in Phalanx (for a common trilinos installation for Albany) is to throw if two evaluators try to evaluate the same field.

Obviously, a quick fix is to disable this feature, but I was the one to request it in Trilinos, so you probably can see that I don't like the idea of disabling it. This is because it _can_ indeed be a bug (and a very hard to find one) if two evaluators evaluate the same field.

So we had to do a bunch of fixes to the albany problems, to avoid duplicates. @ikalash already fixed a bunch of these errors in ATO, Aeras and LCM. This PR fixes the LandIce tests.

The PR contains a sizeable refactoring of the StokesFOXYZ classes. The underlying idea is that we want to automatize as much as possible the creation of the evaluators, with the goal to minimize code duplication and facilitate extension of StokesFOBase to other StokesFOXYZ problems.

There is still one test failing (FO_GIS_GisUnstructuredPerformance_Tpetra), where the residual of Newton seems to not really decrease. Not sure why. However. I think this PR should be merged, so that it will allow some tests to stop failing, as well as other LandIce developer to adapt to design changes.

### DIFF
--- a/src/LandIce/problems/LandIce_StokesFOBase.hpp
+++ b/src/LandIce/problems/LandIce_StokesFOBase.hpp
@@ -1453,13 +1453,9 @@ StokesFOBase::constructStokesFOBaseResponsesEvaluators (PHX::FieldManager<PHAL::
 
     // Figure out if observed surface velocity RMS is scalar (if present at all)
     if (!isInvalid(surfaceSideName)) {
-      auto it1 = stateMgr.getRegisteredSideSetStates().find(surfaceSideName);
-      if (it1!=stateMgr.getRegisteredSideSetStates().end()) {
-        std::string surfEBName = meshSpecs.sideSetMeshSpecs.at(surfaceSideName)[0]->ebName;
-        auto it2 = it1->second.find(surfEBName);
-        if (it2!=it1->second.end()) {
-          auto where = it2->second.find("observed_surface_velocity_RMS");
-          paramList->set<bool>("Scalar RMS",where!=it2->second.end() && where->second->rank()==3);
+      if (is_ss_input_field[surfaceSideName]["observed_surface_velocity_RMS"]) {
+        if (ss_field_rank[surfaceSideName].at("observed_surface_velocity_RMS")==0) {
+          paramList->set<bool>("Scalar RMS",true);
         }
       }
     }
@@ -1475,7 +1471,7 @@ StokesFOBase::constructStokesFOBaseResponsesEvaluators (PHX::FieldManager<PHAL::
     paramList->set<std::string>("Thickness Side QP Variable Name","ice_thickness_" + basalSideName);
     paramList->set<std::string>("Thickness Side Variable Name","ice_thickness_" + basalSideName);
     paramList->set<std::string>("Bed Topography Side Variable Name","bed_topography_" + basalSideName);
-    paramList->set<std::string>("Surface Velocity Side QP Variable Name","surface_velocity");
+    paramList->set<std::string>("Surface Velocity Side QP Variable Name",dof_names[0] + "_" + surfaceSideName);
     paramList->set<std::string>("Averaged Vertical Velocity Side Variable Name","Averaged Velocity");
     paramList->set<std::string>("SMB Side QP Variable Name","surface_mass_balance_" + basalSideName);
     paramList->set<std::string>("SMB RMS Side QP Variable Name","surface_mass_balance_RMS_" + basalSideName);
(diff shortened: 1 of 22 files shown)


jewatkins:
@ bartgol If the problem is with memoization, turning it off should yield the correct result. It's possible that turning on memoization could give a different result if it's not implemented correctly. That's because, currently, it has to be manually inserted to each evaluator in the tree.

src/LandIce/problems/LandIce_StokesFOBase.hpp
```diff
@@ -835,133 +1010,16 @@ void StokesFOBase::constructBasalBCEvaluators (PHX::FieldManager<PHAL::AlbanyTra
...
-    //---- Interpolate thickness gradient on QP on side
-    ev = evalUtils.getPSTUtils().constructDOFGradInterpolationSideEvaluator(ice_thickness_side, ssName, enableMemoizer);
-    fm0.template registerEvaluator<EvalT>(ev);
-
-    //---- Interpolate surface height on QP on side
-    ev = evalUtils.getPSTUtils().constructDOFInterpolationSideEvaluator(surface_height_side, ssName);
-    fm0.template registerEvaluator<EvalT>(ev);
-
-    //---- Interpolate basal_friction (if needed) on QP on side
-    basalMemoizer = enableMemoizer ? (!is_dist_param["basal_friction"] ? true : false) : false;
```
jewatkins: It might have something to do with this tricky business with the basal boundary conditions. `basalMemoizer` turns on the memoizer for the basal bc evaluators if it's not a distributed parameter.

bartgol: Maybe. Although the performance test does not have any distributed parameters.

bartgol:
I think I tracked it down. I was probably enabling memoization for solution-dependent evaluators. Not bad.

bartgol:
@ikalash I need to test the code with the funky mesh/param depend on solution options. If those pass on my laptop, I will push the ready to merge version.

bartgol:
FIY, with the design in this branch (automatized interpolation evaluators construction) I am disabling memoization if there are distributed parameters at all. This is because if a field depends on a distributed parameter, we should disable memoization for that, but I have no way of checking this dependencies in an automatic way.

The nice/clean thing to do would be to have a `Albany::FieldManager`, which checks that all evaluators that are claiming to use memoization indeed have the right to do so (i.e., they do not depend on solution or any other field that is non-memoizable), and if they don't, disable memoization in the post registration setup.

jewatkins:
@bartgol That's fine. Your suggested approach will have to be done anyway for other options so we can hold off on bc memoization until then.

bartgol:
To be clear, memoization _is_ available if there are _no_ distributed parameters. Not always the case for us, but sometimes yes.

bartgol:
@ikalash Yes, we should test that option. We sometimes use it in LandIce (and will probably use it even more in the future), and it features funky ETI patterns, which are different from when those options are all disabled. I just caught one when testing my branch.

ikalash:
I just opened at issue #413 .

bartgol:
Alright, Aeras, ATO, LandIce, LCM, and demo pdes tests pass on my workstation (gcc 7.2). LandIce tests pass also with funky mesh/param depend on solution options enabled.

ikalash:
Great.  I'm OK with merging this in and seeing what happens with the nightlies tonight.

bartgol:
I thought this PR already went in, since tests stopped failing. @ikalash are the nightlies enabling multiple evaluators for the same field in phalanx?

ikalash:
No, @bartgol , it did not.  I knew you took off for the break and I wanted to see / fix other issues in the LandIce tests so I set ```-D Phalanx_ALLOW_MULTIPLE_EVALUATORS_FOR_SAME_FIELD:BOOL=ON``` in the nightlies.  Once the PR goes in I can remove that option and see what happens.  Note that the Enthalpy test still fails even if we have ```-D Phalanx_ALLOW_MULTIPLE_EVALUATORS_FOR_SAME_FIELD:BOOL=ON``` (https://github.com/SNLComputation/Albany/issues/392).  I wasn't sure if this PR was supposed to have a fix for that test too - does it?

bartgol:
I'm re-running the tests, but IIRC all tests passed on this branch. Including the enthalpy one.

ikalash:
OK, that would be great if they do.  Hopefully it's still the case.  I can remove the ```-D Phalanx_ALLOW_MULTIPLE_EVALUATORS_FOR_SAME_FIELD:BOOL=ON``` line in Trilinos once this PR goes in to see what happens.

bartgol:
100% tests passed on my workstation. I'm going to merge.

PR #424, opened 2019-01-24
Fix 399 part 2

### LINKED ISSUE #399: Make sure no two evaluators evaluate the same field

bartgol:
This branch fixes #399 (at least within LandIce).

It features consistent modifications of StokesFOXYZ classes, as well as changes to some LandIce evaluators (mostly adding template arguments on input/output fields scalar types).

There is a catch: some LandIce tests fail when MESH_DEPENDS_ON_XYZ is enabled. I tracked down the problem, and I _believe_ the problem was indeed in master, in some input files (both yaml and ascii). This will require to create some new ascii files, and possibly calibrate parameters in order to achieve convergence again (and hopefully to a reasonable solution).

While fixing this issue is important, I feel like it should not block this PR from going in, given that all tests pass when no funky MESH_DEPENDS_ON_XYZ option is enabled. This would also allow further development of LandIce to be built on top of the modifications in this branch.


bartgol:
With the last few commits all LandIce tests pass when MESH_DEPENDS_ON_XYZ is on, except for the two ShapeOpt tests (one test, E/T versions). As mentioned before, this is most likely related to bad inputs and/or parameters, which will need to be adjusted. Since this may take time, and since we have no testbed that tests the ShapeOpt tests, we can merge this now, then fix the tests later.

bartgol:
I tested all packages, and outside of LandIce I get only two tests failing (StaticElasticity3D and StaticElasticity3D_Traction) due to FPEs. Since those failures should be unrelated to this branch, I'm going to merge this.

Issue #394, opened 2018-11-27
Failing Peridigm build


ikalash:
It looks like the Peridigm build has been failing since it was created on 11/9/18: 

http://cdash.sandia.gov/CDash-2-3-0/viewBuildError.php?buildid=77971

ikalash:
I'd like to revive this issue, since it is going to be relevant very soon with the refactor @bartgol and I are doing to remove Petra from Albany and replace it with Thyra.  

What are the plans for reviving Peridigm and corresponding timeframe for doing so?  Peridigm has a ton of Petra in it and currently does not compile.  I would not feel comfortable removing Petra from Peridigm knowing that there are issues preventing the code from compiling even with the Petra there.  Perhaps if it isn't clear when Peridigm will be revived, we should discuss moving it to a tag?  Tagging @mperego , @djlittl .

djlittl:
There's really no hope of refactoring the Peridigm coupling anytime soon.  I think moving it to a tag is the best idea.

bartgol:
The question is, when to we tag it? Not current master, obviously, since it's failing. Do we know when it was working?

mperego:
@djlittl can we tag the branch you are working on?

ikalash:
FYI - I turned off this build on the CEE since Peridigm will be moving to a tag soon.

ikalash:
Peridigm is being removed in the unpetrify_branch (moved to a tag) and I've disabled the nightly testing which has been failing for months if not years.  Therefore closing this issue.

PR #454, opened 2019-03-19
Unpetrify branch

bartgol:
This PR removes [T|E]petra objects from as much code as possible, relying on routines in `Albany_ThyraUtils.hpp` to extract/manipulate internal data of Thyra objects.

The code is currently compiling, but not all tests are passing (hence, the WIP).

Still, people can take a look, and see if they have concerns/suggestions.

Edit: I should have pointed out that the code compiles when no physics is enabled (not even demo pdes).


bartgol:
Update: all basic albany tests pass. The physics packages (including the demo pdes) are still failing.

bartgol:
Update: Albany builds with demo pdes and landice enabled. All but one of the demo pdes tests pass. LandIce is a bit worse, with all the GLF tests failing and the PopulateMesh test hanging.

bartgol:
Here's a todo list for this PR. @ikalash , let's check things off this list as we progress. Also, feel free to add more items.

- [ ] pumi meshes/discretizations
- [ ] ATO
- [ ] Aeras
- [ ] LCM
- [ ] miscellanea albany evaluators still relying on Tpetra back-end

bartgol:
Update: the code builds, runs, and passes all tests when enabling demo pdes, landice and scorec, including the funky `ENABLE_X_DEPENDS_ON_Y` optiions for the LandIce tests (except for the ShapeOpt tests, which fail also in master).

src/disc/stk/Albany_STKDiscretization.cpp
```diff
@@ -457,7 +457,7 @@ STKDiscretization(const Teuchos::RCP<Teuchos::ParameterList>&  discParams_,
  , discParams(discParams_)
  , interleavedOrdering(stkMeshStruct_->interleavedOrdering)
 {
-  // Nothing to be done here
+  this->updateMesh();
```
bartgol: This is wrong. The method `updateMesh` calls virtual methods, which C++ does not allow (or, better, it allows it, but it won't do what you expected). This call got into the constructor somehow a while ago, I took it out again in a recent commit.

ikalash:
I know I fixed it in the next commit.

Best,

Irina

Sent from my iPad

On Apr 3, 2019, at 12:18 PM, Luca Bertagna <notifications@github.com<mailto:notifications@github.com>> wrote:


@bartgol commented on this pull request.

________________________________

In src/disc/stk/Albany_STKDiscretization.cpp<https://github.com/SNLComputation/Albany/pull/454#discussion_r271893360>:

> @@ -457,7 +457,7 @@ STKDiscretization(const Teuchos::RCP<Teuchos::ParameterList>&  discParams_,
  , discParams(discParams_)
  , interleavedOrdering(stkMeshStruct_->interleavedOrdering)
 {
-  // Nothing to be done here
+  this->updateMesh();


This is wrong. The method updateMesh calls virtual methods, which C++ does not allow (or, better, it allows it, but it won't do what you expected). This call got into the constructor somehow a while ago, I took it out again in a recent commit.

ikalash:
Update: LCM and SCOREC are ported to Thyra.  There is one caveat, namely I did not implement the Thyra version of ProjectIPtoNodalField for the case of a dynamic graph in unpetrify branch. This will cause SCOREC_ThermoMechanicalCan_timedep_thermomech_tpetra and SCOREC_BimetallicStrip_Traction_Tpetra tests to fail for now. Once all code is switched to static graph, the issue should resolve itself.  A temporary solution for now could be to turn off the IP to Nodal Field and Project IP to Nodal Field responses in the relevant tests.

ikalash:
> If the LCM tests pass (@ikalash says that they all do, except two SCOREC ones that need to be fixed separately), it's good to go on my end. Thanks for all the hard work!

Thanks @lxmota !  We still have to convert ATO, Aeras and AMP before merging it in, but we're working on it.

ikalash:
I wanted to clarify one thing about LCM being done: this does not include peridigm and moertel, which are in the LCM directory.  Those will be moved to tags, per the discussion in issues #458 and #394.

lxmota:
@ikalash Fine with me for LCM, but we probably need to ping @djlittl and @gahansen for Peridigm and moertel, respectively.

ikalash:
> @ikalash Fine with me for LCM, but we probably need to ping @djlittl and @gahansen for Peridigm and moertel, respectively.

@lxmota we already did this - see the discussion in the relevant issues.  Glen and Dave are OK with moving those packages to tags.

bartgol:
@ikalash The SCOREC tests now pass (it was a problem with zoltan-metis ID type mismatch, fixed by reconfiguring zoltan to use long long ids). There are still 5 tests failing:

- 2 in LCM due to the ProjectIPToNodalField response. I forgot what we decided: should we disable the test or remove that response?
- 3 in ATO. Not sure what happens: the first iteration of the optimization problem (including all the newton steps of the ME and the measure enforcement steps) is BFB. Then the second iteration seems to start from the same initial guess (the residual is the same as before), which I guess causes the issue. I'm going to ask Josh some help here.

I'd say we can merge it, so that we can avoid other stuff going in (with other conflicts), and unlock work that is pipelined to this PR. What do you think?

bartgol:
@lxmota when merging master, I saw quite a bit of work in adapt from you. I tried as best as I could to integrate in the new unified framework without losing functionalities. The tests seem to pass, but I can't be 100% sure I did not leave anything behind. If you have some tests outside Albany that you can/want to run to double check things still work correctly, that would be great.

ikalash:
Are the ProjectIPtoNodalField response tests SCOREC ones?  All the non-SCOREC ones should pass - I pushed a fix for this not long ago.

bartgol:
Yes, they are SCOREC-dependent tests. I guess I can try to replicate your fix. Can you point me to the commit?

ikalash:
Ok, those should fail.  The issue is that they rely on a dynamically allocated graph, which I did not implement in ProjectIPtoNF, as we will have to switch to the static graph soon.  I would just disable those responses for now, and make a note that we should re-enable them once we switch the code to the static graph.

lxmota:
@bartgol Recent changes in adapt from me are good to go in terms of the merge.

bartgol:
I removed the [WIP] prefix. This PR is ready for review and to be merged in master. Some ATO tests require a version of trilinos up to date. In particular, trilinos/Trilinos#5188 just went into develop, and it is required for some ATO tests to pass.

mperego:
@ikalash is ENABLE_ALBANY_EPETRE_EXE still useful after the unpetrification?

ikalash:
It is.  You can still build w/ or w/o Epetra, and there are a lot of ALBANY_EPETRA ifdefs throughout the code.

mperego:
Sorry, I wasn't cleae. There is an ENABLE_ALBANY_EPETRA_EXE and ENABLED_ALBANY_EPETRA, and it's not clear to me whether we need the former.

ikalash:
Oh I see - I actually did not realize there was a separate ENABLE_ALBANY_EPETRA configure option.  In this case, yes, the EXE one is obsolete and should be removed.

mperego:
@ikalash thanks for fixing it!

ikalash:
@mperego: sure. I need to clean up the doc dir to remove the old option.  I'll do that later today.

PR #494, opened 2019-06-14
CMake: limiting use of `include_directories', and adding info to installed cmake file

bartgol:
The installed cmake file now also provides a list of albany's tpls, as well
as information on their location.

Using include_directories at a top level directory is not advised, since
it automatically propagate to all the subdirectories added. On the other
hand, it is ok to use it at inner directories.
At the top level, prefer to use target-specific settings, unless you are
*absolutely certain* all targets will need that setting.


bartgol:
We need to hold off the merge of this. Things are quite complicated, due to include paths. I'll try to be short:

In `PHAL_FactoryTraits.hpp` we include a bunch of headers, some implemented in albanyLib, some in LCM. The LCM ones are picked up via a full path (relative from the `src` folder). However, the full path does not exist in the installation folder. Two solutions:

- Replicate the albany src tree in the `include` folder in the installation directory: this requires some cmake work, but would leave the src files unchanged. Perhaps is the "cleanest" solution (avoids potential clash of files in the installation folder)
- Include LCM files just by their name (without path): this requires some cmake work to make sure all the include directories are correctly set before any of the physics packages is processed (or else, as soon as `PHAL_FactoryTraits.hpp` is included, LCM's headers headers won't be found). Making LCM the first processed physics is just a temporary fix (what if tomorrow someone includes (directly or indirectly) `PHAL_FactoryTraits.hpp` inside a file in adapt or disc, which _must_ be processed before LCM?).

I tend to lean for the first solution. It would not impact (in any way) those that use Albany without installing it, so it should be the less disruptive.

If you have opinion against, let me know.

lxmota:
@bartgol I like the first solution too. I looked at the LCM headers. All but one are included for the handling of boundary conditions. They could be moved to other directories if necessary. But the idea of replicating the src hierarchy inside the include directory to avoid naming collisions is a good one. In the past, we had to rename files due to this very situation.

bartgol:
Most of our files are prepended with the physics package name (I think LCM is the main exception, maybe), just like Trilinios does (I believe for this very reason).

On the other hand, replicating the src tree may not be enough. When using albany as a library, you would also need to add all the subdirectories to the include directories of the project using albany. This is because, internally, most of Albany just includes file names without the full path (relative to src). This is also a habit we should try to get rid of (not only would solve this problem, but would make the code easier to navigate/read).

lxmota:
@bartgol Maybe we should try to remedy including headers without path, since that would fix all the problems? A bash script could traverse all the source files and replace occurrences of include with no path with full path.

bartgol:
Yeah, probably that's the easiest way. I'll do it and see if it works both when building inside albany and against albany.

bartgol:
Although, if you do this, you need to know _in advance_ all the include directories of all packages, before any file gets compiled. That's another thing I don't like of the current cmake logic: the main albany cmake file needs to know about internal organization of files in each package (so if a package adds a subdirectory, it must change also the main cmakelists file, which affects every user, even those not using the package).

bartgol:
I'm successfully building a small app that calls Albany. I am using _very limited_ stuff from Albany (so far only the solver factory, basically), so there may still be issues with some include paths, but we can fix them later.

I also sneaked a small mod to the build type initialization in this PR: as it was, we could no longer change the build type once set. When building _against_ Albany, one may want to run multiple Albany solvers, which may have different build types, so the user needs to switch build type when switching solver. I added this capability by confining the build type setting to a `Session` struct (with static members), much like it's done in Tpetra. This struct could be later be improved to integrate the handling of other 'global' features, like MPI or Kokkos.

PR #505, opened 2019-07-13
changes for phalanx mdfield refactor

rppawlo:
The following PR in Trilinos is a major refactor of the MDField internals. The DimTag object requirement has been removed. This PR contains changes needed for Albany when the Trilinos push goes through. The PR has passed the autotester and has been tested on cuda and against drekar, charon and empire as well. I'd like to push on Monday morning unless you would like to do more testing of Albany with this branch and the Trilinos PR. I did a simple build of Albany but don't have access to PUMA, so you might want to run more tests than the default build. Can anyone help with this?

https://github.com/trilinos/Trilinos/pull/5520

@ikalash @mperego @bartgol

### DIFF
--- a/src/PHAL_Dimension.cpp
+++ b/src/PHAL_Dimension.cpp
@@ -8,52 +8,22 @@
 
 #include "PHAL_Dimension.hpp"
 
-const char * Dim::name() const
-{ static const char n[] = "Dim" ; return n ; }
-const Dim & Dim::tag()
-{ static const Dim myself ; return myself ; }
+const char * DIM::name() const
+{ static const char n[] = "DIM" ; return n ; }
+const DIM & DIM::tag()
+{ static const DIM myself ; return myself ; }
 
-const char * VecDim::name() const
-{ static const char n[] = "VecDim" ; return n ; }
-const VecDim & VecDim::tag()
-{ static const VecDim myself ; return myself ; }
+const char * QP::name() const
+{ static const char n[] = "QP" ; return n ; }
+const QP & QP::tag()
+{ static const QP myself ; return myself ; }
 
-const char * LayerDim::name() const
-{ static const char n[] = "LayerDim" ; return n ; }
-const LayerDim & LayerDim::tag()
-{ static const LayerDim myself ; return myself ; }
+const char * NODE::name() const
+{ static const char n[] = "NODE" ; return n ; }
+const NODE & NODE::tag()
+{ static const NODE myself ; return myself ; }
 
-const char * QuadPoint::name() const
-{ static const char n[] = "QuadPoint" ; return n ; }
-const QuadPoint & QuadPoint::tag()
-{ static const QuadPoint myself ; return myself ; }
-
-const char * Node::name() const
-{ static const char n[] = "Node" ; return n ; }
-const Node & Node::tag()
-{ static const Node myself ; return myself ; }
-
-const char * Vertex::name() const
-{ static const char n[] = "Vertex" ; return n ; }
-const Vertex & Vertex::tag()
-{ static const Vertex myself ; return myself ; }
-
-const char * Point::name() const
-{ static const char n[] = "Point" ; return n ; }
-const Point & Point::tag()
-{ static const Point myself ; return myself ; }
-
-const char * Cell::name() const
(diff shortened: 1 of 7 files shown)


rppawlo:
ok. So I have a much simpler alternate fix. You can look at it in the branch "phalanx-mdfield-refactor-version2". If you prefer that I can close this branch and create a PR with that branch. I had some issues in other apps with the DimTags not working as field extents (mpl::for_each required an explicit move ctor on DimTags), but it seems that the way albany implements the DimTags, everything works fine. So I would probably suggest using the alternate branch.

bartgol:
Hi @rppawlo ,

Thanks for giving not only a heads up but also a working fix for this. So from what I see it seems that PHX now uses the meta-utility `is_extent<T>` to check whether a tag is indeed a valid extent tag, as well as other utilities to check fields' concepts (much like kokkos does for views). Is that correct? This seems to be something that would affect our compile-time behavior (relatively easy to catch if it shows up in some remote corner of the code that only gets compiled in some funky scenario). Any changes we should expect in the runtime behavior as well?

rppawlo:
@bartgol - yes. I need to identify extents because users can now specify layouts and devices in the template arguments (needed for empire). Runtimes were unchanged in my testing.  I thought it would significantly impact compile times and possibly library sizes. Here's some data I collected before deciding to push:

1. executable size:

Drekar main executable: 0.17% larger
old: 26MB, 27191224
new: 26MB, 27238112
Panzer example/adapters-stk/main driver: 0.31%
old: 19MB, 19091448
new: 19MB, 19150736
Albany: 1.4%
old: 3.7MB, 3775792
old: 3.7MB, 3829536
new: 

2. build times:

all phalanx (libs and tests) with make -j36, opt with debug symbols: 4.3% increase
--------------------------------------------------
old: 2m0s, 1m55s
new: 2m3s, 2m2s

panzer/disc-fe/src/evaluators directory (probably 50 evaluator objects):
ninja/make (on 36 core system)
old: 4m18s
new: 4m18s

touched panzer BasisValues2 and rebuilt just the single object file (-j1) for BasisValues2 (this has probably 20 to 30 different MDFields in the class so it is stressing the compilation of MDField): 1.4% increase
old: 2m19s
new: 2m21s

3. runtimes: Phalanx performance tests at scale for my workstation shows no difference in performance!

bartgol:
Sounds good to me. We'll take care of integrating your branch after thorough testing once the PR in trilinois goes in. Version 2 seems much cleaner, so we probably would go with that. Just a quick question: can we get away with empty DimTag structs now, like you did in your first branch or does Phalanx still need them to provide some methods like before (as derived from shards::ArrayDimTag)?

Thanks!

rppawlo:
phalanx does not use DimTags anymore so those could be empty structs like in the first PR. But it turns out that albany is explicitly using shards arrays outside of phalanx. For shards arrays you definitely need the DimTags. So the first PR separates the phalanx tags from shards tags so that, if you want, in the future you can remove the use of shards arrays from albany. The second pr just directly uses those DimTags with phalanx. The PR already passed and was pushed a few minutes ago.

rppawlo:
@bartgol - I made some minor changes today - deleting deprecated files based on the last round of commits from this morning. This required one more change to Albany. The patch is below. I'm not sure which branch you want to use. I can add this patch to each branch but don't want to do that until it goes into Trilinos. 

```
diff --git a/src/PHAL_Utilities.hpp b/src/PHAL_Utilities.hpp
index db26db7..671c79f 100644
--- a/src/PHAL_Utilities.hpp
+++ b/src/PHAL_Utilities.hpp
@@ -55,7 +55,7 @@ template<typename T>
 class MDFieldIterator {
 public:
   using array_type = typename PHX::MDField<T>::array_type;
-  using return_type = typename PHX::MDFieldTypeTraits<array_type>::return_type;
+  using return_type = typename PHX::MDFieldReturnType<array_type>::return_type;
   //! User ctor.
   explicit MDFieldIterator(PHX::MDField<T>& a);
   //! Increment efficiently.
```

bartgol:
@rppawlo thanks for the patch. I can add it when I merge your branch. Thanks!

bartgol:
I pushed directly to master the a commit with version2 and the patch. Closing this.

PR #509, opened 2019-07-18
[WIP] Adding support for Travis CI

bartgol:
** DO NOT MERGE THIS PR **

This PR tries to bring in travis CI support for Albany. 

This is work in progress. Force push will happen on this branch, so careful when you pull/push.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,119 @@
+language: cpp
+sudo: false
+dist: bionic
+
+cache:
+  directories:
+  - ${HOME}/trilinos/trilinos-install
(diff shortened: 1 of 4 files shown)


bartgol:
I don't think travisCI is an option for us. Building trilinos takes too long, and there's a time limit on builds for the free version. TravisCI _does_ offer the possibility of caching stuff and splitting your build into jobs, but I can't figure out a way to make it work anyways.

I think we should explore the idea of using SNL jenkins, with builds triggered not daily (like the nightlies) but by actions on the github server. Not sure if it works, but worth checking.

Closing.

Issue #519, opened 2019-09-08
Significant slowdown in Albany finite element assembly when running with Epetra


ikalash:
I was doing some large performance runs for some Land Ice Antarctic problems and this revealed that there has been a significant degradation of the Albany finite element assembly when running with Epetra - it is more than 2x slower than with Tpetra.  The problem seems to happen only for large enough test cases (4km, 2km, 1km AIS, not 8km AIS).  Here is some sample data:

  | ML | MueLu 
-- | -- | --
Alb. Total Time | 3196.6s | 1577.5s
NOX Lin Solve Time | 353.7s | 299.8s
Alb. Total Fill Time | 2676.7s | 1116.8s
Number Nonlinear Iters | 46 | 46
Number Linear Iters | 607 | 455

More can be found in the attached slides.  Recently @brian-kelley ran the same cases and got comparable timings for FEA with Epetra and Tpetra (see attached pdf).  He was running an Albany from Jan. 23 (b5f3930) prior to a lot of the refactors we did using Thyra.   He had a Trilinos from end of July (1f15cb62ca), so I'm pretty sure the issue is with Albany.  

The nightlies do not show a significant change but the problems run there are small.  Here's some data from the CDash site: 

AIS MueLu test today: http://cdash.sandia.gov/CDash-2-3-0/testDetails.php?test=4261731&build=80588
AIS MueLu test 1/21: http://cdash.sandia.gov/CDash-2-3-0/testDetails.php?test=4893660&build=88781
AIS ML test today: http://cdash.sandia.gov/CDash-2-3-0/testDetails.php?test=4893659&build=88781
AIS ML test 1/21: http://cdash.sandia.gov/CDash-2-3-0/testDetails.php?test=4261730&build=80588

@bartgol , @mperego , @jewatkins : any idea on what might have happened?  I don't really see how the Thyra refactor could have messed things up like this but maybe I'm wrong.  @jewatkins , could any of the memoization changes you did have had this result (e.g. memoization stopped happening for Epetra)?  

Another question is do we care about Epetra for ALI still?  I'm still testing things but it looks like Muelu is faster or comparable than ML.  If we will go with Tpetra moving further, we may not 
need to worry about this.

[Line Smoothing Report.pdf](https://github.com/SNLComputation/Albany/files/3587197/Line.Smoothing.Report.pdf)
[...]

PR #522, opened 2019-09-24
Attempt to fix slowdown in Epetra build (issue #519)

bartgol:
This mini PR attempts to remove (or limit) the slowdown observed in some Epetra test cases (see #519).

In particular, we avoid creating a Kokkos view of the local matrix at every call to evaluateFields during the Jacobian scattering. Instead, we preload the kokkos device data once from the Application side _before_ calling the evaluation method.

@ikalash Can you please try this fix with your timers and see if it improves a bit the performance?

### DIFF
--- a/src/Albany_Application.cpp
+++ b/src/Albany_Application.cpp
@@ -1477,6 +1477,14 @@ Application::computeGlobalJacobianImpl(
                   this, ps, explicit_scheme));
     }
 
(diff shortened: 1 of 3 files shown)


ikalash:
Thanks @bartgol !  I'm OK merging this in now, then checking the dashboard tomorrow, and running some of the test cases to see how much speedup we get once it's merged.  So go ahead and merge it in.

Issue #442, opened 2019-03-07
Deal with depracation of DynamicProfile option for Tpetra::CrsGraph constructor 


ikalash:
I received the following email from @kddevin about an upcoming change to Tpetra::CrsGraph, which will deprecate the DynamicProfile option.  I am posting it here to allow all Albany users to chime in, in particular, this change may be of interest to @bartgol and @ibaned .  

Roger, Irina and Mauro:

As you know, the Tpetra team is deprecating the DynamicProfile option for constructing CrsGraph and CrsMatrix.  We hope to have deprecation notices in Tpetra within the next few weeks.  We'd like your opinion on the best approach (from an application perspective) on carrying out the deprecation.

The challenge arises because we are both changing the default behavior of some constructors and removing constructors specifying dynamic-vs-static profile.

Currently, there exist constructors with the default behavior specified to be DynamicProfile; e.g.,

```
    CrsGraph(blah, blah, profileType = DynamicProfile)
```

Users calling
```
    CrsGraph(blah, blah)
```
get DynamicProfile currently.

After DynamicProfile is removed, the constructor will simply be
```
    CrsGraph(blah, blah);
```
but the behavior will be StaticProfile.  It isn't clear how to tell users of the current constructor with the default value that their behavior is changing.

The Tpetra team thought of a couple of approaches for the deprecation:  a two-phase approach and a one-step "rip off the bandage" approach.  They are detailed below.  For both, there will be a CMake option that allows users to enable/disable deprecated code during the transition period.

Please share your thoughts on these two approaches.  Do you have a preferred approach?  Do you see any pitfalls or advantages?

Thanks for your help.

Karen  (for the Tpetra team)

Two-phase deprecation:  first default type (which warns that behavior is
changing) and then the interface with profileType argument

  Phase one:  Behavior deprecation
```
  #ifdef ENABLE_DEPRECATED
    const profileType defaultProfileType = DynamicProfile;
    deprecated DynamicProfile enum value
  #else
    const profileType defaultProfileType = StaticProfile;
    Remove DynamicProfile enum value
  #endif
  Constructor CrsGraph(blah blah, profileType = defaultProfileType);
```

  At next release point:

  remove all the DynamicProfile code, including DynamicProfile enum value

  Phase two:  Interface deprecation
```
  #ifdef ENABLE_DEPRECATED
    deprecated Constructor CrsGraph(blah blah, profileType)
       (where this constructor ignores profileType because only StaticProfile remains)
[...]

PR #529, opened 2019-10-26
Build the CrsGraph using static allocation to comply with Tpetra changes

mperego:
Modified the Thyra_CrsMatrixFactory to construct the Crs graph with static profile (issue #442).
We first store all the entries of the graph locally, and at the fill Complete stage of the CrsMatrixFactory, we construct the Graph and fill it.
I did not see any change in performance. Given that we might move to FE_Graph and FE_Matrix I don't think we should worry too much about the implementation now.

Tested only with Demo PDEs and LandIce, but since the Factory is common it should work also for other packages in Albany.


ikalash:
Thank you for doing this @mperego !  I fixed a couple of compilation errors for certain builds.  I've also run ctest in some of my nightly builds.  Looks like ATO tests fail:

```
The following tests FAILED:
	240 - 2Matl_Homog_Epetra (Failed)
	241 - 2Matl_Homog_Tpetra (Failed)
	242 - Constraint_2D_adj_Epetra (Failed)
	243 - Constraint_2D_fd_Epetra (Failed)
	244 - DistParam_2D_Epetra (Failed)
	245 - DistParam_2D_Tpetra (Failed)
	246 - Filtered_2D_Epetra (Failed)
	247 - Filtered_2D_Tpetra (Failed)
	248 - FixedBlocks_Epetra (Failed)
	249 - FixedBlocks_Tpetra (Failed)
	250 - FixedField_Epetra (Failed)
	251 - Homogenize_2D_Epetra (Failed)
	252 - Homogenize_2D_Tpetra (Failed)
	253 - Mitchell_2D_Epetra (Failed)
	254 - Mitchell_2D_Tpetra (Failed)
	255 - Mitchell_3D_oc_Epetra (Failed)
	256 - Mitchell_3D_oc_Tpetra (Failed)
	257 - Mitchell_3D_oc_dp_Epetra (Failed)
	258 - Mitchell_3D_oc_dp_Tpetra (Failed)
	259 - MultiPhys_Homogenize_2D_Epetra (Failed)
	260 - MultiPhys_Homogenize_2D_Tpetra (Failed)
	261 - RegHeaviside_2D_Epetra (Failed)
	262 - RegHeaviside_2D_Tpetra (Failed)
	263 - RegHeaviside_3D_Epetra (Failed)
	264 - RegHeaviside_3D_Tpetra (Failed)
	265 - Restart_Epetra (Failed)
	266 - Restart_Tpetra (Failed)
	267 - Thermal_Mech_2D_Epetra (Failed)
	268 - Thermal_Mech_2D_Tpetra (Failed)
	269 - Thermal_MultiMech_3D_Epetra (Failed)
	270 - Thermal_MultiMech_3D_Tpetra (Failed)
```
Seems these issues only happen in parallel.   Here is how they die:

```
Number of parameter vectors  = 0
Number of response vectors  = 1
Number of distributed parameters vectors  = 0
[camobap:23970] *** Process received signal ***
[camobap:23970] Signal: Segmentation fault (11)
[camobap:23970] Signal code:  (128)
[camobap:23970] Failing at address: (nil)
[camobap:23970] [ 0] /usr/lib64/libpthread.so.0(+0x12c60)[0x7f4295d90c60]
[camobap:23970] [ 1] /home/ikalash/nightlyCDash/build/IKTAlbanyNoEpetra/src/libalbanyLib.so(_ZN6Albany21ThyraCrsMatrixFactory19insertGlobalIndicesElRKN7Teuchos9ArrayViewIKlEE+0xa2)[0x7f42b5204e32]
```
I think insertGlobalIndices is the culprit.  Can you please have a look?

mperego:
@ikalash I don't understand why your fixes are related to this branch. In particular the check on the GO index 
`ALBANY_EXPECT(indices[i]<=max_safe_gid, "Error! Input gids exceed Epetra_GO ranges.\n");`
is there also in master

ikalash:
@mperego : I deleted the assert b/c max_safe_gid is not defined in the scope of that assert, so the code failed to compile in debug mode.  I believe the code around there is Tpetra-based, not Epetra-based, so I removed the assert on grounds that I don't think it's relevant for non-Epetra code.  I did not compare to master.

mperego:
@ikalash Oh OK, I copied it in the wrong place. Thanks.

ikalash:
No problem.  Once ATO is fixed, we can go ahead and merge this in I think.

mperego:
I fixed it. Now ATO tests run fine on my workstation.
The issue was that ATO was inserting global IDs not owned by the calling processor which would give a negative local ID that in turn would be used as an index of an array.
Now we ignore the entries. Which is what CrsGraph does.

ikalash:
Thanks @mperego .  Is there an ifdef to turn off the dynamic graph deprecated code in Trilinos to check things with that option?

mperego:
There is a branch with all the deprecated code remove, but it seems there is a bug with it (see Trilinos issue trilinos/Trilinos/#6163). I'm working with them. I'd start merging this into master anyway.

src/utility/Albany_ThyraCrsMatrixFactory.hpp
```diff
@@ -43,7 +47,7 @@ struct ThyraCrsMatrixFactory {
 
   bool is_filled () const { return m_filled; }
 
-  bool is_static_profile () const { return m_static_profile; } 
+  bool is_static_profile () const { return true; }
```
bartgol: Are we using static profile always, even for Epetra? If so, then no point in keeping this method, we should purge it.

src/utility/Albany_ThyraCrsMatrixFactory.cpp
```diff
@@ -28,34 +25,35 @@ struct ThyraCrsMatrixFactory::Impl {
...
-                    const bool static_profile)
+                       const Teuchos::RCP<const Thyra_VectorSpace> range_vs,
+                       const int /*nonzeros_per_row*/)
  : m_graph(new Impl())
  , m_domain_vs(domain_vs)
  , m_range_vs(range_vs)
  , m_filled (false)
- , m_static_profile (static_profile) 
 {
   auto bt = Albany::build_type();
```
bartgol: This is orthogonal to the PR, but it came to my mind: should we check the type of the two vs rather than the build type? There's no reason to rely on the build type rather than the nature of the two vs, especially since the two vs will later be casted down to E or T maps...

src/utility/Albany_ThyraCrsMatrixFactory.cpp
```diff
@@ -28,34 +25,35 @@ struct ThyraCrsMatrixFactory::Impl {
...
-    auto t_range = getTpetraMap(range_vs);
-    if (static_profile) {
-      m_graph->t_graph = Teuchos::rcp(new Tpetra_CrsGraph(t_range,nonzeros_per_row,Tpetra::StaticProfile));
-    } else {
-      m_graph->t_graph = Teuchos::rcp(new Tpetra_CrsGraph(t_range,nonzeros_per_row,Tpetra::DynamicProfile));
-    }
+    t_range = getTpetraMap(range_vs);
+    t_local_graph.resize(t_range->getNodeNumElements());
+
+//    auto t_domain = getTpetraMap(domain_vs);
```
bartgol: IMHO, you should delete all the old code. Keep the files smaller.

src/utility/Albany_ThyraCrsMatrixFactory.cpp
```diff
@@ -131,30 +139,82 @@ void ThyraCrsMatrixFactory::insertGlobalIndices (const GO row, const Teuchos::Ar
...
+          e_indices[i++] = index;
+        auto row = e_range->GID(lrow);
+        m_graph->e_graph->InsertGlobalIndices(row,e_indices.size(),e_indices.getRawPtr());
+      }
+    }
+
     auto e_domain = getEpetraBlockMap(m_domain_vs);
-    auto e_range  = getEpetraBlockMap(m_range_vs);
     m_graph->e_graph->FillComplete(*e_domain,*e_range);
     m_graph->e_graph->OptimizeStorage();
```
bartgol: Don't forget to clear the local graph storage.

ikalash:
I'm running tests on this now - will approve once they complete.

src/utility/Albany_ThyraCrsMatrixFactory.hpp
mperego: OK, I'll purge it. Hope it's not used anywhere.

src/utility/Albany_ThyraCrsMatrixFactory.cpp
mperego: Are you are saying that we should try casting them into Epetra/Tpetra maps  and check whether they are null or not?

src/utility/Albany_ThyraCrsMatrixFactory.cpp
mperego: OK

src/utility/Albany_ThyraCrsMatrixFactory.cpp
mperego: I cleared the local graph storage. The only things I did not clear are the range maps.

mperego: Here https://github.com/SNLComputation/Albany/pull/529/files#diff-1cde7201d33ac3652684350f23422d55R189

bartgol: Thanks. For some reason I did not see it.

bartgol: The range maps are locals though, right? So they'll cleaned up upon return.

mperego: They are not local, otherwise one would have to create them each time insertGlobalIndices is called. I'll clear them.

mperego:
Thanks @ikalash. Tim Fuller is looking into the Tpetra issue. I'll wait a little bit to see if he understand what's the issue. If I don't hear from him the end of the day I'll merge this anyway.

src/utility/Albany_ThyraCrsMatrixFactory.cpp
bartgol: Ah, right. Well, they are not a storage overhead, assuming the input vector spaces were storing an epetra block map in their extra data section, which should always be the case (except for the replicated vs corresponding to multi-vector domains, but that's not the case here).

PR #531, opened 2019-11-01
WIP: MDField memoization for parameters

jewatkins:
This PR adds the ability to use MDField memoization for evaluators that depend on parameters. This is important for Albany Land Ice initialization because parameters that are used for optimization do not change at each model evaluation. Currently, these evaluators are being computed each time the model evaluator is called. Since Albany stores the results of these evaluators anyways, the evaluators only need to be recomputed when the parameters change.

I added a new input parameter to use this capability - `"Use MDField Memoization For Parameters"`. `"Use MDField Memoization"` works as before - i.e. it assumes parameters are always changing so it will recompute all dependent evaluators.

I also refactored the application `postRegSetup()` to be less verbose. As a consequence, the phalanx graph is now printed for ALL evaluation types. This seems to include a lot graphs of unused evaluation types. We may want to remove unused evaluation types or I could just suppress the output for all but a few.

Another critical change is that `loadWorksetBucketInfo()` now requires an "evaluation name". It was necessary to be able to pass the evaluation name while loading to ensure the correct `savedFields` were used for memoization. The evaluation name can be found based on evaluation type (e.g. `const std::string evalName = PHAL::evalName<EvalT>("FM", wsPhysIndex[ws]);` ). The name is only used for memoization so it may not be critical for some apps to pass the correct name.


jewatkins:
@mperego Shouldn't the analysis tests in `tests/small/LandIce/FO_GIS` use the AlbanyAnalysis executable? I don't see any nightly testing for the analysis path.

mperego:
> @mperego Shouldn't the analysis tests in `tests/small/LandIce/FO_GIS` use the AlbanyAnalysis executable? I don't see any nightly testing for the analysis path.

They should. I meant to do so since long time, but as of now all those tests check mainly sensitivities, but they don't run ROL. The only tests that use ROL are the SteadyHeatConstrainedOpt2D tests, however there is no regression tests (we would need to change AlbanyAnalysis to do so, I think, or print the resulting mesh)

jewatkins:
@mperego Can we print the resulting vector of parameters to a file and do a diff. That's pretty much what I've been doing to check the memoization. I guess one issue is that currently the runs are a bit long. Maybe we can lower the ROL iterations to 1?

mperego:
> @mperego Can we print the resulting vector of parameters to a file and do a diff. That's pretty much what I've been doing to check the memoization. I guess one issue is that currently the runs are a bit long. Maybe we can lower the ROL iterations to 1?

That sounds good. Maybe two iterations to be a little bit safer.

jewatkins:
Any idea how to print that vector to a file? Where does it print to std out?

@ikalash Is there an existing infrastructure to compare an output file to a gold file using ctest in Albany?

bartgol:
In `Main_Analysis.cpp` there is the line

```
status = slvrfctry.checkAnalysisTestResults(0, p);
```
Why can't we use this to check the parameters?

mperego:
> In `Main_Analysis.cpp` there is the line
> 
> ```
> status = slvrfctry.checkAnalysisTestResults(0, p);
> ```
> 
> Why can't we use this to check the parameters?

We can, that's what I was planning to do. But I think it needs some work.

jewatkins:
Hmm yeah it seems there's already some functionality in `Main_Analysis.cpp`. Is that used anywhere?

bartgol:
> @mperego Any idea how to print that vector to a file? Where does it print to std out?
> 
There is the function `writeMatrixMarket(X,prefix,counter)`, which writes a thyra object X (V,MV or LinOp) to the file prefix.counter.mm (or prefix.mm if counter is negative).

> @ikalash Is there an existing infrastructure to compare an output file to a gold file using ctest in Albany?

The file `runtest.cmake` in tests/small/ATO is a good starting point.

mperego:
Not sure, and even if it's used it might not be doing the right thing. I meant to work on it but last time I work on optimization I run out of my time. I'm going to go back to that soon, if you prefer to wait.

bartgol:
Ah, I think it's ok to use that for "small" parameters, since you have to list the target vector in the input file. For distributed parameters, we may cook up something similar, where the target parameter is stored in the mesh maybe? Or read from an exodus file on the fly (but we would have to check that exodus was compiled in).

jewatkins:
It looks like the "typical" way to compare files is to use `checkXTestResults` so I could probably try to get `checkAnalysisTestResults` to work.

jewatkins:
After looking through it, it seems the easiest thing would be to just check a handful of values as Luca suggested. I'll probably just add that for now.

@mperego 2 iterations take on the order of 15-60 seconds (longer than the other cases in the GIS folder). Is the okay or should we lower some tolerance?

bartgol:
> @mperego 2 iterations take on the order of 15-60 seconds (longer than the other cases in the GIS folder). Is the okay or should we lower some tolerance?

I think it's ok to have `Analysis` tests taking a bit longer than `Forward` tests. Now that we have labels, we can exclude tests with 'Analysis' label. On the other hand, 60s is a long time (the AIS tests take that long too though).

I'm debated. I'd like "long" tests to be in the order of 10-20 seconds. But I'm willing to accept a handful of exceptions, if necessary.

I guess a related question is: what do we want the tests to test? If we just want to make sure the infrastructure did not get broken (a "verification" kind of test), then coarse tolerances, and 1-2 iterations are enough. If we want to test the model (a "validation" kind of test), then we need longer tests.

jewatkins:
I think this would be more of an infrastructure test. I'm not sure if it's practical to have long validation tests with the current infrastructure that we have (unless we want to load all of our nightly testing by a considerable amount). I think it would probably take more than 2 iterations to get that type of information anyways so much more than 60s.

bartgol:
So 1 iteration you said is 15-60s. That's quite the spread. What does the spread have to do with? Does it depend on the machine or you simply have more tests, and some take 15s and some 60s? If it's the latter, given that we only want to test the infrastructure, I vote (if possible), to keep only the tests which take 30s tops.

jewatkins:
Yeah it's the latter (multiple tests) and it's based on 2 iterations. I'll try to shorten it. I think we can always add another test that's longer later and give it some macro or label so that it only runs in certain cases.

jewatkins:
I think this is ready to go. I'll wait until tomorrow to push to see if the analysis cases I added to master screwed up anything.

PR #563, opened 2020-02-23
TDM project added, test cases added, AMP project is removed

RuixiongHu:
Commit from RPI.

TDM (three-d manufacturing) project is added, which is in Albany/src/TDM.

CMakeList.txt, src/CMakeList.txt and problemFactory file are adjusted respectively.

The previous AMP project has been deleted.

Test cases has been added to /tests/large, which happens to replace the old AMP test folder.


ikalash:
@RuixiongHu : I just tried building your branch and noticed that none of the tests were being built.  I fixed that.  You still need to modify the CMakeLists.txt to create the tests you want to run.  Can you please do that?  Then I can turn TDM on in my nightly builds, so that your code will be tested nightly.

ikalash:
Also, does TDM depend on LCM and SCOREC?  If so, this logic should be modified: https://github.com/RuixiongHu/Albany/blob/master/CMakeLists.txt#L596-L606 . 

Perhaps you could add me a reviewer on the PR?

RuixiongHu:
Hello Irina,

Thanks a lot for the help, I do apologize for not being familiiar with this whole process. 

I really appreciate your help on setting up the test cases. I've checked the CMakeList.txt in every test folder and tried to finish the add_test{} routine, but it somewhat failed when I do 'make test'. But I could actually run each individual of them manually so I'm afraid I made some mistake in CMakeList.txt.  I followed the othere SOCREC test cases but no good. And I'm not sure if parallel cases are using the exact same way without providing number of partitions.

Manually they could be run as `mpirun -n <np> Albany/build/src/Albany <input file name>`

Could you please take a look at that and see what might be the problem? 
https://github.com/RuixiongHu/Albany/blob/master/tests/large/TDM/Additive3D/BarePlate/CMakeLists.txt#L16-L19

Please let me know any examples I could follow and I'll get the work done.

TDM does not depend on LCM. So I think that part should be fine.

Thanks a lot for the help!
Ruixiong

ikalash:
I assume you want to run the cases that have 'Parallel' in their name on 4 procs which would be the norm in Albany for the parallel nightly tests.  Can you please confirm?  And the tests that do not have 'Parallel' should be run on just 1 proc (in serial)?  Once I get a reply from you, I can fix the logic in your CMakeLists.txt.

ikalash:
I assume you want to run the cases that have 'Parallel' in their name on 4 procs which would be the norm in Albany for the parallel nightly tests.  Can you please confirm?  And the tests that do not have 'Parallel' should be run on just 1 proc (in serial)?  Once I get a reply from you, I can fix the logic in your CMakeLists.txt.

ikalash:
Also, I noticed that there is a missing input file, e.g. 
```
PUMI Input File Name: ../ModelAndMesh/Bare_plate_4P.smb	 
```
are not there, the code seg faults.  This is not very nice for users b/c they have no idea what went wrong.  You should put in some error checking to detect a mission file and throw an error, rather than have the code seg fault.

RuixiongHu:
Yeah, that is true. I'm fixing the CMakeLists.txt for mesh and models and
I'll push it to my repo when I'm done and I'll let you know when it's
ready. Thanks a lot for the hel[!

RuixiongHu:
That's the naming convention for the SCOREC mesh. It automatically append a
number after the name when starts to run. For a parallel mesh I specify the
name as xxx_4P.smb, it will starts to look for  xxx_4P0.smb  xxx_4P1.smb
xxx_4P2.smb until the last partition. I'll go run them manually again when
I'm ready and I'll let you know how it goes, thank you !

ikalash:
I have a fix for your tests running.  I will push it as soon as github lets me - seems it is having issues today (sigh...).  

A couple comments/requests regarding the tests:
- For me 
```
1/6 Test #1: TDM_BarePlate_Serial .............***Failed   43.38 sec
2/6 Test #2: TDM_BarePlate_Parallel ...........***Failed   15.35 sec
```
fail.  Can you please look into/fix these tests?
- Test 5 is WAY too long: 
```
5/6 Test #5: TDM_Depth_Serial .................   Passed  1639.96 sec
```
This is in a non-debug build.  This ran almost 30 minutes!  It should be shortened.  Perhaps just keep the pararallel version of the test? 
- Seems tests 5 and 6 (TDM_Depth_Serial and TDM_Depth_Parallel) have a lot of nonlinear solver stagnation:
```
: ||F|| = 4.007e-03  step = 1.000e+00  dx = 9.178e-07 (Failed!)
6: ||F|| = 4.007e-03  step = 1.000e+00  dx = 5.782e-07 (Failed!)
6: ||F|| = 4.007e-03  step = 1.000e+00  dx = 2.891e-07 (Failed!)
6: ||F|| = 4.007e-03  step = 1.000e+00  dx = 1.446e-07 (Failed!)
6: ||F|| = 4.007e-03  step = 1.000e+00  dx = 7.228e-08 (Failed!)
6: ||F|| = 4.007e-03  step = 1.000e+00  dx = 3.614e-08 (Failed!)
6: ||F|| = 4.007e-03  step = 1.000e+00  dx = 2.277e-08 (Failed!)
6: ||F|| = 4.007e-03  step = 1.000e+00  dx = 1.138e-08 (Failed!)
6: ||F|| = 4.007e-03  step = 1.000e+00  dx = 5.692e-09 (Failed!)
6: ||F|| = 4.007e-03  step = 1.000e+00  dx = 3.586e-09 (Failed!)
6: ||F|| = 4.007e-03  step = 1.000e+00  dx = 2.424e-09 (Failed!)
```
Usually this indicates some issue - I suggest you look into it.

ikalash:
> That's the naming convention for the SCOREC mesh. It automatically append a number after the name when starts to run. For a parallel mesh I specify the name as xxx_4P.smb, it will starts to look for xxx_4P0.smb xxx_4P1.smb xxx_4P2.smb until the last partition. I'll go run them manually again when I'm ready and I'll let you know how it goes, thank you !

I think you may have misunderstood my comment...  The *smb files are there and the code runs with them.  What I am saying is you get a seg fault if those files are not provided.  Instead, the code should throw an error saying that the run cannot continue b/c the *smb files are missing.  Let me know if that doesn't make sense.

ikalash:
> Yeah, that is true. I'm fixing the CMakeLists.txt for mesh and models

It's already there, no?  In MeshAndModels?  I had to put in some logic to configure those files but they should be accessible now when you run ctest.

RuixiongHu:
I just reduced the max step and right now if I run the serial one it will
take my machine about 10 secs to stop, I don't know if this is acceptable.
Please let me know if ctest speed are different from regular run and I'll
think about how to make it smaller.

I really appreciate the help, the modification I'm referring to is exactly
adding configure_file{} for mesh folders as what you did.

I just manually executed the test 1 & 2 locally and they do run, like all
others. I'll build again with the new CMakeFiles and see if any of them
passes, if not I'll look for the cause of test 1&2.

And for the last question, do you mean that I should be adding error loops
whener mesh files are not found for a tracking purpose, not that some of
the mesh files are missing?

RuixiongHu:
@ikalash Hello Irina, they now all passed on my machine, and the Depth_serial takes 5 secs for now, please let me know if you're still having any problem. 

I'll probably manipulate the Depth tests to make it a proper lenght but also run to a reasonable step to reflect the physics. And I'll commit once I'm done and I guess it would be good to go.

ikalash:
@RuixiongHu : I tested everything again and it looks good.  Please disregard my comments above about the BarePlate test failures, and about the exception throwing if mesh files are not found.  Seems this is working as expected now.

Regarding test times: I think it's OK if the tests are 2-3 mins long in a non-debug build.  The tests should actually catch bugs run in the code which means it may require running longer than 2-3 seconds.  So I think you can increase your tests a bit if it makes sense.  If not, that's good too.

RuixiongHu:
@ikalash Thanks a lot for the help. I just modified them and I think they're in the correcrt size. And I added extra reviewers. Please let me know if there's anything I need to do. Talk to you later.

PR #564, opened 2020-02-24
PHAL::SDirichletField class and fixes/additions to PHAL::SDirichlet

mperego:
List of changes:

1. Added `PHAL::SDirichletField` class to implement symmetric Dirichlet conditions in the case the
   dirichlet values are provided using a distributed field
2. Added/fixed Tangent adn DistParamDeriv implementation to `PAHL::SDirichlet` class
3. Performing a `preEvaluate` for the volume field, manager to allow computation
   of the dirichlet field (used e.g. in the enthalpy solver) before the dirichlet `preEvaluate` is run
4. modified two tests to use new symmetric dirichlet conditions

Issues:
When the dirichlet field is computed by a volume field manager, we need to `preEvaluate` it before calling the `PHAL::SDirichletField::preEvaluate`. However this can be expensive. In fact in the case of the `PHAL::LoadStateField` class, we are pre-evaluating all the fields needed by the volume field manager, also those not needed at the pre-evaluate time.
Not sure how to proceed. It would be nice to be able to pre-evaluate quantities in general. As an example one could pre-evaluate the total area of an ice sheet and use it to scale the responses. Another use is that one could pre-compute the lumped mass matrix and then use it to scale terms in a sort of finite-volume approach.

Another issue I have is that I get a weird error at the preEvaluate phase of the `PHAL::LoadStateField` at the Jacobian evaluation, when mesh depends on parameters.. @bartgol  any clue of what could be happening?

Thanks,
Mauro

### DIFF
--- a/src/Albany_Application.cpp
+++ b/src/Albany_Application.cpp
@@ -1680,6 +1680,9 @@ Application::computeGlobalJacobianImpl(
 
     dfm_set(workset, x, xdot, xdotdot, rc_mgr);
 
+    if(problem->useSDBCs() == true)
+      dfm->preEvaluate<EvalT>(workset);
+
     loadWorksetNodesetInfo(workset);
 
     if (scaleBCdofs == true) {
--- a/src/evaluators/bc/PHAL_SDirichlet_Def.hpp
+++ b/src/evaluators/bc/PHAL_SDirichlet_Def.hpp
@@ -86,6 +86,26 @@ SDirichlet<PHAL::AlbanyTraits::Jacobian, Traits>::SDirichlet(
 {
 }
 
+
+template <typename Traits>
+void
+SDirichlet<PHAL::AlbanyTraits::Jacobian, Traits>::preEvaluate(
+    typename Traits::EvalData dirichlet_workset)
+{
+  if(Teuchos::nonnull(dirichlet_workset.f)) {
+    Teuchos::RCP<Thyra_Vector const> x = dirichlet_workset.x;
+    Teuchos::ArrayRCP<ST>            x_view =
+        Teuchos::arcp_const_cast<ST>(Albany::getLocalData(x));
+    // Grab the vector of node GIDs for this Node Set ID
+    std::vector<std::vector<int>> const& ns_nodes =
+        dirichlet_workset.nodeSets->find(this->nodeSetID)->second;
+    for (size_t ns_node = 0; ns_node < ns_nodes.size(); ns_node++) {
+      int const dof = ns_nodes[ns_node][this->offset];
+      x_view[dof]   = this->value.val();
+    }
+  }
+}
+
 //
 //
 //
@@ -154,15 +174,8 @@ SDirichlet<PHAL::AlbanyTraits::Jacobian, Traits>::evaluateFields(
                     Teuchos::arcp_const_cast<ST>(Albany::getLocalData(x)) :
                     Teuchos::null;
 
-  Teuchos::Array<GO> global_index(1);
-  Teuchos::Array<LO> index(1);
-  Teuchos::Array<ST> entry(1);
   Teuchos::Array<ST> entries;
   Teuchos::Array<LO> indices;
-
-#if defined(ALBANY_LCM)
-  auto const& fixed_dofs = dirichlet_workset.fixed_dofs_;
-#endif
 
   this->set_row_and_col_is_dbc(dirichlet_workset);
 
@@ -192,7 +205,6 @@ SDirichlet<PHAL::AlbanyTraits::Jacobian, Traits>::evaluateFields(
     }
     Albany::setLocalRowValues(J, local_row, indices(), entries());
   }
-  return;
 }
 
 //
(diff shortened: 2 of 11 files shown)


src/evaluators/bc/PHAL_SDirichlet_Def.hpp
```diff
@@ -86,6 +86,26 @@ SDirichlet<PHAL::AlbanyTraits::Jacobian, Traits>::SDirichlet(
 {
 }
 
+
+template <typename Traits>
+void
+SDirichlet<PHAL::AlbanyTraits::Jacobian, Traits>::preEvaluate(
+    typename Traits::EvalData dirichlet_workset)
+{
+  if(Teuchos::nonnull(dirichlet_workset.f)) {
```
bartgol: How come tests passed without this pre-evaluate? Is this really necessary?

bartgol:
@mperego your dirichlet field cannot depend on x, obviously, so unless it depends on distributed parameters, couldn't you comput it offline?

src/evaluators/bc/PHAL_SDirichlet_Def.hpp
mperego: I can imagine that during the optimization loop one could ask to compute the Jacobian and residual at the same time. It does not seem to happen now.

mperego:
> @mperego your dirichlet field cannot depend on x, obviously, so unless it depends on distributed parameters, couldn't you comput it offline?

Yes I could. But as I mentioned in the description there are other cases where it could be handy to use pre-evaluate to compute stuff..

src/evaluators/bc/PHAL_SDirichlet_Def.hpp
bartgol: I see, thanks.

mperego:
I guess that what I could do is to write an evaluator to compute the temperature that only does a preEvaluate and not an evaluateFields, so that we avoid evaluating fields multiple times.

bartgol:
If the evaluate is the same as preEvaluate, then yes, you only need a preEvaluate, and you can leave the evaluate empty. After all, the preEvaluate is called on a per-workset basis, so everything will be fine.

ikalash:
@mperego : do you want me to test this on my machine in the same way the nightlies are run?

mperego:
> @mperego : do you want me to test this on my machine in the same way the nightlies are run?

No, thanks. I want to re-implement some evaluators first.

ikalash:
Ok, no problem.  Let me know when it's ready and I can test it.

src/Albany_Application.cpp
```diff
@@ -1358,6 +1358,15 @@ Application::computeGlobalResidualImpl(
            << ", computeGlobalResid workset.x = \n ";
       describe(workset.x.getConst(), *out, Teuchos::VERB_EXTREME);
 #endif
+
+      for (int ws = 0; ws < numWorksets; ws++) {
+        const std::string evalName = PHAL::evalName<EvalT>("FM", wsPhysIndex[ws]);
+        loadWorksetBucketInfo<EvalT>(workset, ws, evalName);
```
jewatkins: I think this will break the memoization for parameter dependencies because `evalName` needs to be a unique string (i.e. different from the `evalName` used in `evaluateFields`). If you want to use memoization for `preEvaluate` and `evaluateFields` you'll have to create unique names and add it to PHAL::Setup (e.g. `"preEvalFM"` and `"evalFM"`). If you don't really need memoization in `preEvaluate`, just pass it a dummy string so it doesn't affect `evaluateFields` (e.g. `evalName = "preEvaluate"`).

src/LandIce/evaluators/LandIce_PressureMeltingEnthalpy_Def.hpp
```diff
@@ -51,6 +51,17 @@ void PressureMeltingEnthalpy<EvalT,Traits,PressST,SurfTempST>::
...
 
+template<typename EvalT, typename Traits, typename PressST, typename SurfTempST>
+void PressureMeltingEnthalpy<EvalT,Traits,PressST,SurfTempST>::
+preEvaluate(typename Traits::EvalData d)
+{
+  const double powm6 = 1e-6; // [k^2], k=1000
+
+  for (std::size_t cell = 0; cell < d.numCells; ++cell)
+    for (std::size_t node = 0; node < numNodes; ++node)
+      surfaceEnthalpy(cell,node) = rho_i * c_i * ( std::min(surfaceTemp(cell,node),Tm) - T0 ) * powm6;
```
jewatkins: Are you computing this to use it somewhere else in volume FM `preEvaluate`? I don't see it but maybe it's not in this PR or you haven't implemented it yet.

src/evaluators/state/PHAL_LoadStateField_Def.hpp
```diff
@@ -40,6 +40,22 @@ void LoadStateFieldBase<EvalT, Traits, ScalarType>::postRegistrationSetup(typena
   if (d.memoizer_active()) memoizer.enable_memoizer();
 }
 
+// **********************************************************************
+template<typename EvalT, typename Traits, typename ScalarType>
+void LoadStateFieldBase<EvalT, Traits, ScalarType>::preEvaluate(typename Traits::EvalData workset)
+{
+  if (memoizer.have_saved_data(workset,this->evaluatedFields())) return;
```
jewatkins: If this get's called in `preEvaluate` with memoization on, `evaluateFields` will get completely skipped when using a single workset since `_prevWorksetIndex ` gets set to zero here. It looks like the code is the same so I think that's okay, just wanted to point that out.

src/Albany_Application.cpp
mperego: OK, thanks, I'll pass a dummy string for now.

src/LandIce/evaluators/LandIce_PressureMeltingEnthalpy_Def.hpp
mperego: This is scattered (with `PHAL::ScatterScalarNodalParameter`) and then used in the Dirichlet conditions.

src/evaluators/state/PHAL_LoadStateField_Def.hpp
mperego: I'm reverting this change.

mperego:
> Ok, no problem. Let me know when it's ready and I can test it.

@ikalash would you mind testing it now? Thanks!!

src/LandIce/evaluators/LandIce_PressureMeltingEnthalpy_Def.hpp
jewatkins: okay thanks

mperego:
I think it is pretty safe to merge. I removed the preEvaluate part for the volume field manager

PR #577, opened 2020-04-24
Cleanup Discretization Classes

gahansen:
Clean out some unused discretization functionality
in preparation for adding block data structure
support. This commit removes reference coordinate support
for adaptivity and other things needed for LCM adaptivity.


gahansen:
Thanks everyone! I just pushed the commit.

PR #581, opened 2020-04-28
Purged Dakota stuff

bartgol:
This simply removes all the dakota stuff, as well as the dakota-related tests.


bartgol:
I think this stuff ended in master anyways. Not sure with which PR. I'm closing this.

Issue #578, opened 2020-04-27
Unit test exe not found


ikalash:
I'm seeing that the new unit tests 
```
        179 - Albany_Serial_Unit_Test (Failed)
        180 - Albany_Parallel_Unit_Test (Failed)
```
are failing in the nightlies.  It looks like it's due to the executable tests/unit/evaluators/evaluator_unit_tester not being found.  @gahansen , do I need to enable something for the executable to build?

ikalash:
Update on this: it looks like there is a link error in building Albany that causes this issue:  https://sems-cdash-son.sandia.gov/cdash/viewBuildError.php?buildid=1131.  It's curious that the other tests just fine.  This happens in the skybridge build and the spack build.  At first I thought my changes to the skybridge scripts yesterday may have caused this issue, so I spent a bunch of time trying to investigate that.  Didn't find a problem, and then checked the spack build to find the same problem - so I think it's not the skybridge scripts.  I am verifying that right now by using scripts from last Friday on skybridge.

bartgol:
The link error is because the unit test only links albanyLib, and not LandIce. That's fine if LandIce is not enabled, but if it is, all the landice references in the various factories are broken.

This should be fixed together with #579.

bartgol:
Note: a quick fix would be to link `$ALBANY_LIBRARIES`, which in LandIce builds includes also LandIce.

ikalash:
Thanks @bartgol .  I'm confused as to why the issue only showed up in the spack and skybridge nightlies for now, not the others.  I thought the unit tests are on by default for now.  I can wait for the fix to #579.  Maybe I'd turn off the unit tests in the problematic builds for now to avoid the problem.

bartgol:
I'm not sure why it doesn't show up in all builds that include LandIce. I only know that if you don't enable LandIce, the error is quickly explained.

ikalash:
The builds where the problem shows up definitely have LandIce enabled.  I don't get this...

bartgol:
Sorry, it's the other way around. If LandIce _is_ enabled, you should expect the link error. If it's not enabled, then I have no answer.

bartgol:
Nvm, I fixed the undef ref to LandIce stuff, and now it complains about stk stuff undef refs.

Again, this is due to albanyLib depending on albanySTK (ugh), so it will be fixed when circular deps are eliminated. For now, we should simply link all albany libs, as we do for the other executables.

bartgol:
[...]

PR #584, opened 2020-04-29
Remove circular dependencies

### LINKED ISSUE #579: Small refactor to remove inverted dependencies in Albany
I thought about the issue of circular dependencies in Albany. The main issue is with the various factories (or factory-like) classes in Albany, which require knowledge about the subpackages.

After thinking about it, I think we should do something like this:

1. For each factory type XYZ (e.g., ProblemFactory), create a base abstract class (e.g. ProblemFactoryBase), and have only the "generic" Albany objects created in the 'default' albany factory.
2. Packages can have their own factory, which inherits from the base one.
3. Most of the creation happens in the application. So let's make the factories 'settable' in the application, with the default one always present.
4. In the Main_*.cpp file we add all factories, including LandIce ones. This is ok, since Main_*.cpp is _not_ in AlbanyLib, so it can depend on LandIce too.
5. Similar to 4, the mpas interface has to set LandIce factories in the application
6. Probably need to modify a tad also the SolverFactory (which is probably good anyways, especially since we pruned some stuff and maybe the solver factory can be cleaned a bit).

@mperego @gahansen @ikalash @jewatkins I'm planning to do something on this side, if you're ok with it. Should be pretty quick.

bartgol:
This PR does quite a few things, to be honest. Here's a summary of the main things:

1) Most of core Albany libs are condensed into albanyLib: albanyAdapt, albanySTK, albanySTKRebalance. These libs were so much intertwined with the core albany funcitonalities that it was impossible to make one lib depend on another without creating circular dependencies. For isntance, adapt contains the SolutionManager class, which is essential in Albany; on the other hand, adapt uses a lot of core features of Albany.

2) STK is no longer 'optional'. Since we purged SCOREC support, now STK is the only discretization possible. It made no sense to have it optional. If we ever (re)introduce other discretizations, we can come back and change this.

3) Refinemend capabilities have been pruned. They depended on the stk_adapt subpackage in stk in Trilinos, which seems to be gone.

4) The only 'optional' libs in Albany are now albanyPDEs (the demo problems) and LandIce. In order for main Albany to build without depending on these (when enabled), I abstracted the problem factory design. Each optional package now should implement its problem factory (see src/problems/Albany_ProblemFactory.hpp for an example). This factory needs to be 'enabled' before the Application is created. For Albany/AlbanyAnalysis executables (as well as cism/mpas interfaces), this is already handled for the current supported packages (demo and LandIce). If you add a new package or a new executable, you need to follow the example in utility/Albany_FactoriesHelpers.hpp (or directly use/edit that routine).

5) Purge Dakota support in the code.

This closes #579 and #578.


bartgol:
Thanks for checking that. I forgot to say I couldn't check those, and was gonna see what happened in the nightlies. But since you caught it now, I'm going to fix it.

bartgol:
Uhm, I can't find the place where those functions are called. As a matter of fact, createAndGetAlbanyApp is not even present anymore.

Was that a fresh build of Albany or was the build dir old? Perhaps some old object files polluted the build?

ikalash:
Sorry, yes, I had old code in the directory when I was building.  When I build from scratch I get the following error: 

```
[ 98%] Building CXX object src/LandIce/CMakeFiles/mpasInterface.dir/interface_with_mpas/Interface.cpp.o
In file included from /home/ikalash/nightlyCDash/repos/Albany/src/LandIce/interface_with_cism/ali_driver.cpp:19:
/home/ikalash/nightlyCDash/repos/Albany/src/LandIce/interface_with_cism/ali_driver.H:14:10: fatal error: DyCoreModelRegistry.H: No such file or directory
   14 | #include "DyCoreModelRegistry.H"
      |          ^~~~~~~~~~~~~~~~~~~~~~~
compilation terminated.
[ 99%] Linking CXX shared library libcismInterface.so
g++: error: CMakeFiles/cismInterface.dir/interface_with_cism/ali_driver.cpp.o: No such file or directory
[ 99%] Built target cismInterface
[ 99%] Linking CXX shared library libmpasInterface.so
[ 99%] Built target mpasInterface
Scanning dependencies of target Albany
Scanning dependencies of target AlbanyAnalysis
gmake[2]: *** No rule to make target 'src/LandIce/libcismInterface.so', needed by 'src/Albany'.  Stop.
gmake[2]: *** Waiting for unfinished jobs....
gmake[2]: *** No rule to make target 'src/LandIce/libcismInterface.so', needed by 'src/AlbanyAnalysis'.  Stop.
gmake[2]: *** Waiting for unfinished jobs....
[ 99%] Building CXX object src/CMakeFiles/Albany.dir/Main_Solve.cpp.o
[ 99%] Building CXX object src/CMakeFiles/AlbanyAnalysis.dir/Main_Analysis.cpp.o
```
The missing header DyCoreModelRegistry.H lives in CISM, which is a separate directory.  I realize that this is not nice - having includes from another directory - but that was the way I made the CISM-Albany interface work.  I had the logic set before to have the include dirs set so that everything worked correctly.  I'm guessing you removed some of this logic in your cleanup?

bartgol:
Ah, right, I forgot to pass the cism include dir to the target. Can you try again now?

ikalash:
There is a missing ifdef in src/LandIce/interface_with_cism/CMakeLists.txt.  What is the top-most ifdef supposed to be?

bartgol:
Whoops, sorry, leftover from copy/paste.

It should work now...maybe?

bartgol:
Whoops, had to do another commit. Sorry!!

bartgol:
I did a force push, cause the last 3 commits were all cmake error fixes, so I squashed them all in one. This may force you to reset --hard your branch on the remote one.

ikalash:
There are still issues: 

```
And CMake says...:
-- Configuring done
CMake Error at src/LandIce/interface_with_cism/CMakeLists.txt:14 (add_library):
  Cannot find source file:

    evaluators/LandIce_CismSurfaceGradFO.cpp

  Tried extensions .c .C .c++ .cc .cpp .cxx .cu .m .M .mm .h .hh .h++ .hm
  .hpp .hxx .in .txx
```
Do you want me to give you instructions for how to build CISM-Albany so that you can troubleshoot the remaining problems on your side?

bartgol:
Yeah, I think that's probably easier, so we don't go back and forth here, wasting time for both of us. If you can send me an email, that'd be great. I feel like you already sent me instructions in the past, but I can't find them.

ikalash:
Thanks @bartgol .  I just sent you build instructions.  I tried to resolve the issue above last night, but it led to more issues, and I decided it's probably best for you to try to resolve them (sorry!).  

BTW, I noticed my epetra-free build had an error last night coming from ali_driver.cpp: https://sems-cdash-son.sandia.gov/cdash/viewBuildError.php?buildid=1211 .  Is this due to some of your recent changes?  It seems "Build Type" may not be known in the case the code is not built w/ Epetra (which probably makes sense).

bartgol:
Thanks. I think I'm responsible for virtually all those errors.

ikalash:
Ok, good (I guess... :)).  You should be able to reproduce them with my instructions on  how to build Albany with CISM.

mperego:
I think we can remove the "PARAMETERS_DEPEND_ON_SOLUTION" option. It's never used. I added that mainly for backward compatibility because that was the default configuration (i.e. having parameter to be  ScalarT). We do not need it any longer. It's super quick fix. Luca, can you add it to this PR If nobody objects?

bartgol:
I can stuff it in, yeah.

bartgol:
Ok, I fixed the cism interface build, and removed the option ENABLE_PARAMETERS_DEPEND_ON_SOLUTION.

@ikalash The build should work now (it works on my machine), but I cannot run the cism tests. I tried to hack the run_tests.sh script in the CismAlbany folder, but I can't get it to work (I don't have matlab, and even trying replacing it with octave wouldn't work). Could you please launch the tests on camobap and see if everything works?

ikalash:
Thanks, @bartgol .  Unfortunately it looks like the static libs Albany build is broken.  I get this error on Cori: 

```
-- Albany Executables: Albany; AlbanyAnalysis
CMake Error at src/CMakeLists.txt:806 (set_target_properties):
  set_target_properties called with illegal arguments, maybe missing a
  PROPERTIES specifier?


CMake Error at src/CMakeLists.txt:806 (set_target_properties):
  set_target_properties called with illegal arguments, maybe missing a
  PROPERTIES specifier?
```
That's where the CISM-Albany tests are run.

bartgol:
Ah, I had not tested the static exec case. It was an easy fix. I just built static executables, and it went through. I never built trilinos with static libs, so I can't quickly check the static libs too.

Does CISM not build with shared libs?

ikalash:
CISM builds with shared libs, but on certain supercomputers like Cori, dynamically linked executables do not run.  So all codes need to be built with static libs there.  We want to keep supporting that option in Albany, therefore.

bartgol:
Got it. Thanks.

Btw, I thought my built went through, but I looked at the wrong window. I still get a link error. I'll fix and push another commit.

bartgol:
I _think_ my link error stem from having a shared lilb trilinos installation, mixed with static albany executables. Is your trilinos installation with static libs? If so, would you mind trying again?

ikalash:
Hmmm, you're right, the static Albany build is OK.  Not sure what happened yesterday, though I _was_ building from scratch using adapted scripts rather than using the nightly test scripts.  Unfortunately there is no an undefined reference error when I build CISM-Albany on top of Albany.  Here's the error: 
```
LandIce_StokesFO.cpp:(.text._ZN7LandIce12StokesFOBase27constructVelocityEvaluatorsIN4PHAL12AlbanyTraits8ResidualEEEvRN3PHX12FieldManagerIS3_EERKN6Albany15MeshSpecsStructERNS9_12StateManagerENS9_18FieldManagerChoiceE[_ZN7LandIce12StokesFOBase27constructVelocityEvaluatorsIN4PHAL12AlbanyTraits8ResidualEEEvRN3PHX12FieldManagerIS3_EERKN6Albany15MeshSpecsStructERNS9_12StateManagerENS9_18FieldManagerChoiceE]+0x4a8e): undefined reference to `LandIce::CismSurfaceGradFO<PHAL::AlbanyTraits::Residual, PHAL::AlbanyTraits>::CismSurfaceGradFO(Teuchos::ParameterList const&, Teuchos::RCP<Albany::Layouts> const&)'
/usr/bin/ld: /project/projectdirs/piscees/nightlyCoriCDash/build/AlbanyInstall/lib/libLandIce.a(LandIce_StokesFO.cpp.o): in function `void LandIce::StokesFOBase::constructVelocityEvaluators<PHAL::AlbanyTraits::Jacobian>(PHX::FieldManager<PHAL::AlbanyTraits>&, Albany::MeshSpecsStruct const&, Albany::StateManager&, Albany::FieldManagerChoice)':
```
I think you rearranged some stuff having to do with CismSurfaceGradFO, maybe?  Attached is the full log if you'd like to have a look at it.  I can also provide instructions for how to reproduce this on Cori.  It's the only place I still have a CALI build these days.

[nightly_log_coriCismAlbanyBuild.txt](https://github.com/SNLComputation/Albany/files/4566725/nightly_log_coriCismAlbanyBuild.txt)

ikalash:
I did some additional testing, and the error in the CALI build doesn't happen on my workstations, only on Cori, FWIW.

bartgol:
Ah, good catch. That link error is indeed a leftover of circular dependencies. I'll push a fix right away.

bartgol:
@ikalash Ok, I think I fixed the issue.

Btw, I had to rewrite git history, cause there were some repeated (and conflicted) mods with master (I think you ported the rythmos changes to the branch in order to compile?). I also had to comment some rythmos tests, cause their input files were gone.

Suggestion: can we merge this before doing other changes in master? This branch has a wide reach, and may conflict with other changes we may push to master.

ikalash:
I checked things on Cori and everything is clean now.  Feel free to go ahead and merge in.  I think I caused some of the conflicts you mention - sorry about that!

bartgol:
Great! No worries, I am the last one who should say anything about messing up git history... =)

I'll merge right away.

PR #588, opened 2020-04-30
LandIce: remove basal friction type GAL_PROJ_EXP_GIVEN_FIELD

mperego:
That type (that I created) is a bit misleading. So I added instead the option "Compute Exponent Of Given Field At Quadrature Points".

### DIFF
--- a/src/LandIce/evaluators/LandIce_BasalFrictionCoefficient_Def.hpp
+++ b/src/LandIce/evaluators/LandIce_BasalFrictionCoefficient_Def.hpp
@@ -40,6 +40,9 @@ BasalFrictionCoefficient (const Teuchos::ParameterList& p,
   Teuchos::ParameterList& beta_list = *p.get<Teuchos::ParameterList*>("Parameter List");
   zero_on_floating = beta_list.get<bool> ("Zero Beta On Floating Ice", false);
 
+  //whether to first interpolate the given field and then exponetiate it (on quad points) or the other way around.
+  interpolate_then_exponentiate = beta_list.get<bool> ("Interpolate Then Exponentiate Given Field", true); 
+
   std::string betaType = util::upper_case((beta_list.isParameter("Type") ? beta_list.get<std::string>("Type") : "Given Field"));
 
   is_side_equation = p.isParameter("Side Set Name");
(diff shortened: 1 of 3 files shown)


src/LandIce/evaluators/LandIce_BasalFrictionCoefficient_Def.hpp
```diff
@@ -405,33 +404,33 @@ evaluateFieldsSide (typename Traits::EvalData workset, ScalarT mu, ScalarT lambd
         break;
 
       case EXP_GIVEN_FIELD:
-        if (is_given_field_param) {
-          for (int ipt=0; ipt<dim; ++ipt) {
-            beta(cell,side,ipt) = std::exp(given_field_param(cell,side,ipt));
+        if(nodal || compute_exponential_at_quadrature_points) {
```
bartgol, line 407: I don't understand. What's the difference between asking 'Nodal' or 'Compute Exponent Of Given Field At Quadrature Points'?

mperego: "nodal" refers to the layout of beta. If the layout is nodal, then we don't consider the "Compute Of Given Field At Quadrature Points" flag because no quad point is involved.
When the layout of beta is not nodal, then we need to compute the exponent of the given field at the quad points.  We can either 1. compute the exponent of the given field at the nodes and then interpolate it at quad points, or 2. first interpolate the given field at quad points and then take the exponent of it.  If "Compute Exponent Of Given Field at Quadrature Points" is true we do 2.

bartgol: Isn't that equivalent to interpolate the given field and then compute the exponent?

bartgol: It seems like we could achieve the same result by piping a DOFInterpolation evaluator with the BasalFrictionCoefficient one.

If we want to keep this 'combo', maybe we could use a more verbose name, which stresses that we are doing a "combo" operation. Something like "Interpolate Then Exponentiate" and "Exponentiate Then Interpolate".

mperego: We could, but then we need to add an extra evaluator for the exponential. I think it makes sense to keep it this way for now. If you prefer I can rename the boolean option "Interpolate Then Exponentiate Given Field"

bartgol: I meant we could leave the exp in this evaluator, and handle the interpolation separately.

But ok, we can keep a combo in here. Yes, that option name seems more verbose to me. Thanks.

mperego: Great, thanks. I'll do the change and then I'll merge it in.

PR #591, opened 2020-05-11
LandIce: cleaning / adding new evaluators

mperego:
This PR includes:
1. Added new evaluator that computes the flux divergence with a finite volume approach. 
2. Removed some enthalpy related evaluators no longer used.
3. Performance improvement when extruded columns dofs are coupled in the evaluation of responses. In that case we need to use larger number of FAD dervatives, but this is only needed when evaluating the response, not when computing the Jacobian in a nonlinear solve.
4. generalized the evaluator for gathering vertically averaged velocity, to a gather a vertically "contracted" solution. We can support different contractions. At the moment we have Average and Sum.
5. Added a new mesh for Humboldt glacier. It's a bit larger than usual tests but we have all the fields we need and it can be used to test different problems.


src/LandIce/problems/LandIce_StokesFOThermoCoupled.hpp
```diff
@@ -405,7 +382,7 @@ constructEnthalpyEvaluators (PHX::FieldManager<PHAL::AlbanyTraits>& fm0,
  *   p->set<std::string>("Gradient BF Side Name", Albany::grad_bf_name + " "+basalSideName);
  *   p->set<std::string>("Weighted Measure Name", Albany::weighted_measure_name + " "+basalSideName);
  *   p->set<std::string>("Velocity Side QP Variable Name", dof_names[0] + "_" + basalSideName);
- *   p->set<std::string>("Vertical Velocity Side QP Variable Name", "W");
+ *   p->set<std::string>("Vertical Velocity Side QP Variable Name", dof_names[0]);
```
bartgol, line 385: `dof_names[0]` should be `Velocity'. Is this a bug? Shouldn't it maybe be `dof_names[1]`? If 'Velocity' is correct, maybe the param name should not be `Vertical Velocity ...`, but `Velocity ...`.

src/LandIce/problems/LandIce_StokesFOThermoCoupled.hpp
```diff
@@ -489,7 +466,7 @@ constructEnthalpyEvaluators (PHX::FieldManager<PHAL::AlbanyTraits>& fm0,
   p->set<std::string>("Diff Enthalpy Variable Name", "Diff Enth");
   p->set<std::string>("Velocity QP Variable Name", dof_names[0]);
   p->set<std::string>("Velocity Gradient QP Variable Name", dof_names[0] + " Gradient");
-  p->set<std::string>("Vertical Velocity QP Variable Name", "W");
+  p->set<std::string>("Vertical Velocity QP Variable Name", dof_names[1]);
```
bartgol, line 469: See comment above. I think this is correct, which makes me thing the above one is wrong.

src/LandIce/problems/LandIce_StokesFOThermoCoupled.hpp
mperego: you are right. I'll fix it.

mperego:
> There's a comment about a dubious variable name.
> Also, I might be wrong, but I think you may have to modify the call to getDerivativeDimension in Application::computeGlobalTangent, to pass `true` to the `responseEvaluation` argument.

No, that one has a different signature. The one I modified is expecting a `MeshSpecsStruct` as the second parameter.

src/LandIce/problems/LandIce_StokesFOThermoCoupled.hpp
mperego: fixed.

PR #592, opened 2020-05-12
Switching to FE Crs matrix for the Jacobian

bartgol:
This PR makes Albany use Epetra_FECrsMatrix and Tpetra::FECrsMatrix for the Jacobian.

This is work in progress.


jewatkins:
This looks good, is there a way to remove the changes from Mauro's branch for future reference?

BTW, memory tracking should start today so we should be able to see the change on CPU and GPU. I'm just going to store max memory allocated by kokkos and host high water mark for now.

mperego:
> This looks good, is there a way to remove the changes from Mauro's branch for future reference?
I think he just need to rebase to master.

bartgol:
> This looks good, is there a way to remove the changes from Mauro's branch for future reference?
 
Yeah, as Mauro said, I need to rebase again.

> BTW, memory tracking should start today so we should be able to see the change on CPU and GPU. I'm just going to store max memory allocated by kokkos and host high water mark for now.

Great, thanks.

bartgol:
Am at a loss. I ran the SteadyHead1D test, where I modified the input file so that we have only 2 elements (to make debug easier and faster). I ran in serial and in parallel. The two start diverging already at the first newton iteration. The initial residual is the same, and so are the initial jacobian (I printed them). The matrices are the same up to a permutation of how the values are stored in a row (which should not change the solvers behavior). I used both ifpack2 and MueLu (1 level, KLU on coarse level, so basically a direct method).

I would be fine with a little divergence between serial and parallel (even though with 3 rows in the matrix, i would not expect it), but here serial converges (F<1e-9) in 5 newton iterations, while parallel after 10 iteration still has F~1e-3. This is wrong.

Any idea of how I could track down the error? I checked the range/domain vector spaces of the jacobian, and they are correct (meaning they match the residual and solution vector spaces respectively (which also happen to be the same)).

Note: I also tried to deep copy the jacobian into a Tpetra_CrsMatrix (not a FECrs) right after computing it, to make sure that a FECrs matrix didn't mess up the solvers.

jewatkins:
What values do you get with master (serial/parallel)?

mperego:
Um.. but if the columns of the jacobian are permuted, then the rhs entries should be permuted as well, no?

mperego:
So row, col, domain and range maps  are all the same map? In the test in Dof-Manager the col map is in general different than the row map, but maybe you built them in a different way.

mperego:
> So row, col, domain and range maps are all the same map? In the test in Dof-Manager the col map is in general different than the row map, but maybe you built them in a different way.

Actually no, never mind... I think that's true when the fe matrix is being filled.

bartgol:
> Um.. but if the columns of the jacobian are permuted, then the rhs entries should be permuted as well, no?

The pairs (colGID, value) are the same. They just appear swapped in the storage. So the mathematical object is the same, it only changes the order in which entries are stored.

bartgol:
> What values do you get with master (serial/parallel)?

Good point, I will check.

mperego:
> > Um.. but if the columns of the jacobian are permuted, then the rhs entries should be permuted as well, no?
> 
> The pairs (colGID, value) are the same. They just appear swapped in the storage. So the mathematical object is the same, it only changes the order in which entries are stored.

Ah OK. and that's true also for the rhs, I assume. Mystery!

bartgol:
> Ah OK. and that's true also for the rhs, I assume. Mystery!

I compare the range of the Jac with the residual vector space. Internally, Thyra calls sameAs from Tpetra, which I think also checks that the gid appear in the same order, so yeah, they have the very same map (as expected, since both are created from the vector space built in the discretization).

bartgol:
> What values do you get with master (serial/parallel)?

So, I checked, and in master serial and parallel give the very same exact numbers, which, btw, coincide with what I get in serial mode in the branch (which is good).

mperego:
If you force NOX to stop after 1 iteration, would the solutions obtained with master and your branch differ?

bartgol:
All this obsession with the jacobian, and now I see that the problem may be with the residual. After the 1st Newton iteration, the value of X is the same (serial vs parallel), but the corresponding computed residual is different.

bartgol:
I think I found it. The overlapped GIDs appear twice in the overlapped vs. I'm assuming this causes issues in the import/export operations in Tpetra/Epetra.

bartgol:
I'm running the tests now. If all tests pass, we can merge this, and see what happens in the nightlies and performance/memory tests tonight.

bartgol:
I still have quite a few failures. I don't think I'll be done with this today. Hopefully tomorrow.

bartgol:
Ok, Epetra turns out to be a can of worms. The main issue is that Epetra_FECrsMatrix does _not_ expose anything related to the overlapped nodes, which has two bad consequences:

1) we cannot use sumIntoLocalRowValues easily. We cannot go from local_row to global_row. This is somewhat 'easy' to solve: I rolled an EpetraFECrsMatrix class, which also stores the overlapped row map, so that we can convert everything back to global, and then call Epetra_FECrsMatrix method sumIntoGlobalValues.
2) this is the hardest: we cannot extract a "view" to the local matrix when doing assembly. This is, again, because the FE matrix in Epetra does not expose the nonlocal stuff. If you grab the internal data, you only get what the base CrsMatrix class returns, which are the rows of the owned row map.

For now, I'm getting around it by prohibiting kokkos-like assembly for the Epetra case, meaning that we assemble "the phalanx/thyra" way, calling the thyra utility `addToLocalRowValues`. However, I'm not really happy about it...

jewatkins:
I think that's okay. I don't think we need the kokkos-like assembly for the Jacobian w/ Epetra anyways since it's not used w/ cuda.

bartgol:
The only thing you lose is that you cannot use openmp as backend (or, better, that you don't get any boost during assembly). But given that omp doesn't buy you a big speedup anyways (perhaps 20% on a good day), you're right: it's fine.

bartgol:
I put in a PR in Trilinos (trilinos/Trilnos#7349). Although not strictly essential (one could not use nonlocal matrix in Epetra_FECrsMatrix), I think I want to wait for that PR to go in before this one can be merged. That is because I like to use the nonlocal matrix for our case (it is faster than the alternative (std vectors for nonlocal data) in case of multiple uses of the matrix).

Meanwhile, there are still a handful of failures that I need to fix anyways.

bartgol:
Turns out that our code kind of strongly depends on a particular ordering of the column map of the matrices: the gids of the overlap map must always come first in the column map. That is, the first size(overlapDomainMap) gids in the column map must coincide with the gids in the overlap domain map (in the same order).

This was always true when we allocated an overlap jacobian, since both Tpetra and Epetra guarantee that the column map contains the domain map as its first chunk. However, this is no longer true for Epetra_FECrsMatrix, while it still holds for Tpetra::FECrsMatrix (which is basically the same as the old overlapped jacobian, with some tricks to get the non-overlapped out of it).

The reason is that an Epetra_FECrsMatrix is created with _non-overlapped_ range/domain maps. So the only guarantee we get is that the GIDs of the non-overlapped domain map come first in the ColMap. But that's not enough for our insertion function addToLocalRowValues to correctly retrieve the column gids from the local indices. In fact, the (local) column indices passed to function come from the wsElemNodeEqId array, which is filled with the overlap map. But the LIDs of the shared nodes in the overlap map may not coincide with those of the ColMap of the matrix.

Bottom line: this got a bit more complicated. Now I have to manually build the column map for the Crs graph, ensuring the overlapped distribution GIDs come first in it, which forces me to require the overlapped distribution in the construction of the ThyraCrsMatrixFactory. It might take me another day or two.

mperego:
Not sure I understand. Why don't you simply get the local indices from the global ones using the column map?

mperego:
I mean, form the local indices (from `wsElemNodeEqId` ) you get the global indices (from the overlap map), and then you use the `colMap.getLocalElement` to get the local indices wanted by the FE Matrix. It turned out that `colMap.getLocalElement`  is not that expensive.

bartgol:
> I mean, form the local indices (from `wsElemNodeEqId` ) you get the global indices (from the overlap map), and then you use the `colMap.getLocalElement` to get the local indices wanted by the FE Matrix. It turned out that `colMap.getLocalElement` is not that expensive.

Right, it's just a bit cumbersome. To recap, the key points are:

- the col map is not the same as the overlap map. It is in general larger (assuming some of the owned nodes are also shared with other processes, otherwise they do coincide, locally)
- if gid1 and gid2 are 'ghosted' (i.e., they are in the overlap map but not in the owned), they may appear in the col map in a different order in which they appear in the overlap map.

So to summarize, we need to always pass the overlap map to create a matrix, to make sure that when we convert nodeIDs to gids we actually recover the correct gids. As I said, not complicated, but I need to go change the code in a few places.

mperego:
I didn't follow when you said that you have to manually build the col map for the graph. I would not go that way. I discussed something similar with Chris Siefert and Mark Hoemmen at some point but I don't think it is worth. I think it i better to get the "correct" column local indices from the global ones when inserting the values into the matrix. We can chat over the phone if you prefer.

mperego:
@bartgol Is this ready to go once your Tpetra PR gets merged?

bartgol:
Not yet. I'm still finalizing it. But should be ready by the time that PR goes in.

bartgol:
Update: I fixed a few more bugs, and now I'm down to 18/171 tests failing: all ISMIP and FO_TEST, plus the CoupledThickness and  the NSVortexShedding2D ones. For all except the last one I have hope to find the issue. The last one still eludes me, but I hope fixing the other ones might fix that too.

bartgol:
Update: I'm down to a single test failing (NSVortexShedding2D_Steady_Blocked). In serial or with 2 mpi ranks, I get a factorization error (FPE). With 4 ranks, the owned maps end up not being 1-1, which is quite odd.

I suspect there's some place where the interleaved flag is not used correctly.

bartgol:
Boy, the bug causing the NS test to fail was quite nasty, but quite important too, although only for cases where interleaved ordering was FALSE, an option that is overwhelmingly under-tested, I think.

Now *all* tests pass. I haven't turned on all the funky options (like MESH_DEPENDS_ON_SOLUTION), but I'm assuming the nightlies will catch any error.

I'm working on rebasing on master. Quite a bit of stuff happened in the meantime, so it's taking some time with conflicts.

mperego:
@bartgol That's great! Thanks for all that work. I'll review the branch once you rebase to master. Why don't you pick a couple of tests (e.g. one in  FO_GIS and one in FO_Thermo), and set Interleaved Ordering to 0? (after Glen's changes it's not a boolean anymore).

mperego:
Ah, tests using ML/MueLu or FROSch won't work with interleaved ordering = 0, so you might be limited in what tests you can choose.

jewatkins:
> Ah, tests using ML/MueLu or FROSch won't work with interleaved ordering = 0, so you might be limited in what tests you can choose.

Do we throw an error in this case? Just curious.

mperego:
> > Ah, tests using ML/MueLu or FROSch won't work with interleaved ordering = 0, so you might be limited in what tests you can choose.
> 
> Do we throw an error in this case? Just curious.

We do now, in the RBM class. So, we do not directly check for those precs, but since those precs use the RBM class, an error will be thrown.

bartgol:
> @bartgol That's great! Thanks for all that work. I'll review the branch once you rebase to master. Why don't you pick a couple of tests (e.g. one in FO_GIS and one in FO_Thermo), and set Interleaved Ordering to 0? (after Glen's changes it's not a boolean anymore).

Can I leave that for another PR? ;-)

bartgol:
@mperego the branch has been successfully rebased, and all tests pass. I know it is a huge PR, so let me know if you want to go over it together.

As a guideline, main changes:

1) Obviously, we support FECrs matrices now, both for Epetra and Tpetra. However, we only create FECrs graphs (and, therefore, matrices) if the user provides overlapped domain/range maps to the CrsMatrixFactory *AND* if the overlapped maps do not coincide with the non-overlapped ones (which can happen, e.g., in serial mode). Otherwise, we stick with simple Crs. So far there are very few (I think 2) places in Albany where we build graphs, and they both use FECrs.
Note: I had to add a warpper for Epetra_FECrsGraph (and matrix), since Epetra does not expose information on the "overlapped" distribution which we need when it's time to insert values (long ot explain, but I can elaborate if you want).

2) The STK discretization had some interfaces (some unused) removed. E.g., meshToGraph was unused (I think it was used in LCM), so I removed it. Also the `getOwnedDOF` and similar have been removed, in favor of using nodal dof managers (available via `getDOFManager` and similar).

3) The StokesH discretization is gone. The main STK one is able to handle every case now. It creates coupling over the whole column _only if_ the mesh is layered _and_ the ss eqn is defined on either top or bottom of the mesh.

4) The overlapped maps in the stk discretization have changed. In particular, owned dofs come first in the ov map. This means that trying to compare graphs/matrices by inspecting entries, since they will be in different order. Note: the GLOBAL ids are unchanged (meaning that the _physical_ location of gids is unchanged).

5) Due to point 4, layered mesh numbering only makes sense when using the nodes GID. This has been changed all over the code. I also added a method to only get column or layer id (for when you only need one of them), to make the needs of the code at the call site clearer.

6) To avoid bugs, when scattering into a matrix, we do _not_ allow anymore to pass indices that are not in the graph. Before, we used to simply ignore tpetra return code (which notified that some indices were not used).

7) Due to 6, I had to introduce a PHAL::ScatterSideEqnResidual evaluator, to use with the SideSetLaplacian test. I _think_ this can probably be used also in place of LandIce::ScatterResidual2D, but I'm not 100% sure, so I left the LandIce evaluators there.

8) During 7, I realized it was clearer to incorporate the DirichletOffNodeset behavior into the ScatterSideEqnResidual: the idea is to zero out the residual (and set J to a non-zero diag value) on nodes outside the side set where the ss eqn is defined. This makes the bc utils clearer. The DirichletOffNodeSet was not really a bc to start with, it was only something we needed to avoid getting singluar jacobians (this need might go away completely when we handle blocks correctly). 

9) Added some thyra utils and removed some unused ones. E.g., added 'isOnetoOne', and removed `buildRestrictionOperator` and other LinearOp manipulation ones).

10) Bonus: added possibility to not report timers at the end of execution (useful for all non-performance related runs, so you don't have to scroll up 2 pages to go past timings).

There might be more comments in the code (if you care to read them).

PR #596, opened 2020-05-26
Modify order of libraries in export_albany.in

bartgol:
The libraries need to appear in the correct order. This means that, if you have `-lA -lB`, then then libB cannot use anything from libA.

Note: if the downstream project used cmake, the order is irrelevant (cmake can figure that out). But then, if they used cmake, we probably would not need this file...


bartgol:
@mperego this seems to work on my machine. Please, double check it works for you too, then merge away.

mperego:
> 
> 
> @mperego this seems to work on my machine. Please, double check it works for you too, then merge away.

Yes, I tried it out and it works for me as well.

PR #598, opened 2020-06-02
Add forward transient sensitivities to Albany along with corresponding observer

ikalash:
This PR goes hand-in-hand with a Piro PR in Trilinos that just got merged in to master.  The PR invokes the forward transient sensitivities I put into Piro and tests them on several MMS problems.  The PR also has a slight refactor of the observer/discretization classes (observer refactor goes hand-in-hand with Piro observer changes) to enable observation of transient sensitivities dx/dp.  The code automatically detects if transient sensitivities are requested and, if so, write them to the output Exodus file under the name "solution dxdp0", "solution dxdp1", etc.  

Note: transient sensitivities will not work if "Number of Parameter Vectors" > 1, due to limitations in Tempus, which does not handle this case.  If attempting to run with "Number of Parameter Vectors" > 1, an error is thrown.

The attached slides summarize the verification results.  

[fwd_transient_sensitivities_albany.pptx](https://github.com/SNLComputation/Albany/files/4719790/fwd_transient_sensitivities_albany.pptx)


src/Albany_Application.cpp
```diff
@@ -442,7 +413,7 @@ Application::initialSetUp(const RCP<Teuchos::ParameterList>& params)
   // Schwarz problems.
   countScale = 0;
   // Create discretization object
-  discFactory = rcp(new Albany::DiscretizationFactory(params, comm, expl));
+  discFactory = rcp(new Albany::DiscretizationFactory(params, comm, num_params_, expl));
```
mperego, line 416: you could deduce the number of parameters from params.

ikalash: I agree, but there's a routine that does this from the PL in Albany::Application, and I didn't want to duplicate that code in the discretization factory; so I just pass it in.  I don't think it's a big deal since we only construct the discretization factory once.

src/LandIce/problems/LandIce_ParamEnum.hpp
```diff
@@ -13,7 +13,10 @@ enum class ParamEnum
   MuCoulomb    = 2,
   MuPowerLaw   = 3,
   Power        = 4,
-  Homotopy     = 5
+  Homotopy     = 5,
+  Kappa_x      = 6,
+  Kappa_y      = 7,
+  Kappa_z      = 8
 };
```
mperego, line 20: Why are you using this for a non LandIce problem? You could use PHAL_SharedParameter instead of the LanIce_Shared Parameter.

src/problems/Albany_ThermalProblem.hpp
```diff
@@ -115,8 +114,9 @@ namespace Albany {
 #include "Albany_ResponseUtilities.hpp"
 //#include "PHAL_Neumann.hpp"
 #include "PHAL_ThermalResid.hpp"
-#include "PHAL_ThermalSource.hpp"
-
+//IKT FIXME - add dependence on LandIce
+#include "../../LandIce/evaluators/LandIce_SharedParameter.hpp"
+#include "../../LandIce/problems/LandIce_ParamEnum.hpp"
```
mperego, line 120: This could create problems with the libraries that have been just cleaned up by Luca. I would give it a try to the PHAL_ShareParameters, so that you don't have to depend on LandIce, it is much cleaner.

src/disc/stk/Albany_OrdinarySTKFieldContainer_Def.hpp
```diff
@@ -68,12 +67,31 @@ OrdinarySTKFieldContainer<Interleaved>::OrdinarySTKFieldContainer(
...
 #ifdef ALBANY_DTK
-  int output_dtk_field =
+  bool output_dtk_field =
       params_->get<bool>("Output DTK Field to Exodus", false);
 #endif
+  //IKT FIXME? - currently won't write dxdp to output file if problem is steady, 
+  //as this output doesn't work in same way.  May want to change in the future.
+  bool output_dxdp_field = false; 
+  if (num_params_ > 0 && num_time_deriv > 0) output_dxdp_field = true; 
+
```
mperego, line 81: It would be nice to have some way to control the whether to output dxdp from input file.

src/Albany_Application.cpp
```diff
@@ -613,7 +584,8 @@ Application::finalSetUp(
       initial_guess,
       paramLib,
       disc,
-      comm));
+      comm,
+      num_params_));
```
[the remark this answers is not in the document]
ikalash: Wouldn't this deduction duplicate the code in calcTangentDerivDimension in Albany::Application?  I would get the # params from the PL but it seems w/ all the different parameter list options we support (old, new, etc.) it would be on the order of 15 lines of code, that would get duplicated in various places.  You know better than me the different ways to specify the params.  Do you have a suggestion how to succinctly infer them from the PL other than this: https://github.com/SNLComputation/Albany/blob/observer-trans-sens/src/Albany_Application.cpp#L2661-L2680 ?  If so, I'm happy to switch to that.  I thought passing the parameter was the simplest way around repeating all that code to parse the PL but maybe I'm overlooking something.

src/disc/stk/Albany_OrdinarySTKFieldContainer_Def.hpp
ikalash: This would be very easy to do - it would be just like this code: https://github.com/SNLComputation/Albany/blob/observer-trans-sens/src/disc/stk/Albany_OrdinarySTKFieldContainer_Def.hpp#L74-L75 .  I actually thought the opposite of what you said, based on my experience with the DTK output (I put that Exodus output option in for DTK to turn it off from PL) - I thought that it's better not to give an additional PL option.  For the DTK output stuff, it becomes a huge mess b/c people forget that the option is in the input file then wonder with the DTK stuff's not there.  Right now, dxdp only get written for a transient problem with sensitivities enabled, so for the majority of Albany problems, it won't get written.  Anyway, I'm happy to add a PL option similar to DTK if you think it's better.  Would you like me to do that?

src/problems/Albany_ThermalProblem.hpp
ikalash: So, I started with using PHAL_ShareParameters but I couldn't get it to work correctly.  I spent a few hours trying to figure it out.  Then when I switched to the land ice shared parameter, it worked right away.  I figured it wasn't worth my time to figure out what I was doing wrong with the shared parameters since that wasn't really relevant to the sensitivity calculation itself.  The fact that we care about transient sensitivities for LandIce, where we'd be using the LandIce::SharedParameter, solidified this decision.  I did put ifdef guards to prevent configuration errors in the case LandIce is not enabled.  I could move the problem inside the LandIce directory if you think that'll be less confusing.  If you really think it's worth taking the time to figure out why the PHAL_ShareParameters way was not working, I could try to do that.  I thought it wasn't worth the effort.

src/LandIce/problems/LandIce_ParamEnum.hpp
ikalash: I couldn't get it to work for unknown reasons (likely lack of experience on my part) - see my other reply.  What I could do it remove the "Kappa_x", "Kappa_y", "Kappa_z" enums and hijack the others that were already there, as Luca suggested.  Not sure if this would cause more or less confusion.

src/disc/stk/Albany_OrdinarySTKFieldContainer_Def.hpp
mperego: OK, I see your point. It's just that sometime the mesh we get are really big when we are working on large problem. Anyway, since, as you said, it is stored only for transient problems with sensitivities enable, I think it make sense to live it as is. We can always change it in a second time. Thanks.

ikalash:
> One is the fact that you are passing in a lot of places the num_params_, and it seems that in many places you could deduce the number of parameters. Also, it is not clear whether this int refers to the number of scalar parameters or the total number of parameters (including the distributed ones).

I agree with you on this, but I had a hard time coming up with a clean way to deduce the number of params from the PL.  If you have an idea on how to do this, I'm open to suggestions.

> The other issue is that you are using a LandIce class from outside the LandIce folder. We want to avoid this to keep the modules clean. I think you can use PHAL_SharedParameter (in evaluators/utility/) instead.

See my comments in response to your specific comments on the code regarding why I did it this way.

src/disc/stk/Albany_OrdinarySTKFieldContainer_Def.hpp
ikalash: I'll go ahead and resolve this point then.  It's very easy to add the option - just one like, similar to what's done for the DTK output.  We can always do it in the future, like you said.

src/problems/Albany_ThermalProblem.hpp
mperego: I see. We could always create a Thermal::SharedParameter class, basically copying the LandIce one. In this way we would avoid inter-dependencies between albanyLib and landIceLib. Would that make sense to you? Any comment on this @bartgol?

ikalash: I like this idea, but lets see what Luca says.

src/Albany_Application.cpp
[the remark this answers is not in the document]
mperego: No, I agree you would have to use that code (which would be simpler once we get rid of the old syntax). It's really to avoid passing num_params in all those places. Anyway, I'm fine with the current implementation. It is probably not worth the effort to remove that parameter.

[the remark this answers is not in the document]
ikalash: I agree with you 100% that it would be cleaner to pass the num_params around I just couldn't think of a clean way to avoid it.  I think after I discovered that there are so many ways the parameters can be specified, I was even more hesitant to try to infer things from the PL having to do with them...  Anyway, if you're OK with things as it I think I'll keep it, but I'll give it a day or two to see if anyone else, e.g. Glen (another reviewer), has better suggestions.  How about that as a compromise?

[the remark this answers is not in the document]
mperego: Sounds good!

src/problems/Albany_ThermalProblem.hpp
bartgol: We can also copy this version of SharedParameter back into PHAL. Not sure if we should replace the old one or add it along side it. If Irina had some issues with the old one, then it might be worth checking if something is really intrinsically wrong there, and, if so, maybe toss it and keep this.

I'm not a huge fan of the LandIce::SharedParameter class. I wrote it cause I was having some inconsistency of parameters across different field managers, but it might be I was using the PHAL one wrong. The extra template parameters in the class make it a bit stiff, and hard to use, so making the PHAL one work would be much better.

Either way, I don't think including LandIce stuff in the main albany lib is a good pattern, even if ifdeffed out. We spent some effort cleaning up cyclic deps, and this feels like a step back. As I said, migrating LandIce::SharedParameter to PHAL (perhaps as GlobalSharedParameter or something, given its static members) is definitely more appropriate.

bartgol: Notice that porting LandIce::SharedParameter into PHAL would be fine. The class's two last template parameters are an enum type and an enum value. So one can use it with LandIce::ParamEnum, as well as with, say, ThermalParamEnum.

ikalash: Makes sense - thanks Luca.  I'll add the new class and take out the ifdef guards around the Thermal Problem since it will no longer depend on ALI.  I like this solution.

src/Albany_Application.cpp
bartgol: We can take that code and put it in Albany_Utils.hpp. Having the num_params_ extra arg is somewhat suggesting it is an independent argument, while in fact it must agree with what is in the parameter list.

src/problems/Albany_ThermalProblem.hpp
ikalash: I just pushed some changes that rename LandIce::SharedParameter PHAL::SharedParameter.  This allows one to remove the LandIce dependency of ThermalProblem.  Much better!  I will mark this as resolved too.

src/Albany_Application.cpp
ikalash: @bartgol : I like this suggestion.  Let me try it.

src/problems/Albany_ThermalProblem.hpp
mperego: That's great. thanks. What about the old PHAL::SharedParameters in evaluators/utility? Did you remeve it?

bartgol: We should not remove it, imho. That file also contains SharedParameterVec, which allows to use one evaluator for a vector of parameters. The version in LandIce_SharedParameter.hpp does not support that (although, it shouldn't be hard to add it). I'm not sure if SharedParameterVec is used at all (although it seems like ThermalProblem could use it), but I would keep that file. I understand it would generate a name clashing between old and new SharedParameter, which is why I would call the new one something like GlobalSharedParametter or StaticSharedParameter (though static has N+1 meanings and might be confusing).

ikalash: I forgot in my initial commit, but I just did in a second commit.  It was not being used anyway, and when I tried to use it, I had issues, so it's probably best to remove it in favor of the new PHAL::SharedParameter.  Do you think the evaluators/utility dir is a better place for it than evaluators/pde?

ikalash: By "it" I mean the new PHAL_SharedParameter that replaces LandIce_SharedParameter.

ikalash: @bartgol : shoot I pushed my changes before seeing your comment.  The old PHAL::SharedParameter class was not used anywhere.  I'm not sure what to do...  I can bring it back and give the new SharedParameter a different name, maybe, as you suggest?

bartgol: If it wasn't used, then it's fine, I guess. I thought it was, since we do have optimization tests in main albany, but I see now that their parameters are in the DBCs, and Dirichlet evaluators also act as SharedParameter evaluators (not sure if this is the best approach, since it's mixing apples and oranges in the same class, but this is not the time and place to discuss that).

If you can, put a TODO comment in the (new) SharedParameter evaluator, reminding us to implement support for vector parameters.

ikalash: I will do this, Luca.  Do you think it's better for the new SharedParameter file to live in evaluators/pde or evaluators/utility?  I put it in the former, but maybe the latter is better?

bartgol: > Do you think the evaluators/utility dir is a better place for it than evaluators/pde?

Yes, I think it should be located in `evaluators/utility`.

ikalash: I have moved it and added the comment about the TODO.  I will mark this thread re-resolved.

ikalash:
@mperego : I think I addressed all your concerns.  Please let me know if you agree.  If so, I'd like to go ahead and merge this in.

ikalash:
Thanks For approving @mperego and @gahansen .  And thanks for the comments, Mauro!  They helped to make the code more clean so I really appreciate it!

PR #600, opened 2020-06-15
Enthalpy Kokkos Refactor (Volume Evaluators)

mcarlson801:
This PR is a refactor of all of the volume evaluators in the Land Ice Enthalpy problem to enable kokkos parallelization.  Original/legacy implementations can be enabled by setting ALBANY_KOKKOS_UNDER_DEVELOPMENT to off.

### DIFF
--- a/src/LandIce/evaluators/LandIce_EnthalpyBasalResid_Def.hpp
+++ b/src/LandIce/evaluators/LandIce_EnthalpyBasalResid_Def.hpp
@@ -22,8 +22,7 @@ namespace LandIce
 template<typename EvalT, typename Traits, typename Type>
 EnthalpyBasalResid<EvalT,Traits,Type>::
 EnthalpyBasalResid(const Teuchos::ParameterList& p, const Teuchos::RCP<Albany::Layouts>& dl)
- : enthalpyBasalResid(p.get<std::string> ("Enthalpy Basal Residual Variable Name"), dl->node_scalar),
-   sideNodes("sideNodes", 1, 1)
+ : enthalpyBasalResid(p.get<std::string> ("Enthalpy Basal Residual Variable Name"), dl->node_scalar)
 {
   basalSideName = p.get<std::string>("Side Set Name");
 
@@ -52,64 +51,22 @@ EnthalpyBasalResid(const Teuchos::ParameterList& p, const Teuchos::RCP<Albany::L
   vecDimFO     = std::min((int)dims[2],2);
 
   // Index of the nodes on the sides in the numeration of the cell
-  // Teuchos::RCP<shards::CellTopology> cellType;
-  // cellType = p.get<Teuchos::RCP <shards::CellTopology> > ("Cell Type");
-  // sideDim      = cellType->getDimension()-1;
-  // sideNodes.resize(numSides);
-  // for (int side=0; side<numSides; ++side)
-  // {
-  //   // Need to get the subcell exact count, since different sides may have different number of nodes (e.g., Wedge)
-  //   int thisSideNodes = cellType->getNodeCount(sideDim,side);
-  //   sideNodes[side].resize(thisSideNodes);
-  //   for (int node=0; node<thisSideNodes; ++node)
-  //   {
-  //     sideNodes[side][node] = cellType->getNodeMap(sideDim,side,node);
-  //   }
-  // }
-
   Teuchos::RCP<shards::CellTopology> cellType;
   cellType = p.get<Teuchos::RCP <shards::CellTopology> > ("Cell Type");
   sideDim      = cellType->getDimension()-1;
-  int nodeMax = 0;
-  for (int side=0; side<numSides; ++side) {
+  sideNodes.resize(numSides);
+  for (int side=0; side<numSides; ++side)
+  {
+    // Need to get the subcell exact count, since different sides may have different number of nodes (e.g., Wedge)
     int thisSideNodes = cellType->getNodeCount(sideDim,side);
-    nodeMax = std::max(nodeMax, thisSideNodes);
-  }
-  Kokkos::resize(sideNodes, numSides, nodeMax);
-  for (int side=0; side<numSides; ++side) {
-    int thisSideNodes = cellType->getNodeCount(sideDim,side);
-    for (int node=0; node<thisSideNodes; ++node) {
-      sideNodes(side,node) = cellType->getNodeMap(sideDim,side,node);
+    sideNodes[side].resize(thisSideNodes);
+    for (int node=0; node<thisSideNodes; ++node)
+    {
+      sideNodes[side][node] = cellType->getNodeMap(sideDim,side,node);
     }
   }
 
   this->setName("Enthalpy Basal Residual" + PHX::print<EvalT>());
-}
-
-template<typename EvalT, typename Traits, typename Type>
-KOKKOS_INLINE_FUNCTION
-void EnthalpyBasalResid<EvalT,Traits,Type>::
-operator() (const Clear_Residual_Tag& tag, const int& node, const int& cell) const{
-  // enthalpyBasalResid(cell,node) = 0.;
-}
-
-template<typename EvalT, typename Traits, typename Type>
-KOKKOS_INLINE_FUNCTION
-void EnthalpyBasalResid<EvalT,Traits,Type>::
-operator() (const Enthalpy_Basil_Resid_Tag& tag, const int& idx) const{
-  // // Get the local data of side and cell
-  // const int cell = sideSet[idx].elem_LID;
-  // const int side = sideSet[idx].side_local_id;
-
-  // for (int node = 0; node < numSideNodes; ++node)
-  // {
-  //   int cnode = sideNodes[side][node];
-
-  //   for (int qp = 0; qp < numSideQPs; ++qp)
-  //   {
-  //     enthalpyBasalResid(cell,cnode) += basalMeltRateQP(cell,side,qp) *  BF(cell,side,node,qp) * w_measure(cell,side,qp);
-  //   }
-  // }
 }
 
 template<typename EvalT, typename Traits, typename Type>
@@ -131,14 +88,10 @@ evaluateFields(typename Traits::EvalData d)
     for (int node = 0; node < numCellNodes; ++node)
       enthalpyBasalResid(cell,node) = 0.;
 
-  // Kokkos::parallel_for(Clear_Residual_Policy({0,0}, {numCellNodes,d.numCells}), *this);
-
   if (d.sideSets->find(basalSideName)==d.sideSets->end())
     return;
 
   const std::vector<Albany::SideStruct>& sideSet = d.sideSets->at(basalSideName);
-
-  // Kokkos::parallel_for(Enthalpy_Basil_Resid_Policy(0, sideSet.size()), *this);
 
   for (auto const& it_side : sideSet)
   {
@@ -148,8 +101,7 @@ evaluateFields(typename Traits::EvalData d)
 
     for (int node = 0; node < numSideNodes; ++node)
     {
-      // int cnode = sideNodes[side][node];
-      int cnode = sideNodes(side,node);
+      int cnode = sideNodes[side][node];
       enthalpyBasalResid(cell,cnode) = 0.;
 
       for (int qp = 0; qp < numSideQPs; ++qp)
--- a/src/LandIce/evaluators/LandIce_EnthalpyBasalResid_Def.hpp
+++ b/src/LandIce/evaluators/LandIce_EnthalpyBasalResid_Def.hpp
@@ -22,7 +22,8 @@ namespace LandIce
 template<typename EvalT, typename Traits, typename Type>
 EnthalpyBasalResid<EvalT,Traits,Type>::
 EnthalpyBasalResid(const Teuchos::ParameterList& p, const Teuchos::RCP<Albany::Layouts>& dl)
- : enthalpyBasalResid(p.get<std::string> ("Enthalpy Basal Residual Variable Name"), dl->node_scalar)
+ : enthalpyBasalResid(p.get<std::string> ("Enthalpy Basal Residual Variable Name"), dl->node_scalar),
+   sideNodes("sideNodes", 1, 1)
 {
   basalSideName = p.get<std::string>("Side Set Name");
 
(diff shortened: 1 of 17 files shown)


jewatkins:
@mperego @ikalash

src/LandIce/evaluators/LandIce_EnthalpyBasalResid_Def.hpp
```diff
@@ -51,24 +52,66 @@ EnthalpyBasalResid(const Teuchos::ParameterList& p, const Teuchos::RCP<Albany::L
   vecDimFO     = std::min((int)dims[2],2);
 
   // Index of the nodes on the sides in the numeration of the cell
+  // Teuchos::RCP<shards::CellTopology> cellType;
```
mperego, line 55: can this commented out code be deleted?

src/LandIce/evaluators/LandIce_EnthalpyBasalResid_Def.hpp
```diff
@@ -51,24 +52,66 @@ EnthalpyBasalResid(const Teuchos::ParameterList& p, const Teuchos::RCP<Albany::L
...
+void EnthalpyBasalResid<EvalT,Traits,Type>::
+operator() (const Clear_Residual_Tag& tag, const int& node, const int& cell) const{
+  // enthalpyBasalResid(cell,node) = 0.;
+}
+
+template<typename EvalT, typename Traits, typename Type>
+KOKKOS_INLINE_FUNCTION
+void EnthalpyBasalResid<EvalT,Traits,Type>::
+operator() (const Enthalpy_Basil_Resid_Tag& tag, const int& idx) const{
+  // // Get the local data of side and cell
```
mperego, line 100: What's the purpose of this empty operator?

src/LandIce/evaluators/LandIce_EnthalpyBasalResid_Def.hpp
```diff
@@ -51,24 +52,66 @@ EnthalpyBasalResid(const Teuchos::ParameterList& p, const Teuchos::RCP<Albany::L
...
   }
 
   this->setName("Enthalpy Basal Residual" + PHX::print<EvalT>());
 }
 
+template<typename EvalT, typename Traits, typename Type>
+KOKKOS_INLINE_FUNCTION
+void EnthalpyBasalResid<EvalT,Traits,Type>::
+operator() (const Clear_Residual_Tag& tag, const int& node, const int& cell) const{
+  // enthalpyBasalResid(cell,node) = 0.;
```
mperego, line 93: what's the purpose of this empty operator()?

src/LandIce/evaluators/LandIce_EnthalpyBasalResid_Def.hpp
mcarlson801: The changes to EnthalpyBasalResid were not intended to be included in this PR. They have been reverted and the finished version will be a part of the future boundary condition PR.

src/LandIce/evaluators/LandIce_EnthalpyResid.hpp
```diff
@@ -91,7 +91,38 @@ class EnthalpyResid : public PHX::EvaluatorWithBaseImpl<Traits>,
   double flux_reg_alpha;
   double flux_reg_beta;
 
+  const double powm3 = 1e-3;  //[k], k=1000
+  const double powm6 = 1e-6;  //[k^2], k=1000
+  const double pow3 = 1e3;  //[k^{-1}], k=1000
+
```
jewatkins, line 97: I'd go ahead and make these `static constexpr` instead of `const` and hopefully the compiler will optimize these out. Replace `const` with `constexpr` in the other kernels too whenever the value is known at compile time.

src/LandIce/evaluators/LandIce_EnthalpyResid_Def.hpp
```diff
@@ -26,6 +27,13 @@ Type distance (const Type& x0, const Type& x1, const Type& x2,
     return 0;
 }
 
+template<typename Type>
+KOKKOS_INLINE_FUNCTION
+Type deviceMax(Type a, Type b)
+{
+  return (a > b) ? a : b;
+}
+
```
jewatkins, line 36: We'll have to make a utility namespace for this sort of thing. We should probably be using the Sacado max too. I'll look into it, maybe we already have something.

src/LandIce/evaluators/LandIce_EnthalpyResid.hpp
bartgol: If you make them constexp, you might just keep them in the `_Def.hpp` file, without storing them in the class. If you make them static, the compiler cannot optimize them away (a static var must be available the whole execution of the program). Besides, constexpr can be "recreated" many times, since they won't even be there anyways at run time (most likely, unless odrused).

src/LandIce/evaluators/LandIce_EnthalpyResid_Def.hpp
bartgol: I think `std::max` should work. I'm pretty sure Sacado overloads it.

mcarlson801: They were originally std::max and std::min respectively but that resulted in compiler warnings about host code being inaccessible from device and then segmentation faults at runtime.

src/LandIce/evaluators/LandIce_EnthalpyResid.hpp
jewatkins: hmm then it's probably better just to move them to the various operators and just use constexpr, even if it has to be defined multiple times.

src/LandIce/evaluators/LandIce_EnthalpyResid_Def.hpp
jewatkins: I don't think std::max will work on device which I think means we have to create our own overload.

src/LandIce/evaluators/LandIce_Temperature_Def.hpp
```diff
@@ -79,6 +79,23 @@ Temperature(const Teuchos::ParameterList& p, const Teuchos::RCP<Albany::Layouts>
   Tm    = physics.get<double>("Atmospheric Pressure Melting Temperature"); //273.15
 }
 
+template<typename EvalT, typename Traits, typename TemperatureST>
+KOKKOS_INLINE_FUNCTION
+void Temperature<EvalT,Traits,TemperatureST>::
+operator() (const int &node, const int &cell) const{
+
+  if ( enthalpy(cell,node) < enthalpyHs(cell,node) )
+    temperature(cell,node) = pow6 * enthalpy(cell,node)/(rho_i * c_i) + T0;
```
bartgol, line 88: Again, precompute scalar factor.

src/LandIce/evaluators/LandIce_EnthalpyResid_Def.hpp
jewatkins: So I think max() is available but it's not in std namespace.

src/LandIce/evaluators/LandIce_EnthalpyResid_Def.hpp
```diff
@@ -127,6 +135,128 @@ EnthalpyResid(const Teuchos::ParameterList& p, const Teuchos::RCP<Albany::Layout
   flux_reg_beta = flux_reg_list.get<double>("beta");
 }
 
+template<typename EvalT, typename Traits, typename VelocityST, typename MeltTempST>
+KOKKOS_INLINE_FUNCTION
+void EnthalpyResid<EvalT,Traits,VelocityST,MeltTempST>::
+stabilizationInitialization(int cell, VelocityST& vmax_xy, ScalarT& vmax, ScalarT& vmax_z, 
+  MeshScalarT& diam, MeshScalarT& diam_xy, MeshScalarT& diam_z, ScalarT& wSU) const {
+  for (std::size_t qp = 0; qp < numQPs; ++qp) {
+      ScalarT w = verticalVel(cell,qp);
```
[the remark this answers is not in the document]
jewatkins: So I think we've had issues in the past with using auto for fad types in kernels. I don't remember the exact issue unfortunately. It might worth exploring again.

src/LandIce/evaluators/LandIce_EnthalpyResid_Def.hpp
```diff
@@ -127,6 +135,128 @@ EnthalpyResid(const Teuchos::ParameterList& p, const Teuchos::RCP<Albany::Layout
...
+  MeshScalarT diam = 0.0, diam_xy = 0.0, diam_z = 0.0;
+  ScalarT wSU = 0.0;
+
+  stabilizationInitialization(cell, vmax_xy, vmax, vmax_z, diam, diam_xy, diam_z, wSU);
+
+  for (std::size_t node = 0; node < numNodes; ++node) {
+    evaluateResidNode(cell, node);
+
+    for (std::size_t qp = 0; qp < numQPs; ++qp) {
+      Residual(cell,node) += pow3*(EnthalpyGrad(cell,qp,0)*wGradBF(cell,node,qp,0) +
```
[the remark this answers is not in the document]
jewatkins: I think it's better to make some temporary storage in register/cache memory but I could be wrong.

src/LandIce/evaluators/LandIce_EnthalpyResid_Def.hpp
```diff
@@ -127,6 +135,128 @@ EnthalpyResid(const Teuchos::ParameterList& p, const Teuchos::RCP<Albany::Layout
...
+  }
+
+}
+
+template<typename EvalT, typename Traits, typename VelocityST, typename MeltTempST>
+KOKKOS_INLINE_FUNCTION
+void EnthalpyResid<EvalT,Traits,VelocityST,MeltTempST>::
+operator() (const Other_Stabilization_Tag& tag, const int& cell) const{
+
+  for (std::size_t node = 0; node < numNodes; ++node) {
```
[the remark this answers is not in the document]
jewatkins: Yeah we're planning to expose more parallelism and test this in a second iteration. I'm a bit worried about data access patterns when parallelizing over nodes. We just wanted to push this in for now.

src/LandIce/evaluators/LandIce_EnthalpyResid_Def.hpp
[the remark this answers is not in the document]
bartgol: Interesting. I also remember some issue with auto. But if I remember correctly, we were using auto temporaries _as if_ they were some sort of scalar type, while in Sacado they may be expressions.

I would explore it, but maybe we should check with Eric Phipps if storing expressions for temporaries is indeed beneficial or not. I _think_ it is.

src/LandIce/evaluators/LandIce_HydrostaticPressure.hpp
```diff
@@ -59,7 +59,17 @@ class HydrostaticPressure : public PHX::EvaluatorWithBaseImpl<Traits>,
   double rho_i;  // [kg m^{-3}]
   double g;      // [m s^{-2}]
 
+  const double pow3 = 1.0e3;
+
   PHAL::MDFieldMemoizer<Traits> memoizer;
+
+  typedef Kokkos::View<int***, PHX::Device>::execution_space ExecutionSpace;
+
+  typedef Kokkos::MDRangePolicy< ExecutionSpace, Kokkos::Rank<2> > Hydrostatic_Pressure_Policy;
```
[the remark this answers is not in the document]
jewatkins: So for this kernel the latter you mentioned is exactly what we want to do. I was hoping the MDRange policy provides this without having to worry about retrieving indices (i.e. adding code bloat). My understanding was that the MDRange policy was supposed to only be used for tightly nested parallelism like this one. We've been running some tests and it doesn't look like it's working as I hoped so we may have to just use a regular RangePolicy.

We'll have to edit the view layout for other, more complex kernels to improve data access patterns. The layout should be fine for this one.

src/LandIce/evaluators/LandIce_EnthalpyResid_Def.hpp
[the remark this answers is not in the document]
jewatkins: It probably depends on whether you'll be reusing the memory or not. In this case, auto might actually make sense.

src/LandIce/evaluators/LandIce_EnthalpyResid_Def.hpp
mcarlson801: It looks like max() and min() are available, the issue was trying to use std::max and std::min. The deviceMax and deviceMin functions are now removed.

jewatkins: That's great and good to know. I wonder who's dropping the sacado device max() into the global namespace.

src/LandIce/evaluators/LandIce_HydrostaticPressure.hpp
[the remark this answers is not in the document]
mcarlson801: I did a timing comparison for MDRange and Range and got an interesting result. The reason I have MDRange policies in place with the node/qp index first is that for tetrahedral elements, there is a pretty significant speedup for some evaluators. 

Here is one example from the comparison on tets:

MDRange (cells, nodes/qp) Phalanx: Evaluator 62: [\<Residual\>] Phi: 0.13328 - 6.92016% [101]
MDRange (nodes/qp, cells) Phalanx: Evaluator 62: [\<Residual\>] Phi: 0.00465843 - 0.246764% [101]
Range   (cells)                      Phalanx: Evaluator 62: [\<Residual\>] Phi: 0.0891155 - 4.89937% [101]

However, after running the comparison on wedge elements, as expected, the simple range policy over cells is the clear better performer.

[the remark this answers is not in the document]
bartgol: What about Range(cells*nodes) or Range(cells*qp)?

Also, the way you order in MDRange clearly impacts the performance, since it affects how reads will be (coalesced or not) on GPU. I _thought_ Phalanx uses layout left on Cuda, but your timings make me wonder whether it was always Right instead. I have to check.

Also, if you try Range(cells*nodes), you should try to retrieve icell and inode both ways:

```
icell = iter / numNodes;
inode = iter % numNodes;
```
and
```
inode = iter / numCells;
icell = iter % numCells;
```
One of these will be significantly faster than the other. What I would like to know, however, is how these compare to the corresponding MDRange. Since our tiles are 1x1, if I recall correctly a conversation with a kokkos developer, then MDRange might not be beneficial (and even hurt a bit). But I might be wrong. Worth checking, if you have time.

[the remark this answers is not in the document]
jewatkins: Max, are you saying Range(cell) is performing better than MDRange(cell,node) and MDRange(node, cell) on wedge elements?

The fact that the times for the MDRange(node, cell) for both residual and jacobian on tets are the same seems extremely suspicious to me. Just want to confirm, you have `export CUDA_LAUNCH_BLOCKING=1` in your script, correct?

Luca, FYI, these are timings for a different evaluator, `LandIce_LiquidWaterFraction_Def.hpp`. There's some possible thread divergence there so memory accesses may not be coalesced. Phalanx does use left layout (as far as I know unless things have changed recently). These MDRange policies are giving us different results depending on the element type (for wedges, it seems to be working correctly on a simple kernel (i.e. (cell, node) is faster) but not for tets). What you suggest is worth trying to see how it compares with the MDRange policy but maybe on the simple kernel (i.e. HydrostaticPressure). If `icell` as the faster index doesn't perform better in that experiment, I would not be happy. If we go that route though, we'll have to create a layer of abstraction to deal with different data layouts which I was hoping to avoid by using MDRange policies.

If what you're saying is correct about MDRange policy, I'm not sure when they would ever be useful. My understanding was that it was for tightly nested loops and team policies were for more complex nested loops. I could be wrong too though and these things have been changing so it's probably better to ask a kokkos dev and test it ourselves.

My suggestion would be to just go with the Range(cell) policy for now and we can come back to this later. Max, I have some code/scripts we could use to test all these cases out in an isolated environment, that might be more efficient.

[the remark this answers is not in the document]
bartgol: Wait, why would you prefer Range(numCells) over Range(numCells*numNodes) or Range(numCells*numQPs)? Assuming fully independent nodes/QPs (which is usually the case, if not always), you would be leaving, for nodes, a factor 4 (or more for hexa/wedges) on the table (qps it depends, for stokesFo nothing, since there's 1 qp only).

Obv, one needs to carefully do the math with indices, since you want the fast striding thread idx to match the fast striding idx in the view layout. But if you do that, then you gain quite a bit.

As for MDRanges, we should indeed talk to kokkos devs. My impression is based on a memory of a conversation I had in late 2016, so I may have completely made up stuff. One thing that could explain the MDRange behavior is if internally kokkos maps, say, 1st index to blocks (or warp) and 2nd index to warp (or warp lane). Different elements yield different blocking structures (or, morally, different internal "team" policies), which may utilize the gpu in very different ways. So going from 4 nodes/elem to 6 nodes/elem might screw up gpu occupancy, leaving a lot of stuff on the table. Beware, this is a very wild guess.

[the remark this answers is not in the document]
jewatkins: I prefer Range(numCells) **for now** because it's often enough for full occupancy and these kernels are a very small portion of the overall runtime so it won't make a different in the short term. Range(numCells*numNodes) requires an abstraction for different layouts which can be done but it would take more time and would have to be maintained. I'd like to avoid that with the MDRange policies. If that turns out to be less performant than Range(numCells*numNodes) than we can create the abstraction but I really think that should be something kokkos should be able to handle.

You only gain quite a bit when there's not enough cells to fill the gpu (i.e. strong scaling). In some cases, you actually lose performance, for example, when threading over nodes can affect coalesced memory access.

There's a kokkos training next week, maybe we can get some of our questions answered. Even so, I'd still like to see some isolated evidence of a Range(numCells*numNodes) policy performing better than an MDRange policy across different architectures to send over to the kokkos team. I think I did this before with Aeras but that was 3 or 4 years ago. We need some updated data because I think there's been some changes to the MDRange policy.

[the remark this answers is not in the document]
bartgol: Thanks, now I see your point. Then it's ok, we can keep Range(numCells) if it works better than the MDRange.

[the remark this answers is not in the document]
jewatkins: We'll keep you posted. Hopefully we can transition to something that works better all around soon.

src/LandIce/evaluators/LandIce_EnthalpyResid_Def.hpp
mcarlson801: I built this on Blake to make sure everything works for serial and the compiler couldn't find max() and min(). We'll need to find a portable solution for this after all.

jewatkins: Okay I think the best option will be to create some sort of utility namespace for this. We can discuss later today.

bartgol: This seems very odd. Have you tried to include cmath? Perhaps intel's header are different from gcc (or other intel's other than blake, for some reason); cmath might have been included indirectly in other builds, but not here.

Another solution could be this:

```
#ifndef KOKKOS_ENABLE_CUDA
template <typename T>
KOKKOS_INLINE_FUNCTION
const T& max (const T& a, const T& b) { return a > b ? a : b; }
#else
using std::max;
#endif
```

jewatkins: Your latter suggestion is what I was thinking but I think we should put it somewhere for general use (along with min).

bartgol: Oh, yeah, it shouldn't be buried deep inside some evaluator class Def file. Maybe `Albany_Utils.hpp`, but that file may become a panacea. Perhaps `Albany_KokkosUtils.hpp` or `Albany_MathUtils.hpp`.

jewatkins: `Albany_KokkosUtils.hpp` sounds good.

mperego:
@mcarlson801 do you think this PR can go in today or tomorrow? We are almost ready with the Hessian work, so it'd be nice to have this merged in before we start merging master into the Hessian branch. If it's not ready it's fine, of course.

mcarlson801:
@mperego Yeah, this PR is ready to go. It just needs to be looked over one more time by @jwatkins before it gets merged.

PR #601, opened 2020-06-16
Rough out some block discretization infrastructure.

gahansen:
This commit is a start at the block discretization class suggested in the wiki post

https://github.com/SNLComputation/Albany/wiki/Blocked-linear-system

(Update 11/14/18).

There isn't a lot in the class yet, but a fair amount of cleanup of some Albany structure that is unneeded as we move forward and would complicate the block work.


bartgol:
@gahansen I have comments that are too long for github reviews. Let's chat at the telecon today.

gahansen:
I will break this commit up into multiple commits as suggested in the meeting. I'll convert the BlockDiscretization class from a template to a container using a typedef. Thanks all for your help.

PR #607, opened 2020-07-04
LandIce: refactoring of NullSpaceUtility

mperego:
This is a refactoring of the nullSpaceUtility class. Becuase of different implementations of elasticity/nonelesticity problems it became hard to look at the code and understand how to compute the null space dimension. Hope you find this refactor  useful.  Before merging the branch I'll fix the input files of the performance repo, so that they comply with these changes.

For scalar equations it often makes sense to have constants in the near null space
For elesticity type problems, the null space is constituted of rigid body modes,
 consisting of constants/transalations and rotations associated to a physical vector
Here we assume that the we have a set of equations and possibly the the first 2 or 3  equations
 are for a 2d or 3d physical vector (e.g. velocity).
We can compute constant modes for all the equations and then add rotations for the vector.
It is possible to choose whether to compute constant modes and/or rotation modes.
The dimension of the null space is determined by the class and does not need to be computed by the user.
Different checks have beed added to make sure that the methods of the RigidBodyModes class
 are called in the right sequence.
Also we check whether the interleaved ordering is compatible with what implemented.

### DIFF
--- a/src/LandIce/problems/LandIce_StokesFOBase.cpp
+++ b/src/LandIce/problems/LandIce_StokesFOBase.cpp
@@ -105,6 +105,17 @@ StokesFOBase (const Teuchos::RCP<Teuchos::ParameterList>& params_,
   // By default, we are not coupled to any other physics
   temperature_coupled = false;
   hydrology_coupled = false;
+
+  // Determine what Rigid Body Modes to compute and pass to the preconditioner
+  std::string rmb_list_name = "LandIce Rigid Body Modes For Preconditioner";
+  if(params->isSublist(rmb_list_name)) {
+    auto rbm_list = params_->sublist("LandIce Rigid Body Modes For Preconditioner");
+    computeConstantModes = params->sublist(rmb_list_name).get<bool>("Compute Constant Modes");
+    computeRotationModes = params->sublist(rmb_list_name).get<bool>("Compute Rotation Modes");
+  } else {
+    computeConstantModes = true;
+    computeRotationModes = false;
+  }
 }
 
 void StokesFOBase::buildProblem (Teuchos::ArrayRCP<Teuchos::RCP<Albany::MeshSpecsStruct> >  meshSpecs,
@@ -285,7 +296,7 @@ StokesFOBase::getStokesFOBaseProblemParameters () const
   validPL->sublist("LandIce Noise", false, "");
   validPL->set<bool>("Use Time Parameter", false, "Solely to use Solver Method = Continuation");
   validPL->set<bool>("Print Stress Tensor", false, "Whether to save stress tensor in the mesh");
-
+  validPL->sublist("LandIce Rigid Body Modes For Preconditioner", false, "");
   return validPL;
 }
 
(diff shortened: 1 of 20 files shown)


bartgol:
@mperego Could you please rewrite the commit message, so that there is an empty line between the 1 line header and the full description? Not having it is messing up the output of `git log --oneline` (which is my (and probably many other) alias for git log).

jewatkins:
This looks a lot cleaner now thanks

src/LandIce/problems/LandIce_StokesFOBase.cpp
```diff
@@ -105,6 +105,17 @@ StokesFOBase (const Teuchos::RCP<Teuchos::ParameterList>& params_,
   // By default, we are not coupled to any other physics
   temperature_coupled = false;
   hydrology_coupled = false;
+
+  // Determine what Rigid Body Modes to compute and pass to the preconditioner
+  std::string rmb_list_name = "LandIce Rigid Body Modes For Preconditioner";
+  if(params->isSublist(rmb_list_name)) {
+    auto rbm_list = params_->sublist("LandIce Rigid Body Modes For Preconditioner");
+    computeConstantModes = params->sublist(rmb_list_name).get<bool>("Compute Constant Modes");
+    computeRotationModes = params->sublist(rmb_list_name).get<bool>("Compute Rotation Modes");
```
jewatkins, line 114: Maybe set these to default to true and false, respectively just in case one is set in the sublist but not the other.

mperego:
> @mperego Could you please rewrite the commit message, so that there is an empty line between the 1 line header and the full description? Not having it is messing up the output of `git log --oneline` (which is my (and probably many other) alias for git log).

will do.

mperego:
> This looks a lot cleaner now thanks

thanks!

src/LandIce/problems/LandIce_StokesFOBase.cpp
mperego: I'd prefer this way. I think it is safer not to provide defaults  in the sublist, so that if one mispells the parameters, an error will be thrown. This could also be achieved by checking if the  parameters in the sublist are valid, but it would take a little more coding.

mperego:
I'm going to fix the commit message and push this in a couple of hours if I don't get any objections.

src/LandIce/problems/LandIce_StokesFOBase.cpp
jewatkins: Okay great, if the default is throw a readable error then it should be fine.

jewatkins:
Looks good to me.

ikalash:
When I tested this PR, I got that 3 tests failed: FO_Test_ConfinedShelf_Epetra, FO_Test_ConfinedShelf_DepthIntBC_Epetra, FO_GIS_AdjointSensitivity_StiffeningBasalFriction_Epetra .  Perhaps the input files need to be updated?  Anyway, we can look at these in the dashboard tomorrow morning to see the verbose error.  My dirs where I was testing earlier have already been blown away.

jewatkins:
Hmm odd it looks like it only happened on your workstation, serial/openmp but the debug build passed. Looks like a segfault, no info. Maybe it will be okay tomorrow?

ikalash:
I just pulled the code again and the issue is still there.  I will open a separate issue on the failling tests.

PR #610, opened 2020-07-10
LandIce: Cleaning a bit computation of flow factor

mperego:
1. Remove the flag ENABLE_USE_CISM_FLOW_PARAMETERS
2.  compute the flow factor used in basal sliding laws in the same way as that used for the ice viscosity


ikalash:
Are the CISM flow parameters default then, with the ENABLE_USE_CISM_FLOW_PARAMETERS flag removed?

mperego:
> Are the CISM flow parameters default then, with the ENABLE_USE_CISM_FLOW_PARAMETERS flag removed?

yes. All the tests were already using the default  (CISM) option.

PR #613, opened 2020-07-20
Nodes to cell interpolation

mperego:
Added alternative method to compute nodes to cell interpolation
 
Node to cell interpolation can now be computed either by
1. evaluating the field at the cell barycenter
2. averaging the field over the cell (previous behavior)
    
Note, the two options are equivalent for affine elements, triangles and tets, but not for quads, hexas and wedges
   
This work  is motivated by making standalone Albany and MALI consistent. Because of the temperature DOFs location, the only way to compute the same temperature over a cell is for Albany to evaluate it on prisms' barycenters.  LandIce problems has beed modified accordingly.

In addtion the Humboldt mesh was updated and a check has been added to STK AsciiMesh2D to make sure that the global ids are sorted (in fact  the temperature field was not be correctly imported from ascii meshes written by MPAS).


src/LandIce/problems/LandIce_StokesFOHydrology.cpp
```diff
@@ -126,6 +126,7 @@ void LandIce::StokesFOHydrology::buildProblem (Teuchos::ArrayRCP<Teuchos::RCP<Al
 
   Intrepid2::DefaultCubatureFactory   cubFactory;
   cellCubature = cubFactory.create<PHX::Device, RealType, RealType>(*cellType, meshSpecs[0]->cubatureDegree);
+  cellCubature1pt = cubFactory.create<PHX::Device, RealType, RealType>(*cellType, 0);
```
bartgol, line 129: This is minor, but there really is no need to create a cubature in the problem, to then pass it to the evaluator. The evaluator can get the cell topo from the intrepid basis, and from that create the 1pt cubature internally. This might help keep interfaces and classes a tag smaller (avoiding confusion).

src/evaluators/interpolation/PHAL_NodesToCellInterpolation_Def.hpp
```diff
@@ -13,10 +13,25 @@ namespace PHAL {
 template<typename EvalT, typename Traits, typename ScalarT>
 NodesToCellInterpolationBase<EvalT, Traits, ScalarT>::
 NodesToCellInterpolationBase (const Teuchos::ParameterList& p,
-                              const Teuchos::RCP<Albany::Layouts>& dl) :
-  BF        (p.get<std::string>("BF Variable Name"), dl->node_qp_scalar),
-  w_measure (p.get<std::string>("Weighted Measure Name"), dl->qp_scalar)
+                              const Teuchos::RCP<Albany::Layouts>& dl)
 {
+  std::string interpType = p.get<std::string>("Interpolation Type");
+  TEUCHOS_TEST_FOR_EXCEPTION ((interpType != "Value At Cell Barycenter") && (interpType != "Cell Average"),
```
bartgol, line 19: We could place the enum in `problems/Albany_ProblemUtils.hpp`, and force users to pass the enum itself rather than a string (less error prone). Catch typos at compile time rather than run time.

src/evaluators/interpolation/PHAL_NodesToCellInterpolation_Def.hpp
```diff
@@ -13,10 +13,25 @@ namespace PHAL {
...
+      std::runtime_error, "Interpolation Type can either be \"Value At Cell Barycenter\" or \"Cell Average\"");
+
+  interpolationType = (interpType == "Value At Cell Barycenter") ?
+      ValueAtCellBarycenter : CellAverage;
+
+  if(interpolationType == ValueAtCellBarycenter) {
+    onePointCubature = p.get<Teuchos::RCP <Intrepid2::Cubature<PHX::Device> > >("One Point Cubature");
```
bartgol, line 26: See comment above. The user should not have to pass the cubature. If, during merges or other refactors, or for whatever reason, someone passes a cubature that is not a 1pt cubature, we would not catch the error. Sure, we can put a Teuchos test to check it, but why bother if we can build the cubature internally?

src/evaluators/interpolation/PHAL_NodesToCellInterpolation_Def.hpp
```diff
@@ -70,38 +100,57 @@ void NodesToCellInterpolationBase<EvalT, Traits, ScalarT>::evaluateFields (typen
...
   for (int cell=0; cell<workset.numCells; ++cell)
   {
-    meas = 0.0;
-    for (int qp(0); qp<numQPs; ++qp)
-    {
-      meas += w_measure(cell,qp);
-    }
+    if(interpolationType == CellAverage) {
+      meas = 0.0;
+      for (int qp(0); qp<numQPs; ++qp)
```
bartgol, line 105: Not for this PR, but in the future we should avoid looping over qps for affine geometries (since the jacobian is constant). I think you mentioned that Intrepid2 is implementing (or already has) some logic to handle this case. Perhaps we should ask Intrepid2 to do this kind of geometric calculations...

src/LandIce/problems/LandIce_StokesFOHydrology.cpp
mperego: Good point. I'll do that.

src/evaluators/interpolation/PHAL_NodesToCellInterpolation_Def.hpp
mperego: I can do that. I just thought it would be better not to add an enum to `problems/Albany_ProblemUtils.hpp` that is only used by one evaluator.

src/evaluators/interpolation/PHAL_NodesToCellInterpolation_Def.hpp
mperego: agreed

src/evaluators/interpolation/PHAL_NodesToCellInterpolation_Def.hpp
mperego: I agree, at some point we should use Intrepid2 in a more systematic way. That Intrepid2 optimization should be ready by the end of this FY

src/evaluators/interpolation/PHAL_NodesToCellInterpolation_Def.hpp
bartgol: Yeah, that's a good point. I guess we could do either way. We might have it as a public member type in the class, then customers would have to include that file to use it. It might become a bit cumbersome, I guess.

However, since it is a binary choice, you could also think about making it a bool argument.

Anyhow, I guess there's not a ton to gain by turning the string into an enum/bool. So feel free to leave as is.

mperego: Sorry but I'll opt for the option that requires the least effort.. i.e. leave it as is.
I don't like bools where it is not clear what the other option is...   if you have    `bool cellAverage`, it is not clear what you are doing when it is false.

bartgol: Agreed.

mperego:
@bartgol I moved the creation of the cubature in the problem. I rebased to keep the commits clean.

bartgol:
Looks good, thanks!

jewatkins:
Looks good to me. This will most likely conflict with one of @mcarlson801 branches since he added a kernel for NodesToCell. Max, let us know if you need help resolving conflicts once this goes in.

mperego:
> Looks good to me. This will most likely conflict with one of @mcarlson801 branches since he added a kernel for NodesToCell. Max, let us know if you need help resolving conflicts once this goes in.

yeah, sorry about that. Should I hod this until max's branch goes in?

jewatkins:
I think it's better if this goes in first, then Max can rewrite the kernels based on changes.

mcarlson801:
@mperego I took a look at the new implementation of NodeToCellInterpolation and it should be straightforward for me to merge it with my PR. It will probably be easier if your changes go in first, then I can update my PR and merge with Luca's new changes all at once.

mperego:
@jewatkins @ikalash  I expect that some of the ali performance tests will fail tomorrow. Becasue of the change in the interpolation algorithm some the response will be slightly off. I guess it's easier to let them run tonight and that fix the regression results tomorrow.

bartgol:
@mperego didn't you update the regression values too? I think I saw some mods for those.

jewatkins:
I don't think we actually check the response in the performance tests. Do we want to start checking surface velocity mismatch like the regression tests?

bartgol:
I don't think so. But we should still evaluate responses, just in case something is off in some response evaluator.

jewatkins:
So turn on the surface velocity mismatch response but don't run the regression?

mperego:
> @mperego didn't you update the regression values too? I think I saw some mods for those.

I'm referring to this repo https://github.com/ikalash/ali-perf-tests

mperego:
> I don't think we actually check the response in the performance tests. Do we want to start checking surface velocity mismatch like the regression tests?

Ah ok, then we should be good.

mperego:
Not sure if it is worth adding regression tests to the performance runs as well.

bartgol:
> So turn on the surface velocity mismatch response but don't run the regression?

Yeah, that's what I would do. Especially with all the wrap/unwrap stuff of Thyra, it is possible that some performance loss slips in if we don't monitor responses as well. Sure, it is not a very heavy part of the computation, but there's no reason not to monitor those too, imho.

ikalash:
My vote is to turn on regression testing in the performance tests, yes.  This will allow us to monitor performance of the response calculation, as well as verify that things don't get broken. I have seen it where issues creep into the code that are not caught by the smaller / unit-like tests and only show up in bigger problems - another reason to turn on the comparison checking in the performance nightlies.  We can discuss this at the Albany meeting this afternoon, if folks would like.

PR #612, opened 2020-07-21
GPU build failures on ride

bartgol:
There are some GPU build failures on ride, which are due to type mismatch for two size_type template arguments (size_t on one side, and unsigned int (the size type for cuda) on the other), during a view copy constructor.

I am not sure why these are showing up now. I don't recall any code in the FECrs PR that might have changed our types. Additionally, I _think_ our `DeviceLocalMatrix` type should be compatible with the `local_matrix_type` inside Tpetra::CrsMatrix (they might have different local ordinal types, since Albany usee `int32_t`, while Trilinos uses plain `int`, but I don't think that is related to the error). It is almost as if one of the two views had the wrong execution space, which is hard to believe.

One solution might be to use Tpetra::CrsMatrix to get the type of a device local matrix. However, I'd like to understand why we get this error in the first place.


bartgol:
Wow, I didn't know that [hub](https://hub.github.com/) (and I'm guessing also its successor, [gh](https://github.com/cli/cli)) was able to turn an issue into a PR. Neat.

PR #615, opened 2020-07-24
Add a new EvalT called HessianVec and add the specializations of the evaluators related to HessianVec

kliegeois:
This PR adds the capability to use Automatic Differentiation (AD) to compute the Hessian-vector products of the response functions by nesting AD types.

Currently, four contributions of the Hessian-vector products are computed:
- ,
- ,
- ,
- , 

where  is the solution,  is a first distributed parameter,  is a potentially different second distributed parameter,  is the response function,  is a direction vector with the same dimension as the vector ,  is a direction vector with the same dimension as the vector , and  is a direction vector with the same dimension as the vector .

The implementation has been tested comparing results computed with AD and with Finite Differences (FD) of the gradients on a heat problem and FO_GIS examples.
However, this PR does not include tests yet. They will be included in a second PR.

The ctest tests are passing with a release build but not with a debug build (known issue of lclSuccess ==false in FECrsGraph).

The strategy used to compute these Hessian-vector product contributions (either AD or FD) is controlled in the .yaml with the parameter 
```
Problem:
  Distributed Parameters:
    Hessian-vector products use AD:
```
which is `true` by default. If wanted, I can modify the parameter lists and set this option inside a different sub-list.

Future work should address:
- derivatives with respect to non-distributed parameters,
- derivatives with respect to values of boundary conditions,
- derivatives with respect to time derivative values,
- Hessian-vector products related to the  function.


mperego:
@kliegeois That's great. thanks!

@bartgol @ikalash @jewatkins , I've suggested @kliegeois to create this PR even if it has no tests for Hessians yet, because I prefer this to be merged in sooner rather than later. Of course @kliegeois tested already the correctness of the second derivatives, what's missing are proper regression tests and this will be addressed soon. Also in order for second derivatives to be computed correclty we need a modificantion in Sacado, and we don't want to hold this PR because of that.

src/evaluators/gather/PHAL_GatherScalarNodalParameter_Def.hpp
```diff
@@ -320,4 +320,156 @@ evaluateFields(typename Traits::EvalData workset)
...
+    for (std::size_t node = 0; node < num_nodes; ++node) {
+
+      // Initialize Fad type for parameter value
+      const LO id = wsElDofs((int)cell,(int)node,0);
+      RealType pvec_val = (id >= 0) ? pvec_constView[id] : 0;
+
+      auto val = (this->val)(cell,node);
+      val = FadType(num_deriv, pvec_val);
+      if (is_direction_active && (g_xp_is_active||g_pp_is_active))
+        val.val().fastAccessDx(0) = (id >= 0) ? vvec_constView[id] : 0;
```
bartgol, line 393: Why do we modify only Dx(0)? Could you add a quick comment on what's happening here? I would have expected something like `Dx(param_index)`

src/evaluators/gather/PHAL_GatherScalarNodalParameter_Def.hpp
```diff
@@ -320,4 +320,156 @@ evaluateFields(typename Traits::EvalData workset)
...
+    for (std::size_t node = 0; node < num_deriv; ++node) {
+      const GO base_id = layeredMeshNumbering.getColumnId(elNodeID[node]);
+      const GO ginode = layeredMeshNumbering.getId(base_id, fieldLevel);
+      const LO p_lid= p_indexer->getLocalElement(ginode);
+      RealType pvec_val = (p_lid >= 0) ? pvec_constView[p_lid] : 0;
+
+      auto val = (this->val)(cell,node);
+      val = FadType(val.size(), pvec_val);
+      if (is_direction_active && (g_xp_is_active||g_pp_is_active))
+        val.val().fastAccessDx(0) = (p_lid >= 0) ? vvec_constView[p_lid] : 0;
```
bartgol, line 468: Same here, why Dx(0)?

src/evaluators/response/QCAD_ResponseFieldValue.hpp
```diff
@@ -126,6 +126,47 @@ namespace QCAD {
     int numNodes;
   };
 
+  template<typename Traits>
+  class FieldValueScatterScalarResponse<PHAL::AlbanyTraits::HessianVec,Traits> :
```
bartgol, line 130: Are we even using this evaluator anymore? I know all the work has been done already, so we might as well, but we might as well trim unused (and unlikely to be used) stuff...

src/evaluators/utility/QCAD_EvaluatorTools.hpp
```diff
@@ -75,6 +75,20 @@ namespace QCAD
     std::string getEvalType() const;
   };
 
+  // HessianVec
+  template<typename Traits>
+  class EvaluatorTools<PHAL::AlbanyTraits::HessianVec, Traits>
```
bartgol, line 80: Here's another class that I'm not sure we should maintain.

src/responses/Albany_SolutionTwoNormResponseFunction.cpp
```diff
@@ -146,3 +146,71 @@ evaluateDistParamDeriv(
...
+    const double current_time,
+    const Teuchos::RCP<const Thyra_MultiVector>& v,
+    const Teuchos::RCP<const Thyra_Vector>& x,
+    const Teuchos::RCP<const Thyra_Vector>& xdot,
+    const Teuchos::RCP<const Thyra_Vector>& xdotdot,
+    const Teuchos::Array<ParamVec>& param_array,
+    const Teuchos::RCP<Thyra_MultiVector>& Hv_dp)
+{
+  if (!Hv_dp.is_null()) {
+    Hv_dp->assign(0.0);
```
bartgol, line 162: This is wrong, if I understand correctly what's happening. Hxx = d^2 norm(x)/ dx^2, which is not 0. The 1st derivative is x / |x|, the second derivative should be `1/||x|| *Id + x*x^T / ||x||^3 = Id/g + xx^T/g^3`, where Id is the identity and g is the response. So when you apply it to a vector v, the result is `Hxx_v = v/g + (v,x) x/g^3`, where (.,.) is the usual dot product. One dot product plus some saxby should do the trick.

bartgol:
I'm ok merging this without tests. The only thing I would ask is that we check in debug mode too. @kliegeois , can you please try to comment out the `lclSuccess` test in Tpetra_FECrsGraph_def.hpp, and try to run in debug mode again? I just want to be sure we're not masking other potential issues.

src/Albany_ModelEvaluator.cpp
```diff
@@ -928,6 +965,65 @@ evalModelImpl(const Thyra_InArgs&  inArgs,
...
     }
 
+    // Need to handle hess_vec_prod_g
+
+    const Teuchos::RCP<const Thyra_MultiVector> delta_x = inArgs.get_x_direction();
+    const Teuchos::RCP<Thyra_MultiVector> g_hess_xx_v =
+      outArgs.supports(Thyra_ModelEvaluator::OUT_ARG_hess_vec_prod_g_xx, j) ?
+      outArgs.get_hess_vec_prod_g_xx(j) : Teuchos::null;
+
+    if (Teuchos::nonnull(delta_x) && Teuchos::nonnull(g_hess_xx_v)) {
```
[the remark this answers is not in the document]
kliegeois: Do you suggest that if one is `Teuchos::null` and not the other, an error should be thrown?
I think that Trilinos does not check that yet.

src/evaluators/gather/PHAL_GatherScalarNodalParameter_Def.hpp
kliegeois: The reason is that, we are using the  formula where  is a scalar to compute the Hessian-vector product contributions. 

Therefore, we have two nested derivatives and the outermost one (mathematically), i.e. the one with respect to , is the one related to the innermost `FAD` type. We know that this innermost `FAD` type will always have a size of one and only `Dx(0)` has to be specified. It is clearer?

src/evaluators/gather/PHAL_GatherScalarNodalParameter_Def.hpp
kliegeois: Same answer as above.

src/evaluators/response/QCAD_ResponseFieldValue.hpp
kliegeois: I have no strong feeling regarding this class and the other commented one.
I had to provide some implementation (even empty) to build and link correctly.
If they are not used, I assume that they can be removed. Should it be included in this PR or in another one?

src/evaluators/utility/QCAD_EvaluatorTools.hpp
kliegeois: Same answer as above.

src/responses/Albany_SolutionTwoNormResponseFunction.cpp
kliegeois: I see what you mean. Thanks for highlighting this issue!

kliegeois:
Thanks a lot for your review @bartgol and @jewatkins !
@bartgol, thanks for highlighting the issue and your suggestion regarding the debug build.
I will implement the modifications discussed in the comments and verify whether the ctest tests pass with the debug build.

src/Albany_ModelEvaluator.cpp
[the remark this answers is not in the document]
bartgol: Yeah, I was thinking about that. It might be a sign of something wrong in our setup (or in Trilinos setup). Perhaps a warning would be enough?

src/evaluators/gather/PHAL_GatherScalarNodalParameter_Def.hpp
bartgol: Oh, the inner most Fad does not have a length equal to the number parameters, but rather to the current parameter "vector dimension"?

src/evaluators/response/QCAD_ResponseFieldValue.hpp
bartgol: Perhaps another one. However, if these classes give you headaches later on, you may consider removing them. Unless @mperego thinks otherwise.

src/evaluators/gather/PHAL_GatherScalarNodalParameter_Def.hpp
kliegeois: Oh, I think that I have misunderstood what you meant, especially when reading `Dx(param_index)`.
Here, up to now, we only evaluate one contribution of the Hessian-vector product at a time when we are in that `evaluateFields` function. This means that, somehow, we have basically only one current parameter "vector dimension", yes.

Later, we might consider evaluate multiple contributions together, it will increase the sizes of the `FAD` types but it could be more cost effective.

kliegeois:
@bartgol, I made the discussed modifications. For the two norm, I think that the formula that you gave has a small signed error and that the + should be a -. Do you agree?
The tests passed both with a release and a debug build.

src/Albany_ModelEvaluator.cpp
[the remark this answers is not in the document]
mperego: I think we should throw an error when `g_hess_xx_v` is nonnull (which means that Thyra wants the hessian vec product to be evaluated) but `delta_x` is null. It's fine if `delta_x` is nonnull when  `g_hess_xx_v` is null (that would be the case if you want to compute   `g_hess_xp_v` but not `g_hess_xx_v`) . So I would do replace line 975 with 
`if (Teuchos::nonnull(g_hess_xx_v)) {`
and within the if we throw an error if `delta_x` is null.

src/evaluators/response/QCAD_ResponseFieldValue.hpp
mperego: Yeah, @bartgol let's look at those QCAD evaluators after this PR is merged in. We should either remove them or rename them depending on whether we used them or not. I can have a look at that.

src/responses/Albany_ScalarResponseFunction.cpp
```diff
@@ -18,6 +18,16 @@ ScalarResponseFunction::responseVectorSpace() const
   return createLocallyReplicatedVectorSpace(num_responses,comm);
 }
 
+Teuchos::RCP<const Thyra_VectorSpace>
+ScalarResponseFunction::responseDualVectorSpace() const
```
mperego, line 22: I this ever used? Also shouldn't it be the scalar 1, instead of 0?

src/responses/Albany_AbstractResponseFunction.hpp
```diff
@@ -34,6 +34,9 @@ namespace Albany {
     //! Get the vector space associated with this response.
     virtual Teuchos::RCP<const Thyra_VectorSpace> responseVectorSpace() const = 0;
 
+    //! Get the dual vector space associated with this response.
+    virtual Teuchos::RCP<const Thyra_VectorSpace> responseDualVectorSpace() const = 0;
```
mperego, line 38: Is the responseDualVecSpace ever used?

src/responses/Albany_SolutionTwoNormResponseFunction.cpp
```diff
@@ -146,3 +146,76 @@ evaluateDistParamDeriv(
...
+    const Teuchos::RCP<const Thyra_Vector>& xdotdot,
+    const Teuchos::Array<ParamVec>& param_array,
+    const Teuchos::RCP<Thyra_MultiVector>& Hv_dp)
+{
+  Teuchos::ScalarTraits<ST>::magnitudeType nrm = x->norm_2();
+  Teuchos::ScalarTraits<ST>::magnitudeType nrm3 = std::pow(nrm,3);
+
+  if (!Hv_dp.is_null()) {
+    for (int j=0; j<Hv_dp->domain()->dim(); ++j) {
+      Thyra::V_StVpStV(Hv_dp->col(j).ptr(),1.0/nrm,*v->col(j),-1.0*x->dot(*v->col(j))/nrm3,*x);
```
mperego, line 166: Can you please add a comment  with a "readable" formula?

src/evaluators/gather/PHAL_GatherScalarNodalParameter_Def.hpp
mperego: Would you mind adding a comment in the code explaining what you are doing? I know that there are very little comments in Albany but we are trying to add more comments as we go. Maybe there are other places where a few words of comments would hlep understanding what's implemented. Thanks!

src/responses/Albany_AbstractResponseFunction.hpp
kliegeois: No, it is not. I forgot to remove that when working on the Thyra objects. I am removing it.

kliegeois:
Thanks @mperego for your review!

I made the proposed modifications, I am currently running the ctests both with a release and a debug build and I push the commits if everything is fine!

kliegeois:
@mperego all the tests passed!

Are the changes fine to both of you @bartgol @mperego ?

mperego:
> @mperego all the tests passed!
> 
> Are the changes fine to both of you @bartgol @mperego ?

Looks good to me! Thanks.

src/responses/Albany_SolutionTwoNormResponseFunction.cpp
```diff
@@ -146,3 +146,77 @@ evaluateDistParamDeriv(
...
+evaluateDistParamHessVecProd_xx(
+    const double current_time,
+    const Teuchos::RCP<const Thyra_MultiVector>& v,
+    const Teuchos::RCP<const Thyra_Vector>& x,
+    const Teuchos::RCP<const Thyra_Vector>& xdot,
+    const Teuchos::RCP<const Thyra_Vector>& xdotdot,
+    const Teuchos::Array<ParamVec>& param_array,
+    const Teuchos::RCP<Thyra_MultiVector>& Hv_dxdx)
+{
+  Teuchos::ScalarTraits<ST>::magnitudeType nrm = x->norm_2();
```
bartgol, line 161: You might move these inside the if statement.

kliegeois: Yes that would be better! I should even include a second test which checks if the norm is not zero before using the formula.

kliegeois:
@bartgol and @mperego thanks a lot for your second review!

src/responses/Albany_SolutionTwoNormResponseFunction.cpp
```diff
@@ -158,13 +158,14 @@ evaluateDistParamHessVecProd_xx(
...
 {
-  Teuchos::ScalarTraits<ST>::magnitudeType nrm = x->norm_2();
-  Teuchos::ScalarTraits<ST>::magnitudeType nrm3 = std::pow(nrm,3);
-
   if (!Hv_dxdx.is_null()) {
-    for (int j=0; j<Hv_dxdx->domain()->dim(); ++j) {
-      // Evaluate Hv_dxdx_j = v_j/||x|| - (v_j,x) x/||x||^3 where (.,.) is the usual dot product:
-      Thyra::V_StVpStV(Hv_dxdx->col(j).ptr(),1.0/nrm,*v->col(j),-1.0*x->dot(*v->col(j))/nrm3,*x);
+    Teuchos::ScalarTraits<ST>::magnitudeType nrm = x->norm_2();
+    if (nrm!=0.0) {
```
bartgol, line 163: I didn't mean to check if the nrm is indeed zero, only to put the norm related stuff inside the HV_dxdx nonnull test.

That said, it is true that if norm=0, then the formula looks bad. Should we throw if norm=0? My assumption is that all these derivatives, once multiplied with other queantities in Piro, ROL, or whatever, do not yield issues with norm=0 _in the analytical formulation_. Something along the line of a=x/|x|, b=x, then ab is well defined at 0, but a is not.

I don't think Albany can know how the Hessian derivatives will be used, so maybe we _should_ error out if the norm is 0?

mperego: Not sure what's best honestly. It's true that if the norm is zero, which could happen, we don't necessarily want the code to throw an error, or even a NAN... A similar case happened when computing the derivative of sqrt(x), at x=0.

bartgol: Are you sure you don't want to throw if the derivatives are NaN or Inf? I'm ok with it, but then we need Piro to be fully prepared to handle "exceptional" cases... If Piro is not geared up to accept Inf or NaN (and decide whether to throw or patch up things), then we _should_ throw. Imho.

mperego: OK, let's throw.

kliegeois:
@mperego thanks for the merge!

PR #621, opened 2020-08-04
Fecrs debug fix

bartgol:
This PR fixes the crashes we get in debug mode related to Tpetra::FECrsGraph.

Note: this PR *requires* trilinos/Trilinos#7773 to be merged in order to work (i.e., I compiled Albany against the branch of that PR).

On my laptop, I manually run one test that failed before, and it now passes. I will let the nigthlies check that everything is fine after we merge this (which, again, needs to happen after the aforementioned PR is merged in Trilinos).

### DIFF
--- a/src/utility/Albany_ThyraCrsMatrixFactory.cpp
+++ b/src/utility/Albany_ThyraCrsMatrixFactory.cpp
@@ -229,8 +229,17 @@ void ThyraCrsMatrixFactory::fillComplete () {
     nnz_per_row.modify_host();
     nnz_per_row.sync_device();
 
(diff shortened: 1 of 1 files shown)


bartgol:
Well, assuming the size of domains isn't too small, we should not have huge penalties. But I agree that it's something we should check.

Also, in the long run, we should fix the node sharing (or element sharing) on the stk meshes at init time, when we load/create the mesh.

mperego:
@bartgol The Trilinos PR has been merged. Should we merge this?

bartgol:
I think it's safe to do so, yeah. Let me do it.

PR #622, opened 2020-08-04
Enthalpy Kokkos Refactor (Boundary Conditions)

mcarlson801:
This pull request contains the Kokkos port of the Land Ice Enthalpy problem’s boundary condition evaluators and the new MDField Layouts necessary to attain high performance on GPUs.

1) Added “collapsed” layouts for sidesets that result in coalesced memory access on GPUs and improved caching on CPUs. Collapsed layouts are noted by the “_sideset” suffix in Albany_Layouts.hpp.

2) Side sets were originally represented by vectors of SideStructs that contained cell and side index information. This has been replaced by a GlobalSideStruct that contains all of the sideset index information for all side set elements over all worksets in Kokkos views. A secondary LocalSideStruct then contains Kokkos subviews that are limited to their assigned workset. The original SideStruct is still there in order to ensure compatibility with problems that have not yet been updated to use the new layouts.

3) All interpolation evaluators used in the LandIce_Enthalpy problem have been updated to accept the new or the legacy layouts. These evaluators have also been ported to Kokkos and now run on the GPU for cuda builds.

4) LoadSideSetStateField has also been updated to accept either the new or legacy layouts. This change includes an update to registerSideSetStateVariable in order to select which layout to use.

5) Enthalpy specific evaluators BasalMeltRate and EnthalpyBasalResid have been updated to accept either new or legacy layouts and have been ported to Kokkos and now run on the GPU for cuda builds.

I'm still working on cleaning up this branch (removing commented code, adding comments) and will be pushing more changes to this branch tomorrow morning but the overall structure is in place and is passing all tests.


src/Albany_StateManager.cpp
```diff
@@ -428,22 +428,204 @@ Albany::StateManager::registerSideSetStateVariable(
...
+    return p;  // Don't re-register the same state name
+  }
+
+  sideSetStatesToStore[sideSetName][ebName][stateName] = dl;
+
+  if (sideSetStateInfo.find(sideSetName) == sideSetStateInfo.end()) {
+    // It's the first time we register states on this side set, so we initiate
+    // the pointer
+    // TODO, when compiler allows, replace following with this for performance:
+    // sideSetStateInfo.emplace(sideSetName,Teuchos::rcp(new
```
mperego, line 506: I think that now we can safely use emplace instead of insert

src/LandIce/evaluators/LandIce_EnthalpyBasalResid_Def.hpp
```diff
@@ -50,25 +52,71 @@ EnthalpyBasalResid(const Teuchos::ParameterList& p, const Teuchos::RCP<Albany::L
...
 
+template<typename EvalT, typename Traits, typename Type>
+KOKKOS_INLINE_FUNCTION
+void EnthalpyBasalResid<EvalT,Traits,Type>::
+operator() (const Enthalpy_Basal_Residual_Collapsed_Tag& tag, const int& sideSet_idx) const{
+  
+  const int cell = sideSet.elem_LID(sideSet_idx);
+  const int side = sideSet.side_local_id(sideSet_idx);
+
+  ScalarT val[4] = {0., 0., 0., 0.};
```
mperego, line 82: Maybe instead of using directly 4, we could define a  variable  `maxNumNodesPerSide = 4`.

src/Albany_StateManager.cpp
bartgol: Yes. Although that comment I put is a bit moot, since "performance" here is negligible. Still, using C++11 features is a good habit (and keeps the code a tad shorter).

src/Albany_StateManager.cpp
```diff
@@ -428,22 +428,204 @@ Albany::StateManager::registerSideSetStateVariable(
...
+    // rank
+    if (dl->rank() == 2)  // node scalar
+      nodalDataBase->registerVectorState(stateName, 1);
+    else if (dl->rank() == 3)  // node vector
+      nodalDataBase->registerVectorState(stateName, stateRef.dim[1]);
+    else if (dl->rank() == 4)  // node tensor
+      nodalDataBase->registerVectorState(
+          stateName, stateRef.dim[1] * stateRef.dim[2]);
+  } else {
+    // Strip one dimension out cause it's Side
```
bartgol, line 591: Max, isn't this impossible? I mean, if you use `registerSideSetState_collapsed`, you should *not* have the `Side` tag, no?

src/LandIce/evaluators/LandIce_BasalMeltRate_Def.hpp
```diff
@@ -16,38 +16,51 @@
 namespace LandIce
 {
 
+template<typename Type>
```
bartgol, line 19: Could you move this to a common header, like `Albany_Utils.hpp` or one of those?

src/LandIce/evaluators/LandIce_BasalMeltRate_Def.hpp
```diff
@@ -107,55 +120,173 @@ postRegistrationSetup(typename Traits::SetupData d, PHX::FieldManager<Traits>& f
...
+  const int dim = nodal ? numSideNodes : numSideQPs;
+
+  // Get the local data of side and cell
+  const int cell = sideSet.elem_LID(sideSet_idx);
+  const int side = sideSet.side_local_id(sideSet_idx);
+
+  for (int node = 0; node < dim; ++node) {
+    //always in presence of water on shelves (assuming that beta==0 <==> on shelves)
+    bool isThereWaterHere = isThereWater || (beta(sideSet_idx,node) == 0.0);
+    ScalarT diffEnthalpy = Enthalpy(sideSet_idx,node) - EnthalpyHs(sideSet_idx,node);
```
bartgol, line 137: Have you tried using `const auto` rather than `ScalarT` for (const) temporary expressions? I think it might make Sacado simply store the expression on the rhs, and Sacato will evaluate it once at the end (possibly doing some better optimization with temporaries when expanding the expression and evaluating it).

Also, have you tried `const auto&` when you just extract one entry? It might save some register usage on GPU.

src/LandIce/evaluators/LandIce_BasalMeltRate_Def.hpp
```diff
@@ -107,55 +120,173 @@ postRegistrationSetup(typename Traits::SetupData d, PHX::FieldManager<Traits>& f
...
+                                                                        ScalarT(0.5 + 0.5* basal_reg_coeff * diffEnthalpy);
+                                                                  //    ScalarT(0.5 + 0.5* (0.5-0.5*std::pow(1-basal_reg_coeff * diffEnthalpy,2)));
+
+    //mstar, [W m^{-2}] = [Pa m s^{-1}]: basal latent heat in temperate ice
+    ScalarT mstar = geoFluxHeat(cell,side,node);
+    for (int dim = 0; dim < vecDimFO; dim++)
+      mstar += 1000./scyr * beta(cell,side,node) * velocity(cell,side,node,dim) * velocity(cell,side,node,dim);
+
+    double dTdz_melting = beta_p * rho_i * g;
+    mstar += 1e-3* k_i * dTdz_melting;
```
bartgol, line 186: It seems to me that `1e-3*k_i*dTdx_melting` could be precomputed at construction time, no?

src/LandIce/evaluators/LandIce_BasalMeltRate_Def.hpp
```diff
@@ -107,55 +120,173 @@ postRegistrationSetup(typename Traits::SetupData d, PHX::FieldManager<Traits>& f
...
+                              "Side set views defined in input file but not properly specified on the mesh.\n");
   if (memoizer.have_saved_data(workset,this->evaluatedFields())) return;
-  int vecDimFO = 2;
-  double pi = atan(1.) * 4.;
-  ScalarT hom = homotopy(0);
-  ScalarT basal_reg_coeff = basalMelt_reg_alpha*exp(basalMelt_reg_beta*hom); // [adim]
-  ScalarT flux_reg_coeff = flux_reg_alpha*exp(flux_reg_beta*hom); // [adim]
 
-  const int dim = nodal ? numSideNodes : numSideQPs;
+  hom = homotopy(0);
```
bartgol, line 207: These three lines can be moved inside the if statement.

src/LandIce/evaluators/LandIce_BasalMeltRate_Def.hpp
```diff
@@ -107,55 +120,173 @@ postRegistrationSetup(typename Traits::SetupData d, PHX::FieldManager<Traits>& f
...
-      for (int node = 0; node < dim; ++node)
+    sideSet = workset.sideSetViews->at(basalSideName);
+  #ifdef ALBANY_KOKKOS_UNDER_DEVELOPMENT
+    if (useCollapsedSidesets) {
+      Kokkos::parallel_for(Basal_Melt_Rate_Collapsed_Policy(0, sideSet.size), *this);
+    } else {
+      Kokkos::parallel_for(Basal_Melt_Rate_Policy(0, sideSet.size), *this);
+    }
+  #else
+    const int dim = nodal ? numSideNodes : numSideQPs;
```
bartgol, line 221: Can be moved inside the if statement.

src/LandIce/evaluators/LandIce_BasalMeltRate_Def.hpp
```diff
@@ -107,55 +120,173 @@ postRegistrationSetup(typename Traits::SetupData d, PHX::FieldManager<Traits>& f
...
+          //always in presence of water on shelves (assuming that beta==0 <==> on shelves)
+          bool isThereWaterHere = isThereWater || (beta(cell,side,node) == 0.0);
+          ScalarT diffEnthalpy = Enthalpy(cell,side,node) - EnthalpyHs(cell,side,node);
+          ScalarT basal_reg_scale = (diffEnthalpy > 0 || !isThereWaterHere) ?  ScalarT(0.5 + 0.5*tanh(basal_reg_coeff * diffEnthalpy)) :
+                                                                              ScalarT(0.5 + 0.5* basal_reg_coeff * diffEnthalpy);
+                                                                      //    ScalarT(0.5 + 0.5* (0.5-0.5*std::pow(1-basal_reg_coeff * diffEnthalpy,2)));
+
+          //mstar, [W m^{-2}] = [Pa m s^{-1}]: basal latent heat in temperate ice
+          ScalarT mstar = geoFluxHeat(cell,side,node);
+          for (int dim = 0; dim < vecDimFO; dim++)
```
bartgol, line 273: I think the local var `dim` defined above will be shadowed by this loop counter. Not that it wrong, but should issue a compiler warning. If you move that local var inside the if statement (as I pointed out above), however, I think there should be no warning.

src/LandIce/evaluators/LandIce_BasalMeltRate_Def.hpp
```diff
@@ -107,55 +120,173 @@ postRegistrationSetup(typename Traits::SetupData d, PHX::FieldManager<Traits>& f
...
+            mstar += 1000./scyr * beta(cell,side,node) * velocity(cell,side,node,dim) * velocity(cell,side,node,dim);
```
bartgol, line 274: 1000./scyr could be pre-computed. Alternatively, you could make `scyr` a `static constexpr double` (and init it in the class), so that `1000.0/scyr` is computed at compile time. You might even consider pulling scyr outside the class, and just declare it locally in the fcn as a `constexpr double`, since it is not really part of the "state" of the object.

src/LandIce/evaluators/LandIce_BasalMeltRate_Def.hpp
```diff
@@ -107,55 +120,173 @@ postRegistrationSetup(typename Traits::SetupData d, PHX::FieldManager<Traits>& f
...
+
+          double dTdz_melting = beta_p * rho_i * g;
+          mstar += 1e-3* k_i * dTdz_melting;
```
bartgol, line 277: Same comment as above about the rhs being able to be precomputed.

src/LandIce/evaluators/LandIce_EnthalpyBasalResid_Def.hpp
bartgol: We are already computing it at construction time `nodeMax`. All we need is to store it in the class. However, we then need to make sure we ask kokkos for some scratch memory for val, since we cannot allocate dynamically inside a kernel.

bartgol: Actually, we don't need val to be an array at all. We can fuse the two loops over numSideNodes, and make val into an automatic variable (so that Sacado does not evaluate it until it's time to store it in the resid).

src/disc/Albany_DiscretizationUtils.hpp
```diff
@@ -37,9 +38,50 @@ class SideStruct
...
 
+
+// This is a stucture that holds all of the sideset information over all worksets. When running populate mesh,
+//   there can be a huge number of worksets with only a handful (1-20) of entries each, which causes a huge
+//   performance hit on GPUs since each of these tiny views must be allocated serially on the GPU. These global/local
+//   side structs avoid this issue by allocating once after computing the global view extents and then providing
+//   local subviews to each workset.
+// Memory layout is enforced to be LayoutRight regardless of architecture since Kokkos parallel sections
+//   will only ever operate on the last extent of any of these Views and therefore they should be contiguous.
+class GlobalSideStruct
```
bartgol, line 51: I think the names of these structs are a bit misleading. Their name made me think they contain info on a single side, while instead they contain info of all sides in the current sideset (and global also spans all worksets). Maybe something like `GlobalSideSetInfo` and `LocalSideSetInfo`?

Also, on a "taste" note (i.e., feel free to ignore): I would not add `Struct` to a class name. We did it elsewhere in Albany, but I'm not a big fan of that.

src/evaluators/interpolation/PHAL_DOFCellToSide.hpp
```diff
@@ -59,14 +62,66 @@ class DOFCellToSideBase : public PHX::EvaluatorWithBaseImpl<Traits>,
...
-    VERTEX_VECTOR
+    VERTEX_VECTOR,
+    VERTEX_VECTOR_SIDESET
   };
 
   LayoutType layout;
 
   MDFieldMemoizer<Traits> memoizer;
+
+  Kokkos::DynRankView<int, PHX::Device> dimsView;
```
bartgol, line 77: Is there a reason to use `DynRankView` over a simple 2d view? I mean, it's always going to be a 2d view, right? Seeing a DynRankView makes me think that we don't know the rank a priori, which confuses me.

src/evaluators/interpolation/PHAL_DOFCellToSide.hpp
```diff
@@ -59,14 +62,66 @@ class DOFCellToSideBase : public PHX::EvaluatorWithBaseImpl<Traits>,
...
+
+public:
+
+  typedef Kokkos::View<int***, PHX::Device>::execution_space ExecutionSpace;
```
bartgol, line 81: You can get the exec space out of `PHX::Device` directly, keeping the code more readable.

bartgol: Also, and this is another 'style/taste' comment, why not using...using (sorry for the pun)? Like

`using ExecutionSpace = ...;`

It's the C++11 way, and I think it's easier to read. Again, this might be a taste issue, so feel free to ignore.

src/evaluators/interpolation/PHAL_DOFCellToSide.hpp
```diff
@@ -59,14 +62,66 @@ class DOFCellToSideBase : public PHX::EvaluatorWithBaseImpl<Traits>,
...
+  struct DOFCellToSide_CellScalar_Tag{};
```
bartgol, line 82: Another "taste" suggestion: if we remove `DOFCellToSide_` from the tag names, as well as the policy names, we keep the code slimmer and more readable.

src/evaluators/interpolation/PHAL_DOFCellToSide_Def.hpp
```diff
@@ -117,24 +140,230 @@ postRegistrationSetup(typename Traits::SetupData d,
...
+template<typename EvalT, typename Traits, typename ScalarT>
+KOKKOS_INLINE_FUNCTION
+void DOFCellToSideBase<EvalT, Traits, ScalarT>::
+operator() (const DOFCellToSide_NodeScalar_Tag& tag, const int& sideSet_idx) const {
+
+  // Get the local data of side and cell
+  const int cell = sideSet.elem_LID(sideSet_idx);
+  const int side = sideSet.side_local_id(sideSet_idx);
+
+  for (int node=0; node<dimsView(2); ++node) {
```
bartgol, line 207: I'm not sure the compiler can optimize and store `dimsView(2)` once and for all. It should be able to prove it's constant *if* the memory traits of the view include the restrict attribute, but I don't think they do. Storing it locally might work better.

src/evaluators/interpolation/PHAL_DOFCellToSide_Def.hpp
```diff
@@ -117,24 +140,230 @@ postRegistrationSetup(typename Traits::SetupData d,
...
+      val_side(cell,side,node) = val_cell(cell,sideNodes(side,node));
```
bartgol, line 208: Would subviewing the input view as (cell,side,ALL) and the output view at (cell,ALL) improve performance by reducing indexing time?

src/evaluators/interpolation/PHAL_DOFVecInterpolationSide.hpp
```diff
@@ -45,19 +45,38 @@ class DOFVecInterpolationSideBase : public PHX::EvaluatorWithBaseImpl<Traits>,
...
   int vecDim;
+  bool useCollapsedSidesets;
 
   MDFieldMemoizer<Traits> memoizer;
+
+  Albany::LocalSideStruct sideSet;
+
+public:
+
+  typedef Kokkos::View<int***, PHX::Device>::execution_space ExecutionSpace;
```
bartgol, line 67: Same comments about types as in DOFCellToSide

src/evaluators/utility/PHAL_ComputeBasisFunctionsSide_Def.hpp
```diff
@@ -116,32 +118,199 @@ postRegistrationSetup(typename Traits::SetupData d,
...
+
+  // Get the local data of side and cell
+  const int cell = sideSet.elem_LID(sideSet_idx);
+  const int side = sideSet.side_local_id(sideSet_idx);
+
+  // Computing tangents (the basis for the manifold)
+  for (int itan=0; itan<numSideDims; ++itan) {
+    for (int icoor=0; icoor<numCellDims; ++icoor) {
+      for (int qp=0; qp<numSideQPs; ++qp) {
+        tangents(sideSet_idx,qp,icoor,itan) = 0.;
```
bartgol, line 172: It might be beneficial to subview at least the sideSet_idx dimension.

We might even fuse all the qp loops into a big loop (and maybe add that dimension to the RangePolicy) and subview both sideSet_idx and qp?

src/LandIce/evaluators/LandIce_BasalMeltRate_Def.hpp
mcarlson801: This was mentioned for the volume evaluator PR as well. In that PR, I changed all of the scaling constants like this to be precomputed and it had no measurable effect on the performance. My understanding is that CUDA is smart enough to optimize these kinds of things away. I think having them like this is more readable but it might not be portable. I'll change these to be precomputed for the commit I'm working on but I think it might not be necessary.

src/evaluators/interpolation/PHAL_DOFCellToSide.hpp
mcarlson801: I agree that would be much nicer to look at it, I'll change these

src/LandIce/evaluators/LandIce_BasalMeltRate_Def.hpp
mcarlson801: Woops! This got left in from a previous commit. These functions have already been put in a new file called Albany_KokkosUtils.hpp that was included with the Volume PR.

src/LandIce/evaluators/LandIce_BasalMeltRate_Def.hpp
bartgol: On CUDA, flops are basically free, but all of these things might add up in CPU builds.

src/Albany_StateManager.cpp
mcarlson801: Yep, the comment just got left in but no dimension gets stripped in registerSideSetStateVariable_collapsed. I'll update the comments there, thanks.

src/LandIce/evaluators/LandIce_BasalMeltRate_Def.hpp
jewatkins: I think you mentioned this in the previous PR. I think I tried this a few years ago and it results in an error when using DFad and/or SFad and/or on GPU. It's worth trying again though but maybe we should wait until we make another pass at optimizing the code.

The benefit on GPU will depend on a few things. If the memory access is contiguous in global memory, it probably makes sense to not store the temp. If it's not, then it's probably better to store in local memory.

src/LandIce/evaluators/LandIce_BasalMeltRate_Def.hpp
jewatkins: This might get optimized out on a CPU build too. But I agree, it's probably better just to pull this out of the kernel just in case.

src/LandIce/evaluators/LandIce_BasalMeltRate_Def.hpp
jewatkins: There's a dim in the else clause too.

src/LandIce/evaluators/LandIce_BasalMeltRate_Def.hpp
jewatkins: It's probably best to change `dim` to something like `numPts` just to avoid confusion.

src/LandIce/evaluators/LandIce_EnthalpyBasalResid_Def.hpp
jewatkins: Let's change this to `constexpr int maxNumNodesPerSide = 4` for clarity. We can think about fusing loops later.

src/evaluators/interpolation/PHAL_DOFCellToSide.hpp
jewatkins: I think we can do a large sweep and change all instances of this to `using ExecutionSpace = PHX::Device::execution_space` later.

src/evaluators/interpolation/PHAL_DOFCellToSide_Def.hpp
jewatkins: I agree, I think we can store these locally using a series of if statements in the constructor.

src/evaluators/interpolation/PHAL_DOFCellToSide_Def.hpp
jewatkins: I think it would but it might also increase code complexity.

src/evaluators/state/PHAL_LoadSideSetStateField_Def.hpp
```diff
@@ -99,17 +101,22 @@ evaluateFields(typename Traits::EvalData workset)
   TEUCHOS_TEST_FOR_EXCEPTION (size>2 && tag2!="Node" && tag2!="Dim" && tag2!="VecDim", std::logic_error,
                               "Error! Invalid field layout in LoadSideSetStateField.\n");
 
+  // Temporary tag for new sideset layout
+  // TODO: replace this check
+  const std::string& tag3 = size>1 ? field.fieldTag().dataLayout().name(1) : "";
```
jewatkins, line 106: Can you make these names more descriptive?

src/Albany_StateManager.cpp
bartgol: Rather than simply update the comment, I would replace the whole body of the else statement with an exception. If someone calles the `_collapsed` version with the wrong layout, we should throw.

src/evaluators/utility/PHAL_ComputeBasisFunctionsSide.hpp
```diff
@@ -64,20 +64,38 @@ class ComputeBasisFunctionsSide : public PHX::EvaluatorWithBaseImpl<Traits>,
...
+  PHX::MDField<MeshScalarT>                               inv_metric;
+  PHX::MDField<RealType>                                  BF;
+  PHX::MDField<MeshScalarT>                               GradBF;
   PHX::MDField<MeshScalarT,Cell,Side,QuadPoint,Dim>       normals;
 
-  std::vector<std::vector<int> > sideNodes;
-  std::vector<Kokkos::DynRankView<int, PHX::Device>> cellsOnSides;
-  std::vector<int> numCellsOnSide;
+  //std::vector<Kokkos::DynRankView<int, PHX::Device>> cellsOnSides;
+  //std::vector<int> numCellsOnSide;
```
jewatkins, line 77: Remove all cell on sides comments if we're not going to use them anymore. There's some in the definition too.

src/evaluators/interpolation/PHAL_DOFCellToSide.hpp
mcarlson801: There are a bunch of existing evaluators where DynRankViews were used for Views with known rank. I was't sure if there was a purpose for doing it like this so I figured I'd just stay consistent with the rest of the code unless someone mentioned otherwise.

bartgol: I think those evaluators are mostly in the BC framework, or anywhere where you need to call Intrepid2 stuff (@mperego can confirm/deny). In those places, it is true that the rank is not known a priori (Intrepid2 works with several layouts, and uses the dyn rank as a shortcut instead of writing several functions, IIRC), so it is "needed".
But if you do know the rank a priori, then there's no need for a dyn rank view (and it might actually be misleading). I'd stick with normal View's if/where possible.

jewatkins: Let's try to get rid of this completely and use a static array. It looks like 5 is the max size?

src/evaluators/interpolation/PHAL_DOFCellToSide.hpp
```diff
@@ -40,9 +40,12 @@ class DOFCellToSideBase : public PHX::EvaluatorWithBaseImpl<Traits>,
 private:
 
   std::string                     sideSetName;
-  std::vector<std::vector<int> >  sideNodes;
   std::vector<int>                dims;
 
+  Kokkos::DynRankView<int, PHX::Device> sideNodes;
```
jewatkins, line 45: I think this one can safely be a `Kokkos::View`. If I remember correctly, DynRankView is required for AD for some reason but this is just used for indexing so it should be safe to change.

src/evaluators/interpolation/PHAL_DOFCellToSide.hpp
mcarlson801: It's a 2d view of size (numSides, numNodes). I assume the maximum number of sides is 6 (prism/wedge?), and the maximum number of nodes on a side is 4?

jewatkins: this one is a 1D view of dims and it looks like dims[4] is the largest value used in the kernel

mcarlson801: Yeah, I was about to edit, I thought this was in reference to sideNodes. I will make this one a static array of size 5.

mcarlson801:
I've pushed a few new commits to this branch to address the comments from everyone's reviews'. There are a few things that I did not address since the will be saved for the upcoming optimization pass.
The current version is passing all tests and the performance for the serial case is equivalent to master so I think it is good to go unless anyone has any further comments or suggestions.

mperego:
> 
> 
> I've pushed a few new commits to this branch to address the comments from everyone's reviews'. There are a few things that I did not address since the will be saved for the upcoming optimization pass.
> The current version is passing all tests and the performance for the serial case is equivalent to master so I think it is good to go unless anyone has any further comments or suggestions.

I'm fine w/ the changes. Thanks for this work!

src/evaluators/utility/PHAL_ComputeBasisFunctionsSide.hpp
jewatkins: I think these comments still need to be removed.

src/evaluators/interpolation/PHAL_DOFCellToSide_Def.hpp
```diff
@@ -117,24 +140,229 @@ postRegistrationSetup(typename Traits::SetupData d,
 
   val_side.dimensions(dims);
 
+  for (int i = 0; i < dims.size(); ++i)
+    dimsArray[i] = dims[i];
```
jewatkins, line 144: Can you throw an error here if dims.size() is larger than 5?

bartgol:
Looks good. Feel free to rewrite git history in your branch, if you think it makes things cleaner (like squashing some mini-commits that fixed stuff that should have happened in a previous commit), then merge away.

bartgol:
Btw, do you have two github user names?

jewatkins:
A couple more minor things but everything else looks good. weaver testing is up starting today I think so maybe we should wait until Monday to push to get a couple of data points without this PR.

src/evaluators/utility/PHAL_ComputeBasisFunctionsSide.hpp
mcarlson801: Woops, I thought I got all of them. I'll do another pass to make sure they all get removed.

src/evaluators/interpolation/PHAL_DOFCellToSide_Def.hpp
mcarlson801: Sure, I'll get that added.

mcarlson801:
> Btw, do you have two github user names?

It should just be the mcarlson801 one being used. I might have used my old user info on one of the machines I made changes on. I'll double check.

mcarlson801:
> A couple more minor things but everything else looks good. weaver testing is up starting today I think so maybe we should wait until Monday to push to get a couple of data points without this PR.

That sounds good to me. It's also great to hear that Weaver is finally in business!

Issue #590, opened 2020-05-04
Rebalance Logic


mperego:
Rebalance is always turned on when `Use Serial Mesh` is `true`.
```
  if(rebalance || (useSerialMesh && comm->getSize() > 1)) {
     rebalanceAdaptedMeshT(params, comm);
```
https://github.com/SNLComputation/Albany/blob/master/src/disc/stk/Albany_GenericSTKMeshStruct.cpp#L639

In fact rebalance can change the number of worksets  per process. Even if `Workset Size` is set ot `-1`, one could have multiple workset per process. I encountered this with extruded meshes using `ExtrudedSTKMeshStruct`. This could be a problem,e.g. for memoization. 
I propose to turn on rebalance only when  `Rebalance Mesh` is true, without assuming that one always wants rebalance also when `Use Serial Mesh` is true. 

@bartgol @gahansen

bartgol:
If you have 'Use Serial Mesh' and you don't rebalance, don't you end up with the mesh only on one processor? That was my understanding: rebalance cause the (serial) mesh is all on rank 0.

mperego:
No.. not sure how it works but turning off re-balancing doesn't make the run serial.

bartgol:
So even with rebalancing off, dofs are divided among ranks?

mperego:
Yes.. unless I'm going crazy (which is always a possibility)

bartgol:
Well, if that's the case, then, as you said, we should change the code, and only do rebalance if explicitly requested.

gahansen:
I think @agsalin converted Albany from an approach that I wrote long ago, where we used rebalance to "spread" the mesh across multiple processors after we input the serial mesh on PE 0. It looks like Andy converted the code to use Ioss's parallel I/O capabilities to do this instead, I suspect that we do not need or want to have the rebalance word in there any longer. Indeed, I suspect rebalance is only potentially useful if we either use adaptivity (via percept), or if the physics balance changes over time. I be glad to take a quick, deeper look and this and prune it (and percept, along with other adaptivity logic that is likely stale now). We can revisit a new rebalance later if needed.

bartgol:
I would keep the rebalance cabability. It seems the only modification to do is a one line change in the GenericSTKMeshStruct, removing ` || (...)` from the if statement.

mperego:
@bartgol I tried to remove the `|| (...)` from the if statement. The tests work except from yours on Hydrology ones. I get this error:
```
[...]

PR #626, opened 2020-08-14
Do mesh rebalance only when explicitely requested

mperego:
Addresses issue #590

### DIFF
--- a/src/disc/stk/Albany_GenericSTKMeshStruct.cpp
+++ b/src/disc/stk/Albany_GenericSTKMeshStruct.cpp
@@ -447,7 +447,9 @@ void GenericSTKMeshStruct::rebalanceInitialMeshT(const Teuchos::RCP<const Teucho
   bool rebalance = params->get<bool>("Rebalance Mesh", false);
   bool useSerialMesh = params->get<bool>("Use Serial Mesh", false);
 
-  if(rebalance || (useSerialMesh && comm->getSize() > 1)) {
+  if(rebalance) {
+    TEUCHOS_TEST_FOR_EXCEPTION (this->side_maps_present, std::runtime_error,
+                                "Error! Rebalance is not supported when side maps are present.\n");
     rebalanceAdaptedMeshT(params, comm);
   }
 }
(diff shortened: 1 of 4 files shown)


bartgol:
Before approving, I want to check why the rebalance is needed for the hydrology tests.

mperego:
> 
> 
> Before approving, I want to check why the rebalance is needed for the hydrology tests. In particular, why is the test working on my laptop _without_ asking for a rebalance.

@bartgol  Any updates on this? It's not  urgent but it would be good to be done with this.

bartgol:
> > Before approving, I want to check why the rebalance is needed for the hydrology tests. In particular, why is the test working on my laptop _without_ asking for a rebalance.
> 
> @bartgol Any updates on this? It's not urgent but it would be good to be done with this.

Sorry, I forgot. Can I do it tomorrow? Got a deadline that I was hoping to clear out today.

mperego:
> 
> 
> > Before approving, I want to check why the rebalance is needed for the hydrology tests. In particular, why is the test working on my laptop _without_ asking for a rebalance.
> 
> @bartgol Any updates on this? It's not urgent but it would be good to be done with this.

Sure, no rush.

mperego:
@bartgol weekly reminder :-)

bartgol:
Ok, I think GmshSTKMeshStruct _does_ need rebalance. The mesh is loaded only on proc 0, and there's no other logic capable of getting a distributed mesh than rebalancing. The IossSTKMeshStruct gets the rebalance/partitioning done on the fly by StkMeshIoBroker (the property `DECOMPOSITION_METHOD` is set in the Io broker, and it seems to be the property that, if set, triggers the decomposition).

So I think it's ok to make it so that Gmsh meshes are always rebalanced. This is currently not achieved, since Gmsh sets 'Use Serial Mesh' to true rather then setting 'Rebalance Mesh' to true. I'm going to make the micro mod, push, then approve the PR.

mperego:
> 
> 
> Ok, I think GmshSTKMeshStruct _does_ need rebalance. The mesh is loaded only on proc 0, and there's no other logic capable of getting a distributed mesh than rebalancing. The IossSTKMeshStruct gets the rebalance/partitioning done on the fly by StkMeshIoBroker (the property `DECOMPOSITION_METHOD` is set in the Io broker, and it seems to be the property that, if set, triggers the decomposition).
> 
> So I think it's ok to make it so that Gmsh meshes are always rebalanced. This is currently not achieved, since Gmsh sets 'Use Serial Mesh' to true rather then setting 'Rebalance Mesh' to true. I'm going to make the micro mod, push, then approve the PR.

OK! Thanks for looking into that and for working on the PR

bartgol (approved):

PR #630, opened 2020-09-04
Enable explicit traspose of Jacobian (for optimization problems)

mperego:
Explicitly transpose the jacobian when Piro sets to true the option
`Piro->Analysis->Optimization Status->Compute Transposed Jacobian`

### DIFF
--- a/src/Albany_ModelEvaluator.cpp
+++ b/src/Albany_ModelEvaluator.cpp
@@ -681,6 +681,7 @@ evalModelImpl(const Thyra_InArgs&  inArgs,
 
   //! If a parameter has changed in value, saved/unsaved fields must be updated
 
+  bool transposeJacobian = false;
   ObserverImpl observer(app);
   auto out = Teuchos::VerboseObjectBase::getDefaultOStream();
   auto& analysisParams = appParams->sublist("Piro").sublist("Analysis");
@@ -709,6 +710,7 @@ evalModelImpl(const Thyra_InArgs&  inArgs,
         iteration = iter;
       }
     }
+    transposeJacobian = opt_paramList.get("Compute Transposed Jacobian", false);
   }
 
   // Thyra vectors
@@ -842,6 +844,10 @@ evalModelImpl(const Thyra_InArgs&  inArgs,
         x, x_dot, x_dotdot,
         sacado_param_vec,
         f_out, W_op_out, dt);
+    if(transposeJacobian) {
+      Albany::transpose(W_op_out);
+    }
+
     f_already_computed = true;
   }
 
--- a/src/utility/Albany_ThyraUtils.cpp
+++ b/src/utility/Albany_ThyraUtils.cpp
@@ -552,7 +552,7 @@ void scale (const Teuchos::RCP<Thyra_LinearOp>& lop, const ST val)
 
 }
 
-Teuchos::RCP<Thyra_LinearOp> transpose (const Teuchos::RCP<const Thyra_LinearOp>& lop) {
+Teuchos::RCP<Thyra_LinearOp> getTransposedOp (const Teuchos::RCP<const Thyra_LinearOp>& lop) {
   Teuchos::RCP<Thyra_LinearOp> lopt;
 
   // Allow failure, since we don't know what the underlying linear algebra is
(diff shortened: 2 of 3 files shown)


mperego:
@bartgol can you have a quick look at this? I had to change your transpose function to directly transpose the jacobian passed as argument. I could not find a way to make you version work (it would not modify the jacobian). I can remove the original function if you prefer.

bartgol (approved):

bartgol:
Are you saying that the operator returned by `getTransposedOp` is the same as the input?

mperego:
> 
> 
> Are you saying that the operator returned by `getTransposedOp` is the same as the input?

If I do `W_op_out = Albany::getTransposedOp(W_op_out)` in the model evaluator, `W_op_out` is not changed. Maybe you would have to return a reference to the RCP pointer? I'm not sure why that happens

bartgol:
Was that Epetra or Tpetra?

mperego:
> 
> 
> Was that Epetra or Tpetra?

Either. I think I tried both of them.

bartgol:
Ok. I'd like to keep this version (where we don't alter the input), but if it's buggy, then we should delete it before someone else decides to use it.

mperego:
> Ok. I'd like to keep this version (where we don't alter the input), but if it's buggy, then we should delete it before someone else decides to use it.

I think the issue is not in that function. Now I remember what the issue was. `outArgs.get_W_op()` returns a `RCP<LinearOpBase>`, not a reference to it, so changing W_op_out does not have any effect to the value stored in `outArgs`.

mperego:
> 
> 
> > Ok. I'd like to keep this version (where we don't alter the input), but if it's buggy, then we should delete it before someone else decides to use it.
> 
> I think the issue is not in that function. Now I remember what the issue was. `outArgs.get_W_op()` returns a `RCP<LinearOpBase>`, not a reference to it, so changing `W_op_out` does not have any effect to the value stored in `outArgs`.

An I cannot do `*W_op_out = *Albany::getTransposedOp(W_op_out)` because the `LinearOpBase<>::operator()=` is private.

bartgol:
> > > Ok. I'd like to keep this version (where we don't alter the input), but if it's buggy, then we should delete it before someone else decides to use it.
> > 
> > 
> > I think the issue is not in that function. Now I remember what the issue was. `outArgs.get_W_op()` returns a `RCP<LinearOpBase>`, not a reference to it, so changing `W_op_out` does not have any effect to the value stored in `outArgs`.
> 
> An I cannot do `*W_op_out = *Albany::getTransposedOp(W_op_out)` because the `LinearOpBase<>::operator()=` is private.

Oh, but you could call `set_W_op(getTransposedOp(W_op_out)`, no?

mperego:
> 
> 
> > > > Ok. I'd like to keep this version (where we don't alter the input), but if it's buggy, then we should delete it before someone else decides to use it.
> > > 
> > > 
> > > I think the issue is not in that function. Now I remember what the issue was. `outArgs.get_W_op()` returns a `RCP<LinearOpBase>`, not a reference to it, so changing `W_op_out` does not have any effect to the value stored in `outArgs`.
> > 
> > 
> > An I cannot do `*W_op_out = *Albany::getTransposedOp(W_op_out)` because the `LinearOpBase<>::operator()=` is private.
> 
> Oh, but you could call `set_W_op(getTransposedOp(W_op_out)`, no?



> set_W_op(getTransposedOp(W_op_out)

well, but `outArgs` is `const`

bartgol:
> > > > > Ok. I'd like to keep this version (where we don't alter the input), but if it's buggy, then we should delete it before someone else decides to use it.
> > > > 
> > > > 
> > > > I think the issue is not in that function. Now I remember what the issue was. `outArgs.get_W_op()` returns a `RCP<LinearOpBase>`, not a reference to it, so changing `W_op_out` does not have any effect to the value stored in `outArgs`.
> > > 
> > > 
> > > An I cannot do `*W_op_out = *Albany::getTransposedOp(W_op_out)` because the `LinearOpBase<>::operator()=` is private.
> > 
> > 
> > Oh, but you could call `set_W_op(getTransposedOp(W_op_out)`, no?
> 
> > set_W_op(getTransposedOp(W_op_out)
> 
> well, but `outArgs` is `const`

Well, f... =)

Ok, then let's keep the other transpose function, cause in other contexts it might still be helpful maybe?

mperego:
> 
> 
> > > > > > Ok. I'd like to keep this version (where we don't alter the input), but if it's buggy, then we should delete it before someone else decides to use it.
> > > > > 
> > > > > 
> > > > > I think the issue is not in that function. Now I remember what the issue was. `outArgs.get_W_op()` returns a `RCP<LinearOpBase>`, not a reference to it, so changing `W_op_out` does not have any effect to the value stored in `outArgs`.
> > > > 
> > > > 
> > > > An I cannot do `*W_op_out = *Albany::getTransposedOp(W_op_out)` because the `LinearOpBase<>::operator()=` is private.
> > > 
> > > 
> > > Oh, but you could call `set_W_op(getTransposedOp(W_op_out)`, no?
> > 
> > 
> > > set_W_op(getTransposedOp(W_op_out)
> > 
> > 
> > well, but `outArgs` is `const`
> 
> Well, f... =)
> 
> Ok, then let's keep the other transpose function, cause in other contexts it might still be helpful maybe?

sure. I'm going to merge this PR.

PR #634, opened 2020-09-18
Add UnitTests for gatherSolution and gatherDistributedParameters with the HessianVec EvaluationType.

kliegeois:
This PR includes 3 modifications related to the Hessian-vector products computation with AD:
- A modification of `src/Albany_ModelEvaluator.cpp` not to use `setHessianSupports`.
- The correction of a bug when `diff2 == 0` and that we want to compute derivatives of `weightedDiff` in `src/LandIce/evaluators/LandIce_ResponseSurfaceVelocityMismatch_Def.hpp`.
- The addition of UnitTests for `gatherSolution` and `gatherDistributedParameters` with the `HessianVec EvaluationType`. Those tests relie on Trilinos PR 7952 [https://github.com/trilinos/Trilinos/pull/7952](url) to test all the derivatives of the AD field.

There are no UnitTests for the scatter evaluators yet.

We might want to discuss how UnitTest should be ordered/separated in files/folders in the `tests/unit/evaluators` folder.
Currently, the tests added are just added to the existing `tests/unit/evaluators/EvaluatorUnitTester.cpp`. This is not viable in the longer term and they should be separated into multiple files to reduce compilation time and to have more readable files.

I did not took the decision on how we should organize the UnitTests as I do not know what you prefer and the plan related to those tests.
Do we eventually want all the evaluators to be tested using `PHX::EvaluatorUnitTester` for all the available evaluation types?


src/Albany_ModelEvaluator.cpp
```diff
@@ -583,8 +583,6 @@ Thyra_OutArgs ModelEvaluator::createOutArgsImpl() const
         i + num_param_vecs,
         Thyra_ModelEvaluator::DERIV_LINEAR_OP);
 
-  result.setHessianSupports(true);
```
mperego, line 586: What was happening until now for the Hessian of f? Was it computed by ROL with finite differences?

src/Albany_ModelEvaluator.cpp
```diff
@@ -619,28 +617,49 @@ Thyra_OutArgs ModelEvaluator::createOutArgsImpl() const
     bool hess_vec_prod_g_px_support = aDHessVec;
     bool hess_vec_prod_g_pp_support = aDHessVec;
 
+    bool hess_vec_prod_f_xx_support = aDHessVec;
```
mperego, line 620: Should not this be false, since we are not yet providing the AD hessian for f?

mperego (commented):
I had a couple of questions. Other than that looks good. Thanks for working on this!

mperego:
@bartgol @gahansen what do you think about Kim's question of whether we should have multiple flies for unit-testing evaluators instead of one? I'd think it will be better but I'm not sure how much code will have to be  repeated in the different files. Thanks.

src/LandIce/evaluators/LandIce_ResponseSurfaceVelocityMismatch_Def.hpp
```diff
@@ -193,6 +193,12 @@ void LandIce::ResponseSurfaceVelocityMismatch<EvalT, Traits>::evaluateFields(typ
         {
           ScalarT diff2 = std::pow(velocity(cell, side, qp, 0)  - observedVelocity (cell, side, qp, 0),2) 
                                + std::pow(velocity(cell, side, qp, 1)  - observedVelocity (cell, side, qp, 1),2);
+
+          // If diff2 is zero, we have to add a small number to it, otherwise the derivative computations will generate NaNs.
+          if(diff2 == Teuchos::ScalarTraits<ScalarT>::zero())
+            //Note: use a = instead of a += would erase all derivative information from diff2 which is unwanted.
```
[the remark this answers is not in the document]
kliegeois: I agree with that.
We can remove the if statement and just add the eps as you suggested.

src/Albany_ModelEvaluator.cpp
kliegeois: This function call sets the AD support to true for all the Hessian computations including the Hessian-vector products of the response functions and the Hessian-vector products of the residual.

Previously, as the AD implementation of the Hessian computations is not finished for all cases (as you know, the Hessian-vector products of the residual are not supported yet for instance), I removed this function call and set the support individually.

To answer your question, for the Hessian-vector products of g, they were computed with AD by default.

However, this might have been the case for the Hessian-vector products of f too.

src/Albany_ModelEvaluator.cpp
kliegeois: You are right, this is not correct yet.
This is an error which comes from the fact that the commits of this PR branch came from a branch were I am developing the AD Hessian-vector products for f.

Sorry for this mistake.

kliegeois:
Thanks a lot @mperego for your review!
I hope that my answers are clear.
I will update the code as suggested.

bartgol:
> @bartgol @gahansen what do you think about Kim's question of whether we should have multiple flies for unit-testing evaluators instead of one? I'd think it will be better but I'm not sure how much code will have to be repeated in the different files. Thanks.

Oh, absolutely. I always thought that was gonna be the case. To be clear, it is ok to jam a few "similar" evaluators in the same cpp file. For instance, I would be fine with
```
ls tests/units/evaluators
  interpolation_evaluators_ut.cpp
  basis_fcn_evaluators_ut.cpp
  bc_evaluators_ut.cpp
  response_evaluators_ut.cpp
  ...
ls tests/unit/disc
  block_disc_ut.cpp
  dof_mgr_ut.cpp
  stk_disc_ut.cpp
  ...
tests/unit/utility
  thyra_utils_ut.cpp
  cas_mgr_ut.cpp
  ..
...
```
I put in a ton of other things that theoretically we should unit test, but we don't, and nobody here has probably time to sit down and write them. Anyhow, all I wanted to say is that yes, we should compartmentalize tests as much as possible. Small files are easier to maintain, and it is also easier to locate a unit test if need be.

gahansen:
I agree also. "Teuchos unit test" is fine with several tests in the same file, one might want to do that inside of Luca's "stk_disc_ut.cpp"; this might contain an overall function that opens an Exodus file for the tests, but then have separate tests in there for the element blocks, mesh graph connectivity, fields, nodesets, sidesets that all lead up to testing the overall connection manager. Sometimes, especially when one is new to a code, it is good to see similar things in the same file so you do not need to hunt around for where they might be and what they are called. But that is why "grep" is handy, so I see it both ways.

mperego:
Thanks @bartgol and @gahansen for your feedback. 
@kliegeois Would you mind creating two separate files for unit testing `gatherSolution` and `gatherDistributedParameters` ? Don't worry about code duplication.  Sorry for the extra work.

kliegeois:
Thanks @mperego, @gahansen, and @bartgol for your feedback.
@mperego, I just committed a commit which splits the .cpp file.
Does it look fine to you?

mperego (approved):
It looks good! I'll merge it today if nobody objects.

Issue #638, opened 2020-10-16
Issue with Aggregate Responses


kliegeois:
There is an issue with Aggregate Responses. If I understand the `AggregateScalarResponseFunction` correctly, its purpose is to have several responses without coupling them by summing them together for instance.

In Thyra, there are capabilities to support multiple response functions.
For instance, it is possible to access the g space of the j-th response: `get_g_space(j)`.

Finally, in `Piro_PerformAnalysis`, there is an option to specify the response index which should be used for the optimization problem using `Response Vector Index` in the .yaml.

Based on this context, I assumed that the Aggregate Responses are intended to work as follows:
- Aggregate `n` responses together,
- Allow the possibility to perform the analysis on the j-th response alone.

However, the Aggregate Responses do not work like that currently.

This is due to lines 98-99 of src/responses/Albany_ResponseFactory.cpp:
```
    if(name == "Aggregate Responses")
      responses.push_back(rcp(new Albany::AggregateScalarResponseFunction(comm, scalar_responses)));
```
which `push_back` all the Aggregate Responses in `responses[0]`.

This implies that:
```
app->getNumResponses()
```
returns 1.

As a consequence, using a .yaml as the one of FO_GIS_betaT
```
    Response Functions:
      Number: 2
      Response 0: Surface Velocity Mismatch
      ResponseParams 0: 
        Regularization Coefficient: 1.00000000000000000e+00
        Asinh Scaling: 1.00000000000000000e+01
        Scaling Coefficient: 1.0
        Cubature Degree: 3
        Field Name: Velocity
      Response 1: Surface Mass Balance Mismatch
      ResponseParams 1: 
        Regularization Coefficient: 1.00000000000000000e+00
        H Coefficient: 1.00000000000000000e+00
        SMB Coefficient: 1.00000000000000000e+02
        Scaling Coefficient: 1.0
        Cubature Degree: 3
        Field Name: Velocity

```
which results in Aggregate Responses (default value if `Collection Method` is not specified) and we if use:
```
        Response Vector Index: 1
```
this results in the error:
```
ROL::ThyraProductME_Objective_SimOpt::value

p=0: *** Caught standard std::exception of type 'Teuchos::Exceptions::InvalidParameter' :

 /home/kim/dev/Albany_all/Albany/src/Albany_ModelEvaluator.cpp:431:
 
 Throw number = 1
 
 Throw test that evaluated to true: l >= app->getNumResponses() || l < 0
 
 
[...]

Issue #119, opened 2017-04-27
Add CDash capability to LCM scripts


ibaned:
For some reason there was not already an issue tracking this...
I'm trying to do this by rewriting the `build.sh` script in CMake.

ibaned:
This seems to be working now.

Issue #14, opened 2016-08-18
Element technology


jtostie:
Higher order surface elements and different integration strategies (48 point) for the composite tet.

lxmota:
I had a conversation with John Emery. He has funds for this. @jwfoulk, @jtostie and I need to talk about this.

mperego:
Is this something that will end up in Intrepid2?

lxmota:
Yes, this is something that would end up being implemented in Intrepid2. I still need to talk to John Emery to see if he is OK with having this in Albany.

PR #640, opened 2020-10-22
WIP: Drop support for old format for responses and parameters

kliegeois:
Albany currently supports two formats for both responses and parameters: the old and the new format.

This PR is a WIP intended to drop support for the old format and to solve the issue #638.

Currently, only 3 tests have been updated:
- SteadyHeatConstrainedOpt2D_Dirichlet_Mixed_Params_Tpetra,
- FO_GIS_AdjointSensitivity_Tpetra,
- SteadyHeat2D_Tpetra.

Those 3 tests passed.

This PR should not be merged yet as not all the tests have been updated.
They will be updated when the update of the format is finished.

The goal of opening this PR as WIP is to enable discussion on the required modifications.

Concerning the format, it seems that for the tested tests of my build of the master branch, `Number of Sensitivity Comparisons` is always 1 or 0. Is it supposed to be the equivalent of `Number of Comparisons` but for sensitivity results? If yes, as the results are always scalar, should we remove `Number of Sensitivity Comparisons` as we have removed `Number of Comparisons`?


mperego:
@kliegeois Thanks for working on this. In the input files of the tests you considered, you are still using the old format for defining the parameters and the responses. You only changed the regression part. I'm wondering how is it possible that the test pass. 

As for the question about the sensitivities, I'm not sure yet, but we should find a way to ensure that the comparisons are actually performed as intended. It happened in the past that a test passed simply because no comparison was performed. We should find a way to avoid this.

kliegeois:
> @kliegeois Thanks for working on this. In the input files of the tests you considered, you are still using the old format for defining the parameters and the responses. You only changed the regression part. I'm wondering how is it possible that the test pass. 
> 
> 
> 
> As for the question about the sensitivities, I'm not sure yet, but we should find a way to ensure that the comparisons are actually performed as intended. It happened in the past that a test passed simply because no comparison was performed. We should find a way to avoid this.

@mperego Thanks for your message. That's weird, it passed on my computer and Blake. I will investigate that.

mperego:
I guess that if we always assume the `Number Of Sensitivity Comparisons` to be equal to 1, we avoid the possibility of not having comparisons performed. Yeah, I would remove  `Number Of Sensitivity Comparisons`.

mperego:
> 
> 
> > @kliegeois Thanks for working on this. In the input files of the tests you considered, you are still using the old format for defining the parameters and the responses. You only changed the regression part. I'm wondering how is it possible that the test pass.
> > As for the question about the sensitivities, I'm not sure yet, but we should find a way to ensure that the comparisons are actually performed as intended. It happened in the past that a test passed simply because no comparison was performed. We should find a way to avoid this.
> 
> @mperego Thanks for your message. That's weird, it passed on my computer and Blake. I will investigate that.

No problem. We should anyway modify the parameter and responses sections of the input files to use the new format. In some of the tests the parameters are already in the new format, but not in all.

kliegeois:
The tests definitely pass:

```
Test project /home/kim/dev/Albany_all/AlbanyB
    Start 119: FO_GIS_AdjointSensitivity_Tpetra
1/1 Test #119: FO_GIS_AdjointSensitivity_Tpetra ...   Passed   72.73 sec

100% tests passed, 0 tests failed out of 1

Label Time Summary:
Forward    =  72.73 sec*proc (1 test)
LandIce    =  72.73 sec*proc (1 test)
Tpetra     =  72.73 sec*proc (1 test)

Total Test time (real) =  72.74 sec
Test project /home/kim/dev/Albany_all/AlbanyB
    Start 14: SteadyHeatConstrainedOpt2D_Dirichlet_Mixed_Params_Tpetra
1/1 Test #14: SteadyHeatConstrainedOpt2D_Dirichlet_Mixed_Params_Tpetra ...   Passed    4.15 sec

100% tests passed, 0 tests failed out of 1

Label Time Summary:
Basic      =   4.15 sec*proc (1 test)
Forward    =   4.15 sec*proc (1 test)
Tpetra     =   4.15 sec*proc (1 test)

Total Test time (real) =   4.16 sec
Test project /home/kim/dev/Albany_all/AlbanyB
    Start 5: SteadyHeat2D_Tpetra
1/1 Test #5: SteadyHeat2D_Tpetra ..............   Passed    3.95 sec

100% tests passed, 0 tests failed out of 1

Label Time Summary:
Basic      =   3.95 sec*proc (1 test)
Forward    =   3.95 sec*proc (1 test)
Tpetra     =   3.95 sec*proc (1 test)

Total Test time (real) =   3.96 sec
```

I am a bit confused; I thought that the modified .yaml files were already following the new format for the responses and the parameters but not for the regression results.
This thought implied that the tests pass.
However, this is inconsistent with your comment which says that both responses and parameters of those files are written in the old format.

Do we have an example of a .yaml file which only use the new format or do we have documentation for the new format?
Alternatively, could you provide an update of `tests/small/LandIce/FO_GIS/input_fo_gis_adjoint_sensitivityT.yaml` which is consistent with the new format?

Thanks!

kliegeois:
@mperego, I think that I found the issue.
I will add a commit today.

kliegeois:
I think that I resolved the issue for `SteadyHeat2D_Tpetra ` (`tests/small/SteadyHeat2D/inputT.yaml`):

```
Test project /home/kim/dev/Albany_all/AlbanyB
    Start 119: FO_GIS_AdjointSensitivity_Tpetra
1/1 Test #119: FO_GIS_AdjointSensitivity_Tpetra ...   Passed   73.40 sec

100% tests passed, 0 tests failed out of 1

Label Time Summary:
Forward    =  73.40 sec*proc (1 test)
LandIce    =  73.40 sec*proc (1 test)
Tpetra     =  73.40 sec*proc (1 test)

Total Test time (real) =  73.41 sec
Test project /home/kim/dev/Albany_all/AlbanyB
    Start 14: SteadyHeatConstrainedOpt2D_Dirichlet_Mixed_Params_Tpetra
1/1 Test #14: SteadyHeatConstrainedOpt2D_Dirichlet_Mixed_Params_Tpetra ...   Passed    4.12 sec

100% tests passed, 0 tests failed out of 1

Label Time Summary:
Basic      =   4.12 sec*proc (1 test)
Forward    =   4.12 sec*proc (1 test)
Tpetra     =   4.12 sec*proc (1 test)

Total Test time (real) =   4.13 sec
Test project /home/kim/dev/Albany_all/AlbanyB
    Start 5: SteadyHeat2D_Tpetra
1/1 Test #5: SteadyHeat2D_Tpetra ..............   Passed    5.86 sec

100% tests passed, 0 tests failed out of 1

Label Time Summary:
Basic      =   5.86 sec*proc (1 test)
Forward    =   5.86 sec*proc (1 test)
Tpetra     =   5.86 sec*proc (1 test)

Total Test time (real) =   5.87 sec
```

Could you confirm me that `tests/small/SteadyHeat2D/inputT.yaml` uses the new format for the parameters and the solutions?

The work is not finished yet, however.

I have two questions:
- Parameters can only be scalar, right? Therefore, would it make sense to remove the `Number:` from `Parameter Vector 0:` ? As this is always supposed to be 1.
- Do we want to allow to have two nested vectors of solutions? In other words, do we want to have cases where:
```
    Response Functions: 
      Number of Response Vectors: 2
      Response Vector 0:
        Number: 1
        Name: Solution Average
        Response 0: Solution Average
      Response Vector 1:
        Number: 1
        Name: Solution Two Norm
        Response 1: Solution Two Norm
```
one of the `Number:` is not 1?

This could be useful when we want to have two solutions where one of the two is the sum of two others. However, removing this possibility could help to ease the format.

mperego:
> 
> 
> I think that I resolved the issue for `SteadyHeat2D_Tpetra ` (`tests/small/SteadyHeat2D/inputT.yaml`):
> 
> ```
> Test project /home/kim/dev/Albany_all/AlbanyB
>     Start 119: FO_GIS_AdjointSensitivity_Tpetra
> 1/1 Test #119: FO_GIS_AdjointSensitivity_Tpetra ...   Passed   73.40 sec
> 
> 100% tests passed, 0 tests failed out of 1
> 
> Label Time Summary:
> Forward    =  73.40 sec*proc (1 test)
> LandIce    =  73.40 sec*proc (1 test)
> Tpetra     =  73.40 sec*proc (1 test)
> 
> Total Test time (real) =  73.41 sec
> Test project /home/kim/dev/Albany_all/AlbanyB
>     Start 14: SteadyHeatConstrainedOpt2D_Dirichlet_Mixed_Params_Tpetra
> 1/1 Test #14: SteadyHeatConstrainedOpt2D_Dirichlet_Mixed_Params_Tpetra ...   Passed    4.12 sec
> 
> 100% tests passed, 0 tests failed out of 1
> 
> Label Time Summary:
> Basic      =   4.12 sec*proc (1 test)
> Forward    =   4.12 sec*proc (1 test)
> Tpetra     =   4.12 sec*proc (1 test)
> 
> Total Test time (real) =   4.13 sec
> Test project /home/kim/dev/Albany_all/AlbanyB
>     Start 5: SteadyHeat2D_Tpetra
> 1/1 Test #5: SteadyHeat2D_Tpetra ..............   Passed    5.86 sec
> 
> 100% tests passed, 0 tests failed out of 1
> 
> Label Time Summary:
> Basic      =   5.86 sec*proc (1 test)
> Forward    =   5.86 sec*proc (1 test)
> Tpetra     =   5.86 sec*proc (1 test)
> 
> Total Test time (real) =   5.87 sec
> ```
> 
> Could you confirm me that `tests/small/SteadyHeat2D/inputT.yaml` uses the new format for the parameters and the solutions?
> 
> The work is not finished yet, however.
> 
> I have two questions:
> 
>     * Parameters can only be scalar, right? Therefore, would it make sense to remove the `Number:` from `Parameter Vector 0:` ? As this is always supposed to be 1.
> 
>     * Do we want to allow to have two nested vectors of solutions? In other words, do we want to have cases where:
> 
> 
> ```
>     Response Functions: 
>       Number of Response Vectors: 2
>       Response Vector 0:
>         Number: 1
>         Name: Solution Average
>         Response 0: Solution Average
>       Response Vector 1:
>         Number: 1
>         Name: Solution Two Norm
>         Response 1: Solution Two Norm
> ```
> 
> one of the `Number:` is not 1?
> 
> This could be useful when we want to have two solutions where one of the two is the sum of two others. However, removing this possibility could help to ease the format.

Yeah, I think it is correct, sorry if it was not clear from my comment yesterday.  I think the parameters can be formed by vector of scalar parameters. As for the response it would be nice if we could have one of the responses to be the sum of other responses. I'll try to come up with a proposal for the parameter and responses briefly

mperego:
b.t.w. if you say that a test pass I trust you, no need to paste the output ;-)

mperego:
@ikalash @jewatkins @bartgol @kliegeois 
I'm trying to come up with a new format for defining the parameters/responses in the input file. How do you like this?

```
   Parameters:   #Unify Scalar and Distributed Parameters
      Number of Parameters: 3
      Parameter 0:
        Type: Vector Of Scalar Parameters  #either "Vector Of Scalar Parameters" or "Distributed Parameter"
        Number: 3
        Scalar Parameter 0: 
          Name: DBC on NS NodeSet0 for DOF T
        Scalar Parameter 1
          Name: DBC on NS NodeSet1 for DOF T
          Lower Bounds: [-2.0e+00]  #optional properties of the parameter
          Upper Bounds: [2.0+00]
        Scalar Parameter 2: 
          Name: DBC on NS NodeSet2 for DOF T
      Parameter 1:
        Type: Vector Of Scalar Parameters
        Number: 1
        Scalar Parameter 0: 
          Name: Quadratic Nonlinear Factor
          Lower Bounds: [-2.0e+00]
          Upper Bounds: [2.0+00]
      Parameter 2:
        Type: Distributed Parameter
        Name: dirichlet_field
        Lower Bound: -1.00000000000000000e+00   #Optional property of the parameter
        Upper Bound: 2.00000000000000000e+00
        Initial Uniform Value: -5.00000000000000000e-01

    Response Functions: 
      Number of Responses: 2  #we don't have any longer a collection method
      Response 0:
        Type: Scalar Response     #either "Scalar Response" (default) or "Sum Of Responses"
        Name: Solution Average
      Response 1:
        Type: Sum Of Responses   
        Number of Responses: 2   #option available only for Type: Sum Of Responses
        Response 0: 
          Name: SMB Mismatch
        Response 1: 
          Name: Solution Two Norm

  #Regressions for the Responses. It's OK not to have a regression section for any response (in this case we do not do regression on response 0.
    Regression For Response 1: 
      Test Value: 1.4
      Relative Tolerance: 1.0e-3
      Sensitivity For Parameter 0: 
        Test Values: [4.5e-01, 0.0, 2.1]  #array<double>, size depends on the number of scalar parameters
      Sensitivity For Parameter 1:
        Test Values: [3.0]          #array<double>, size depends on the number of scalar parameters
      Sensitivity For Parameter 2:
        Test Value: -9.21e-2        #double for Distributed parameters
```

ikalash:
I have a few questions:

- It seems if there is a scalar parameter, you still need to declare it as a 'Vector of Scalar Parameters' of length 1?  It seems a little bit verbose, but I suppose it's OK.  Maybe you can make 'Vector of Scalar Parameters' the default?
- Does 'Sum of Responses' sum the responses - so in your example, it would sum 'SMB Mismatch' and 'Solution Two Norm'?  Do we have that option currently?  How useful is this option in practice?  I don't really see a case where one would want to check the sum of two responses against a baseline in the tests, but maybe I am missing something.
- Is the purpose of adding 'Regression for Response N' to allow the user to not do a regression test for all the responses?  So in your example, you are not checking the value of Response 0.  Currently, I think we have to do regression tests either for all or no responses.  The regression section again becomes more verbose, but it does provide more flexibility.

mperego:
Thanks @ikalash for your feedback 
> 
> I have a few questions:
> 
>     * It seems if there is a scalar parameter, you still need to declare it as a 'Vector of Scalar Parameters' of length 1?  It seems a little bit verbose, but I suppose it's OK.  Maybe you can make 'Vector of Scalar Parameters' the default?

That's correct. It's mainly to avoid having too many Types to support. But we can add a `Scalar Parameter` type. In practice the most used type is `Distributed Parameter`
> 
>     * Does 'Sum of Responses' sum the responses - so in your example, it would sum 'SMB Mismatch' and 'Solution Two Norm'?  Do we have that option currently?  How useful is this option in practice?  I don't really see a case where one would want to check the sum of two responses against a baseline in the tests, but maybe I am missing something.

We already have this capability. We use it for optimization, e.g. to add the mismatch response to the regularization response. Otherwise you would have to create a lot of responses. 
> 
>     * Is the purpose of adding 'Regression for Response N' to allow the user to not do a regression test for all the responses?  So in your example, you are not checking the value of Response 0.  Currently, I think we have to do regression tests either for all or no responses.  The regression section again becomes more verbose, but it does provide more flexibility.

This is inline with the "new" way to declare responses. I agree it makes it more verbose, but also clearer.

ikalash:
@mperego : I see.  I didn't realize we had already the sum of responses option.  I doubt it will be used much other than for the optimization, but this is OK  - it is not the default.  Maybe adding ``Scalar Parameter'' would be good.  Other than that, I'm good with the new syntax.  Thanks for thinking about this and working on cleaning things up!

jewatkins:
> I'm trying to come up with a new format for defining the parameters/responses in the input file. How do you like this?

Looks good to me.

bartgol:
> @mperego : I see. I didn't realize we had already the sum of responses option. I doubt it will be used much other than for the optimization, but this is OK - it is not the default. Maybe adding ``Scalar Parameter'' would be good. Other than that, I'm good with the new syntax. Thanks for thinking about this and working on cleaning things up!

Another possible use of this is to compute, say, integrals/mismatches over multiple sidesets, without having to define a "superset" sideset in the mesh (simply add the two integrals/mismatches responses over the individual sidesets).
Not really essential, but it allows one to simply edit an input file, rather than manually hack the mesh.

bartgol:
@mperego I mostly like it. A couple of proposals:

- string key for the bounds of scalar params should probably be 'Lower Bound' and 'Upper Bound' (singular, not plural).
- what about simply 'Vector' instead of 'Vector of Scalar Parameters' and 'Field' instead of 'Distributed Parameter'? It's kind of repetitive to use 'Parameter(s)', since we are in the parameters sublist already.
- for type, what about 'Scalar', 'Vector', and 'Field'? The last two are what you already proposed, while the first is a special case for a size-1 vector, which might be a bit simpler for users. It might add a bit of code when parsing the parameters lists though.

And a question: do we think it will ever be needed to handle vector distributed parameters?

mperego:
@bartgol OK, below is an updated portion of the input file. I like "Field", but i the code we use "Distributed" instead, so I would think it is better to do "Scalar, Vector, Distributed". The introduction of scalar should also address @ikalash comment. Note that I also changed "Number" into "Dimension", for vector of parameters.
As for having vector of distributed parameters, we might, but I don't see a real need for that. For optimization we can already compose distributed parameter into a Thyra vector.

```
   Parameters:   #Unify Scalar and Distributed Parameters
      Number of Parameters: 3
      Parameter 0:
        Type: Vector  #"Scalar" or "Vector" or "Distributed"
        Dimension: 3
        Scalar 0: 
          Name: DBC on NS NodeSet0 for DOF T
        Scalar 1
          Name: DBC on NS NodeSet1 for DOF T
          Lower Bound: [-2.0e+00]  #optional properties of the parameter
          Upper Bound: [2.0+00]
        Scalar 2: 
          Name: DBC on NS NodeSet2 for DOF T
      Parameter 1:
        Type: Scalar 
        Name: Quadratic Nonlinear Factor
          Lower Bound: [-2.0e+00]
          Upper Bound: [2.0+00]
      Parameter 2:
        Type: Distributed
        Name: dirichlet_field
        Lower Bound: -1.0e+00   #Optional property of the parameter
        Upper Bound: 2.0e+00
        Initial Uniform Value: -5.0e-01

    Response Functions: 
      Number of Responses: 2  #we don't have any longer a collection method
      Response 0:
        Type: Scalar Response     #either "Scalar Response" (default) or "Sum Of Responses"
        Name: Solution Average
      Response 1:
        Type: Sum Of Responses   
        Number of Responses: 2   #option available only for Type: Sum Of Responses
        Response 0: 
          Name: SMB Mismatch
        Response 1: 
          Name: Solution Two Norm

  #Regressions for the Responses. It's OK not to have a regression section for any response (in this case we do not do regression on response 0.
    Regression For Response 1: 
      Test Value: 1.4
      Relative Tolerance: 1.0e-3
      Sensitivity For Parameter 0: 
        Test Values: [4.5e-01, 0.0, 2.1]  #array<double>, size depends on the number of scalar parameters
      Sensitivity For Parameter 1:
        Test Value: 3.0          #double for Scalar parameters
      Sensitivity For Parameter 2:
        Test Value: -9.21e-2        #double for Distributed parameters
```

bartgol:
I asked about vector fields parameters, cause I wondered whether we should hint anywhere at the fact that `Distributed` is always scalar or not. But I am fine as it is.

Another note: lower/upper bounds for scalar parameters need not to be arrays (in yaml syntax). So something like `Upper Bound: 2.0`, rather than `Upper Bound: [2.0]`.

mperego:
I see. Should we have "Scalar Distributed"? It's a bit awkward.

I edited the yaml portion and fixed the format of the lower/upper bounds. thanks.

bartgol:
Yeah, it is a bit awkward. I think for now we can just leave it as it is. If we really want to be good, when doing our fields syncing with the mesh, we could verify that the field is indeed scalar. But I don't see it as a priority.

kliegeois:
I updated the 3 previously discussed tests:
- SteadyHeatConstrainedOpt2D_Dirichlet_Mixed_Params_Tpetra,
- FO_GIS_AdjointSensitivity_Tpetra,
- SteadyHeat2D_Tpetra,
to be consistent with the new format proposed by @mperego.

I updated the .cpp files to read the different parts of the .yaml correctly. 
The 3 tests passed correctly.

As, now, both non-distributed and distributed parameters are listed inside the same `parameterlist`, I assumed that the non-distributed parameters are always listed before the distributed ones. If it is not the case, i.e. if at least one non-distributed parameter is listed after at least one distributed parameter, an error is thrown which asks to reorder the parameters.

This assumption has been done to simplify the computation of the number of non-distributed parameters and the number of distributed parameters.

This assumption is not set in stone and could be removed if you want.

The .cpp files are not yet fully cleaned; I wanted to first update the PR with the 3 working tests.

The remaining work includes:
- Convert all the other tests and verify if they pass,
- Clean the .cpp files.

src/Albany_RegressionTests.cpp
```diff
@@ -41,77 +41,71 @@ checkSolveTestResults(
...
-      failures += scaledCompare(g_view[i], testValues[i], relTol, absTol, s);
-      comparisons++;
-    }
+  // Test the response (g)
+  const int numResponseTests = 1;
+  ALBANY_ASSERT(
+      g != Teuchos::null,
+      "There is Response Test but the response vector is null!");
+  ALBANY_ASSERT(
+      numResponseTests <= g->space()->dim(),
```
mperego, line 50: I would assert that `g->space->dim()==1` explaining in the error message that responses are assumed to be scalar, and maybe remove `numResponseTests`.

src/Albany_RegressionTests.cpp
```diff
@@ -41,77 +41,71 @@ checkSolveTestResults(
...
-            << response_index << ", " << parameter_index << ") is null!");
-    if (numSensTests != dgdp->range()->dim()) {
-      ALBANY_ASSERT(
-          false,
-          "Number of sensitivity tests ("
-              << numSensTests << ") != number of sensitivities ["
-              << response_index << "][" << parameter_index << "] ("
-              << dgdp->range()->dim() << ") !");
-    }
+  else {
```
mperego, line 74: Not sure what `storeTestResults` does, but I think we should not error out if we do not have the sensitivity associated to a parameter. It's fine, we don't  want to require that we check all sensitivities.

src/Albany_RegressionTests.cpp
```diff
@@ -41,77 +41,71 @@ checkSolveTestResults(
...
-    Teuchos::Array<double> testSensValues =
-        sensitivityParams->get<Teuchos::Array<double>>(
-            strint("Sensitivity Test Values", i));
-    ALBANY_ASSERT(
-        numVecs== testSensValues.size(),
-        "Number of Sensitivity Test Values ("
-            << testSensValues.size() << " != number of sensitivity vectors ("
-            << numVecs << ") !");
+    Teuchos::Array<double> testSensValues;
+    if (sensitivityParams->isType<Teuchos::Array<double>>("Test Values"))
```
mperego, line 87: I think it's better to rely on whether the parameter is of Vector type or not. If it is of vector type then we look for 'Test Values' of Array type and make sure that the dimension of the `Test Values` is the same of the parameter one. Similarly for the Scalar/Distributed case.

src/Albany_RegressionTests.cpp
```diff
@@ -172,17 +166,15 @@ RegressionTests::getTestParameters(int response_index) const
...
-        strint("Regression Results", response_index)));
-  }
+
+  result = &(appParams->sublist(
+      strint("Regression For Response", response_index)));
+
 
   TEUCHOS_TEST_FOR_EXCEPTION(
-      result->isType<std::string>("Test Values"),
+      result->isType<std::string>("Test Value"),
```
mperego, line 175: Probably this is not needed, since we would get an error anyway if the Type is wrong.

src/Albany_RegressionTests.cpp
```diff
@@ -240,25 +232,24 @@ RegressionTests::getValidRegressionResultsParameters() const
 
   validPL->set<int>(
       "Number of Comparisons", 0, "Number of responses to regress against");
```
mperego, line 234: I don't think we have 'Number Of Comparisons` anymore

src/Albany_Utils.cpp
```diff
@@ -72,22 +72,12 @@ CalculateNumberParams(const Teuchos::RCP<Teuchos::ParameterList>& problemParams)
...
+  int  num_param_vecs = parameterParams.get("Number of Parameters", 0);
   int np = 0;
   for (int i = 0; i < num_param_vecs; ++i) {
     Teuchos::ParameterList& pList =
-        using_old_parameter_list ?
-            parameterParams :
-            parameterParams.sublist(Albany::strint("Parameter Vector", i));
-    np += pList.get<int>("Number");
+        parameterParams.sublist(Albany::strint("Parameter", i));
+    np += pList.get<int>("Dimension",1);
```
mperego, line 80: I would not provide a default Dimension, it's safer.

src/LandIce/problems/LandIce_SchoofFit.hpp
```diff
@@ -162,21 +162,33 @@ LandIce::SchoofFit::constructEvaluators (PHX::FieldManager<PHAL::AlbanyTraits>&
     save_state[states_to_save[i]] = true;
   }
 
-  Teuchos::ParameterList& dist_params_list =  this->params->sublist("Distributed Parameters");
-  Teuchos::ParameterList* param_list;
-  int numParams = dist_params_list.get<int>("Number of Parameter Vectors",0);
-  for (int p_index=0; p_index< numParams; ++p_index)
+  Teuchos::ParameterList& parameterParams = this->params->sublist("Parameters");
+  int total_num_param_vecs = parameterParams.get("Number of Parameters", 0);
```
mperego, line 166: Same here. Let's not provide defaults, unless we really mean to.

src/Main_Solve.cpp
```diff
@@ -177,79 +177,85 @@ int main(int argc, char *argv[])
...
-    int num_param_vecs = parameterParams.get("Number of Parameter Vectors", 0);
-    bool using_old_parameter_list = false;
-    if (parameterParams.isType<int>("Number")) {
-      int numParameters = parameterParams.get<int>("Number");
-      if (numParameters > 0) {
-        num_param_vecs = 1;
-        using_old_parameter_list = true;
-      }
+    int num_param_vecs = 0;
+    int total_num_param_vecs = parameterParams.get("Number of Parameters", 0);
```
mperego, line 181: Same here, no defaults.

src/Main_Solve.cpp
```diff
@@ -177,79 +177,85 @@ int main(int argc, char *argv[])
...
+
+      std::string parameterType = pList->get<std::string>("Type", "Scalar");
+      if(parameterType == "Scalar") {
+        param_names[l] =
+            Teuchos::rcp(new Teuchos::Array<std::string>(1));
+        (*param_names[l])[0] =
+            pList->get<std::string>("Name");
+      }
+      if(parameterType == "Vector") {
+        const int numParameters = pList->get<int>("Dimension", 1);
```
mperego, line 213: Same here, no defaults.

src/problems/Albany_HeatProblem.cpp
```diff
@@ -42,25 +42,18 @@ HeatProblem( const Teuchos::RCP<Teuchos::ParameterList>& params_,
...
-      if(p.get<std::string>("Name","")== "thermal_conductivity")
-        conductivityIsDistParam = true;
-        break;
-    } 
-  }
   dirichletIsDistParam = false;
-  if(params->isSublist("Distributed Parameters")) {
-    int numDistParams = params->sublist("Distributed Parameters").get<int>("Number of Parameter Vectors",0);
+  if(params->isSublist("Parameters")) {
+    int numDistParams = params->sublist("Parameters").get<int>("Number of Parameters",0);
```
mperego, line 47: Same here, no defaults.

src/problems/Albany_HeatProblem.cpp
```diff
@@ -42,25 +42,18 @@ HeatProblem( const Teuchos::RCP<Teuchos::ParameterList>& params_,
...
     for (int i=0; i<numDistParams; ++i) {
-      Teuchos::ParameterList p = params->sublist("Distributed Parameters").sublist(Albany::strint("Distributed Parameter", i));
-      if(p.get<std::string>("Name","")== "dirichlet_field")
+      Teuchos::ParameterList p = params->sublist("Parameters").sublist(Albany::strint("Parameter", i));
+      if(p.get<std::string>("Name","") == "thermal_conductivity" && p.get<std::string>("Type","Scalar") == "Distributed")
+        conductivityIsDistParam = true;
+      if(p.get<std::string>("Name","") == "dirichlet_field" && p.get<std::string>("Type","Scalar") == "Distributed") {
```
mperego, line 52: same here

mperego (commented):
Wow, that's a lot of work! Even more than I expected. Thanks for doing that.
Looks good to me, I only added some minor comments.

src/Albany_RegressionTests.cpp
kliegeois: This does not error out if we do not have the sensitivity associated to a parameter.
This, instead, does not add sensitivity checks if the parameter list does not include the sublist associated with the current parameter.

I think that the `return failures;` is misleading in this case.
The `failures` variable is an integer which corresponds to the number of failed comparisons. This number is zero when all the tests passed.

kliegeois:
> Wow, that's a lot of work! Even more than I expected. Thanks for doing that.
> Looks good to me, I only added some minor comments.

Thanks a lot @mperego for your in deep review!
I just finished to update the PR taking into account your comments.
I will now continue to update the remaining tests.

mperego:
> 
> 
> > Wow, that's a lot of work! Even more than I expected. Thanks for doing that.
> > Looks good to me, I only added some minor comments.
> 
> Thanks a lot @mperego for your in deep review!
> I just finished to update the PR taking into account your comments.
> I will now continue to update the remaining tests.

Thanks! I looks good to me!

kliegeois:
I have a new question related to the: `Solution Values` response function.

This response function takes as arguments an array of Node GIDs and can be used to test the solution values for those IDs.
This response function can return an array of solutions (if more than one GID is provided).
With the new format, we are assuming that the responses are always scalar. Therefore, I updated [a test](https://github.com/SNLComputation/Albany/blob/master/tests/small/SteadyHeat3D/inputT_nodeGIDArrayResponse.yaml) which was using:
```
    Response Functions: 
      Number: 1
      Response 0: Solution Values
      ResponseParams 0: 
        Culling Strategy: Node GIDs
        Node GID Array: [58, 193, 194]
(...)
  Regression Results: 
    Number of Comparisons: 3
    Test Values: [1.50000000000000000e+00, 1.54466999999999999e+00, 1.55048000000000008e+00]
    Relative Tolerance: 1.00000000000000002e-03
    Number of Sensitivity Comparisons: 0
```

as follows:
```
    Response Functions: 
      Number of Responses: 3
      Response 0:
        Name: Solution Values
        Culling Strategy: Node GIDs
        Node GID Array: [58]
      Response 1:
        Name: Solution Values
        Culling Strategy: Node GIDs
        Node GID Array: [193]
      Response 2:
        Name: Solution Values
        Culling Strategy: Node GIDs
        Node GID Array: [194]
(...)
  Regression For Response 0:
    Test Value: 1.50000000000000000e+00
    Relative Tolerance: 1.00000000000000002e-03
  Regression For Response 1:
    Test Value: 1.54466999999999999e+00
    Relative Tolerance: 1.00000000000000002e-03
  Regression For Response 2:
    Test Value: 1.55048000000000008e+00
    Relative Tolerance: 1.00000000000000002e-03
```

Should we support vector response function?
If what I have done is fine, I think that:
- we should replace `Node GID Array` by `Node GID` which should be an integer instead of an array of integers,
- we should replace `Solution Values` by `Solution Value`.

What do you prefer?

mperego:
> 
> 
> I have a new question related to the: `Solution Values` response function.
> 
> This response function takes as arguments an array of Node GIDs and can be used to test the solution values for those IDs.
> This response function can return an array of solutions (if more than one GID is provided).
> With the new format, we are assuming that the responses are always scalar. Therefore, I updated [a test](https://github.com/SNLComputation/Albany/blob/master/tests/small/SteadyHeat3D/inputT_nodeGIDArrayResponse.yaml) which was using:
> 
> ```
>     Response Functions: 
>       Number: 1
>       Response 0: Solution Values
>       ResponseParams 0: 
>         Culling Strategy: Node GIDs
>         Node GID Array: [58, 193, 194]
> (...)
>   Regression Results: 
>     Number of Comparisons: 3
>     Test Values: [1.50000000000000000e+00, 1.54466999999999999e+00, 1.55048000000000008e+00]
>     Relative Tolerance: 1.00000000000000002e-03
>     Number of Sensitivity Comparisons: 0
> ```
> 
> as follows:
> 
> ```
>     Response Functions: 
>       Number of Responses: 3
>       Response 0:
>         Name: Solution Values
>         Culling Strategy: Node GIDs
>         Node GID Array: [58]
>       Response 1:
>         Name: Solution Values
>         Culling Strategy: Node GIDs
>         Node GID Array: [193]
>       Response 2:
>         Name: Solution Values
>         Culling Strategy: Node GIDs
>         Node GID Array: [194]
> (...)
>   Regression For Response 0:
>     Test Value: 1.50000000000000000e+00
>     Relative Tolerance: 1.00000000000000002e-03
>   Regression For Response 1:
>     Test Value: 1.54466999999999999e+00
>     Relative Tolerance: 1.00000000000000002e-03
>   Regression For Response 2:
>     Test Value: 1.55048000000000008e+00
>     Relative Tolerance: 1.00000000000000002e-03
> ```
> 
> Should we support vector response function?
> If what I have done is fine, I think that:
> 
>     * we should replace `Node GID Array` by `Node GID` which should be an integer instead of an array of integers,
> 
>     * we should replace `Solution Values` by `Solution Value`.
> 
> 
> What do you prefer?

If that's the only Response that is not scalar, I think what you have done is fine and we should not support a vector response. The changes you propose for `Node GID Array` and `Solution Values` make sense to me.

bartgol:
Is there going to be penalties in building N responses instead of 1? I am wondering whether having N import-export calls for 1 GID each is going to be remarkably worse than 1 call with N gids. Clearly, I'm more concerned for the case where N>>1, like maybe 1,000 or so.

I guess the question to ask first is: what's the typical usage of that response?

mperego:
@bartgol I never used that response.. it seems to me more a debug tool. But I might be wrong.

kliegeois:
@bartgol, I think that the only current usage is to verify few entries of the solution vector (typically 3). I have to verify that tomorrow.

@mperego, it is the only occurrence of a vector-typed solution function that I encountered for now. However, I still need to update about half of the small tests and I have not updated any of the large tests. They might be other occurrences in those tests.
Sorry for being slow updating the tests, I verify them one by one to be sure that they pass correctly.

bartgol:
If that's the case, then I'm 100% fine with your changes.

mperego:
@kliegeois Yeah, I know it takes time. Large tests should not bring any surprise in terms of Responses/Parameters, I think.

ikalash:
Do any tests use it?  I remember QCAD used to.

kliegeois:
@ikalash there is a total of 4 tests which use `Solution Values` : 
- tests/small/LandIce/FO_Test/input_fo_domeAnalysis.yaml
- tests/small/LandIce/FO_Test/input_fo_domeFoward.yaml 
- tests/small/SteadyHeat3D/input_nodeGIDArrayResponse.yaml 
- tests/small/SteadyHeat3D/inputT_nodeGIDArrayResponse.yaml

Each of those tests seem to use it to verify 3 values. Even if the first two tests have `Num Values: 100`, they are only testing 3 values.

I ended up creating a Python script which converts the tests from the old and 'new' format to the current format.
I can commit this script if some of you want, this might be useful to update the tests which are not in the repository.

kliegeois:
All the tests passed except:
```
The following tests FAILED:
        109 - FO_GIS_Unstructured_Tpetra (Failed)
        110 - FO_GIS_Unstructured_Memoization (Failed)
        111 - FO_GIS_Unstructured_Tpetra_FROSch (Failed)
Errors while running CTest
```
Those 3 tests failed due to the sensitivity check:
``` 
111:  Sensitivities (0,0):
111:  
111:                      1.882621796611e+07
111:  Sensitivity Test 0,0: 1.882621796611e+07 != 2.078020165630e+07 (rel 1.000000000000e-04 abs 1.000000000000e-04)
111:  
111:  CheckTestResults: Number of Comparisons Attempted = 2
```

For the 3 tests, I have those values:
- Test 109: expected: 2.078020165630e+07, computed: 1.882621076489e+07,
- Test 110: expected: 2.078020165630e+07, computed: 1.882621076489e+07,
- Test 111: expected: 2.078020165630e+07, computed: 1.882621796611e+07.

I will have to investigate why the computed sensitivity is not as expected.

kliegeois:
I just verified recompiling and testing the master branch.
The 3 tests (on the master branch) passed but did not check the computed sensitivity values with the expected ones.
The computed ones are not consistent with the expected ones and are the same as the values computed with the branch of this PR. Therefore, I updated the expected values to be consistent with the computed ones.

I am currently recompiling a last time this branch to double-check the tests.

After that, I can rebase to squash the commits and this PR can be merged.

mperego:
> 
> 
> I just verified recompiling and testing the master branch.
> The 3 tests (on the master branch) passed but did not check the computed sensitivity values with the expected ones.
> The computed ones are not consistent with the expected ones and are the same as the values computed with the branch of this PR. Therefore, I updated the expected values to be consistent with the computed ones.
> 
> I am currently recompiling a last time this branch to double-check the tests.
> 
> After that, I can rebase to squash the commits and this PR can be merged.

Thanks! This is also why we needed to refactor the parameters/responses/regression ... there where situations where one thought that the regression tests were executed but they weren't!

mperego:
> 
> 
> @ikalash there is a total of 4 tests which use `Solution Values` :
> 
>     * tests/small/LandIce/FO_Test/input_fo_domeAnalysis.yaml
> 
>     * tests/small/LandIce/FO_Test/input_fo_domeFoward.yaml
> 
>     * tests/small/SteadyHeat3D/input_nodeGIDArrayResponse.yaml
> 
>     * tests/small/SteadyHeat3D/inputT_nodeGIDArrayResponse.yaml
> 
> 
> Each of those tests seem to use it to verify 3 values. Even if the first two tests have `Num Values: 100`, they are only testing 3 values.
> 
> I ended up creating a Python script which converts the tests from the old and 'new' format to the current format.
> I can commit this script if some of you want, this might be useful to update the tests which are not in the repository.

Yes please commit the script. It will be helpful for a while.

kliegeois:
> > I just verified recompiling and testing the master branch.
> > The 3 tests (on the master branch) passed but did not check the computed sensitivity values with the expected ones.
> > The computed ones are not consistent with the expected ones and are the same as the values computed with the branch of this PR. Therefore, I updated the expected values to be consistent with the computed ones.
> > I am currently recompiling a last time this branch to double-check the tests.
> > After that, I can rebase to squash the commits and this PR can be merged.
> 
> Thanks! This is also why we needed to refactor the parameters/responses/regression ... there where situations where one thought that the regression tests were executed but they weren't!

I just double-check and the tests are fine!

I will clean the python script, add it, and rebase the commits tomorrow!

Thanks!

ikalash:
FYI - I am testing this right now.  Please stay tuned.  Sorry about the delay.

ikalash:
So all the CISM-Albany tests fail b/c they have incorrect response syntax.  @kliegeois : Mauro said you have a script to convert the old yaml files into the ones with the new syntax.  Can you please point me to that script, or send it to me?

kliegeois:
> So all the CISM-Albany tests fail b/c they have incorrect response syntax.  @kliegeois : Mauro said you have a script to convert the old yaml files into the ones with the new syntax.  Can you please point me to that script, or send it to me?

The script is the script tool/convert_format.py. It can update the yaml files included in a folder ` directory ` (and its subfolders) as follows:

```
python tool/convert_format.py directory

```

ikalash:
Thanks @kliegeois , I will give it a try.

ikalash:
@kliegeois : could you please look at the following tests which fail with this PR (in Albany)? 

```
The following tests FAILED:
         89 - Thermal1D (Failed)
         90 - Thermal1D_with_Source (Failed)
         91 - Thermal2D_BackwardEuler (Failed)
         92 - Thermal2D_BackwardEuler_with_Source (Failed)
         93 - Thermal2D_ForwardEuler_with_Source (Failed)
         94 - Thermal2D_ForwardEuler (Failed)
         95 - Thermal3D (Failed)
```
I'm getting errors that look like: 
```
: 
2:  Error, the parameter {name="Regression Results",type="ParameterList",value="..."}
2:  in the parameter (sub)list "Albany Parameters"
2:  was not found in the list of valid parameters!
```
Perhaps something needs to be added to the PL validation for the problem defining these tests?

mperego:
> 
> 
> @kliegeois : could you please look at the following tests which fail with this PR (in Albany)?
> 
> ```
> The following tests FAILED:
>          89 - Thermal1D (Failed)
>          90 - Thermal1D_with_Source (Failed)
>          91 - Thermal2D_BackwardEuler (Failed)
>          92 - Thermal2D_BackwardEuler_with_Source (Failed)
>          93 - Thermal2D_ForwardEuler_with_Source (Failed)
>          94 - Thermal2D_ForwardEuler (Failed)
>          95 - Thermal3D (Failed)
> ```
> 
> I'm getting errors that look like:
> 
> ```
> : 
> 2:  Error, the parameter {name="Regression Results",type="ParameterList",value="..."}
> 2:  in the parameter (sub)list "Albany Parameters"
> 2:  was not found in the list of valid parameters!
> ```
> 
> Perhaps something needs to be added to the PL validation for the problem defining these tests?

I think these tests have not been converted yet. Probably @kliegeois has `ENABLE_DEMO_PDES` off

kliegeois:
> > 
> 
> > 
> 
> > @kliegeois : could you please look at the following tests which fail with this PR (in Albany)?
> 
> > 
> 
> > ```
> 
> > The following tests FAILED:
> 
> >          89 - Thermal1D (Failed)
> 
> >          90 - Thermal1D_with_Source (Failed)
> 
> >          91 - Thermal2D_BackwardEuler (Failed)
> 
> >          92 - Thermal2D_BackwardEuler_with_Source (Failed)
> 
> >          93 - Thermal2D_ForwardEuler_with_Source (Failed)
> 
> >          94 - Thermal2D_ForwardEuler (Failed)
> 
> >          95 - Thermal3D (Failed)
> 
> > ```
> 
> > 
> 
> > I'm getting errors that look like:
> 
> > 
> 
> > ```
> 
> > : 
> 
> > 2:  Error, the parameter {name="Regression Results",type="ParameterList",value="..."}
> 
> > 2:  in the parameter (sub)list "Albany Parameters"
> 
> > 2:  was not found in the list of valid parameters!
> 
> > ```
> 
> > 
> 
> > Perhaps something needs to be added to the PL validation for the problem defining these tests?
> 
> 
> 
> I think these tests have not been converted yet. Probably @kliegeois has `ENABLE_DEMO_PDES` off

I will look at them tomorrow, I will update them and the script (if needed). Thanks for pointing them!

ikalash:
So, I got this error after I used Kim's script to convert the tests.  Thanks for looking at it @kliegeois .

ikalash:
Another thing that seems off to me.  When I convert the CISM-Albany tests and run them, I get the following error: 
```
2:                      6.483025646409e+01
2: dgdp != Teuchos::null ALBANY_ASSERT failed at /nightlyCDash/repos/Albany/src/Albany_RegressionTests.cppdgdp != Teuchos::null ALBANY_ASSERT failed at /nightlyCDash/repos/Albany/src/Albany_RegressionTests.cpp +75
2: There are Sensitivity Tests but the sensitivity vector (0, 0) is null!
```
I don't get why I have this error since there are no sensitivity tests requested in the input file. I actually explicitly turn off sensitivities in the CISM-Albany source code (https://github.com/SNLComputation/Albany/blob/master/src/LandIce/interface_with_cism/ali_driver.cpp#L679) .  I'm attaching a sample (converted) input file - perhaps you can spot what is wrong @kliegeois ?



[input_albany-cism_domeT.yaml.txt](https://github.com/SNLComputation/Albany/files/5490672/input_albany-cism_domeT.yaml.txt)

kliegeois:
> @kliegeois : could you please look at the following tests which fail with this PR (in Albany)?
> 
> ```
> The following tests FAILED:
>          89 - Thermal1D (Failed)
>          90 - Thermal1D_with_Source (Failed)
>          91 - Thermal2D_BackwardEuler (Failed)
>          92 - Thermal2D_BackwardEuler_with_Source (Failed)
>          93 - Thermal2D_ForwardEuler_with_Source (Failed)
>          94 - Thermal2D_ForwardEuler (Failed)
>          95 - Thermal3D (Failed)
> ```
> 
> I'm getting errors that look like:
> 
> ```
> : 
> 2:  Error, the parameter {name="Regression Results",type="ParameterList",value="..."}
> 2:  in the parameter (sub)list "Albany Parameters"
> 2:  was not found in the list of valid parameters!
> ```
> 
> Perhaps something needs to be added to the PL validation for the problem defining these tests?

@ikalash those tests failed because my python script was assuming that the name of the main parameter list in the yaml file was `ANONYMOUS`. This, however, was not the case for the Thermal tests; the name is `ALBANY` in those cases. I updated both the python script and the related yaml files.

Those tests are passing now.
Thanks again for pointing them!

kliegeois:
> Another thing that seems off to me. When I convert the CISM-Albany tests and run them, I get the following error:
> 
> ```
> 2: dgdp != Teuchos::null ALBANY_ASSERT failed at /nightlyCDash/repos/Albany/src/Albany_RegressionTests.cppdgdp != Teuchos::null ALBANY_ASSERT failed at /nightlyCDash/repos/Albany/src/Albany_RegressionTests.cpp +75
> 2: There are Sensitivity Tests but the sensitivity vector (0, 0) is null!
> ```
> 
> I don't get why I have this error since there are no sensitivity tests requested in the input file. I actually explicitly turn off sensitivities in the CISM-Albany source code (https://github.com/SNLComputation/Albany/blob/master/src/LandIce/interface_with_cism/ali_driver.cpp#L679) . I'm attaching a sample (converted) input file - perhaps you can spot what is wrong @kliegeois ?
> 
> [input_albany-cism_domeT.yaml.txt](https://github.com/SNLComputation/Albany/files/5490672/input_albany-cism_domeT.yaml.txt)

@ikalash I think that I found the issue. I did not achieve to run the test myself; however, I am pretty confident that the issue has gone. Could you try it once again?

kliegeois:
@mperego I just verified the two issues that initiated #638:
- the error while using `Response Vector Index: 1`,
- the error `Thyra::Exceptions::IncompatibleVectorSpaces`.

I confirm that both of them are gone while evaluating my tests.
After the merge of this PR, #638 can be closed.

ikalash:
Sorry about the delay.  Everything looks good now.  You have the thumbs up from me to merge this in.

mperego:
@kliegeois have a look at my last commit. If you are fine with that I think we can merge.

One thing that I noticed is that there are some inconsistencies on how we capitalize the prepositions: `Number of Parameters` `Sum Of Responses` `Number of Responses` `Regression For Response`. 
Should we fix these? If so, what convention should we pick?

kliegeois:
@ikalash thanks for having retested the failing tests!

@mperego thanks for this cleanup!
I am fine with your changes.

I did not think about the capital letter of the prepositions; I just followed your suggestion.
But, as you said, I think that it would be better to follow only one convention. I am fine with either of them. 
It seems that, in the older formats, the prepositions are capitalized. We could follow this convention.

kliegeois:
@mperego there are some tests that failed since your commit.
I am looking into that and I am capitalizing the prepositions.

mperego:
> 
> 
> @mperego there are some tests that failed since your commit.
> I am looking into that and I am capitalizing the prepositions.

Ah.. sorry, all tests pass on my Workstation

kliegeois:
> > @mperego there are some tests that failed since your commit.
> > I am looking into that and I am capitalizing the prepositions.
> 
> Ah.. sorry, all tests pass on my Workstation

No problem! Those tests failed:
```
The following tests FAILED:
        106 - LandIce_FO_Dome_Ascii (Failed)
        107 - LandIce_FO_Dome_Restart (Failed)
        108 - FO_MMS_SincCosGlen_Epetra (Failed)
        109 - FO_MMS_SincCosGlen_Tpetra (Failed)
        112 - FO_MMS_SinExp_Neumann_Epetra (Failed)
        113 - FO_MMS_SinExp_Neumann_Tpetra (Failed)
        114 - FO_MMS_CosExp_Basal_Epetra (Failed)
        115 - FO_MMS_CosExp_Basal_Tpetra (Failed)
        116 - FO_MMS_CosExp_BasalAll_Epetra (Failed)
        117 - FO_MMS_CosExp_BasalAll_Tpetra (Failed)
        145 - FO_GIS_Humboldt_Tpetra_FROSch (Failed)
        155 - FO_GIS_AdjointSensitivity_BasalFriction_Epetra (Failed)
        156 - FO_GIS_AdjointSensitivity_BasalFriction_Tpetra (Failed)
        157 - FO_GIS_Analysis_BasalFriction (Failed)
        158 - FO_GIS_Analysis_BasalFriction_Memoization (Failed)
        170 - FO_Thermo_Humboldt_fluxDiv (Failed)
Errors while running CTest
```
but I think that I resolved the issue.

kliegeois:
@mperego this was actually great that your commit generated those failed tests. There were some bugs that have been highlighted by your commit but that were not part of your committed work.

I am waiting for the end of the test battery and this PR can be merged!

mperego:
@kliegeois That's great. I still don't understand why I didn't see these failures on my workstation.

kliegeois:
> @kliegeois That's great. I still don't understand why I didn't see these failures on my workstation. 

@mperego no problem! Can we merge the PR? I do not have the right access to do so.

ikalash:
Hmm, that's weird that you are seeing failures - when I ran the Albany tests in my nightly test environment, everything was clean.

mperego:
Anyway, the changes that @kliegeois made make sense to me, and all tests pass on my worskatation, so I'm going to merge the PR unless I hear any objection.

We'll also have to better handle the regression in the Analysis case.. but we can do that later.

mperego:
Thanks @kliegeois for all this work!

kliegeois:
Thanks @mperego !

PR #642, opened 2020-10-29
Add cmake options to choose the AD type for Hessian-vector product computations.

kliegeois:
This PR is intended to add the plumbing work to configure the AD type used during the Hessian-vector product computations.

Moreover, this PR removes the definition of the Hessian types (i.e. the types which were intended to be used for the explicit computation of the matrix); those types were never used and will not be used as another strategy will be used to reconstruct the Hessian matrix.

This PR has been tested both with SFad and DFad and passed in both cases.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -657,6 +657,26 @@ ELSE()
   Options: SFad, SLFad, DFad")
 ENDIF()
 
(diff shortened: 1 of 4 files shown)


mperego (approved):
Looks good. Thanks @kliegeois also for the clean up!

PR #645, opened 2020-12-02
Remove Albany warnings

kliegeois:
This PR removes the warnings generated when compiling Albany with gcc 7.5.0 with the following flags:

```-Wall -Wno-clobbered -Wno-vla -Wno-pragmas -Wno-unknown-pragmas -Wno-unused-local-typedefs -Wno-literal-suffix -Wno-deprecated-declarations -Wno-misleading-indentation -Wno-int-in-bool-context -Wno-maybe-uninitialized -Wno-nonnull-compare -Wno-address -Wno-inline -Wno-return-type -Werror```

Those flags are the ones used by the gcc 7.2.0 debug build of the Trilinos autotester as illustrated [here](https://github.com/trilinos/Trilinos/blob/master/cmake/std/PullRequestLinuxGCC7.2.0DebugTestingSettings.cmake).

They enable all the warnings except those that are explicitly disabled and treat any warning as an error which stops the building process.

The modifications included in this PR are mainly of 3 types:
- Reordering of the initialization of member data in the constructor to follow the order of their declaration,
- Removal of signed/unsigned comparison,
- Removal of unused functions and variables.

I verified that the tests passed on Blake.

As discussed yesterday during the Albany weekly meeting, we should probably use a nightly build with those flags set to maintain the build warning free.

@mperego @ikalash


bartgol:
Thanks Kim! Huge effort!

I wasn't at the call yesterday (but I am planning on significantly ramping up my Albany/Prospect time this month, and going forward), so I apologize if you already discussed the warnings.

Generally, I don't like to suppress warnings, but I agree that some of them are relatively "innocuous", like an unused typedef. However, some of the warnings suppressed by these options might be a bit risky. In particular, the following may be risky to use, and I would argue against using them: `-Wno-vla -Wno-maybe-uninitialized -Wno-pragmas -Wno-unknown-pragmas`. Others are debatable, but those four could cause memory errors, or bad/wrong CPP parsing, all of which may be hard to track.

ikalash:
Thanks @kliegeois !  Could you please provide the Trilinos and Albany configure scripts that you used to get this to work, so I can test them and ultimately create a nightly with the options?

kliegeois:
Thanks a lot for your comments.

@bartgol,  I used the same flags as the Trilinos autotester.
My initial goal was to be able to use one unique set of flags when I compile a branch before creating a PR to Trilinos and test Albany against it.
I am totally open to be more restrictive on the warnings for Albany; I do not have a strong feeling on which warning we should consider.

@ikalash I just send you the configuration scripts, thanks a lot.

kliegeois:
I observed this morning that the build that I used to remove the warnings was not enabling `ALBANY_DEMO_PDES`.
The added commit allows to compile the related files without warnings.
I verified; all the tests passed except one which times out (FO_GIS_Gis20km_Tpetra) on Blake with debug build.

In parallel, I created [PR #8414](https://github.com/trilinos/Trilinos/pull/8414) to remove remaining warnings from Trilinos/develop.
I should have started by that, sorry.

kliegeois:
@mperego has just accepted [PR #8414 ](https://github.com/trilinos/Trilinos/pull/8414) which is now merged (thanks by the way)!

ikalash:
I ended up fixing the warnings as errors in DTK.  I'll set up a new nightly build on blake once this is merged in.

kliegeois:
Thanks a lot @ikalash ! Yes, sorry, it seems that my configuration was not enabling all Albany folders.

ikalash:
Just to follow up: I did end up creating a warning free nightly on Blake.  It’s running starting last night.

PR #650, opened 2021-01-07
add std namespace to allow build with ALBANY_VERBOSE enabled for gcc 9.x

koomie:


bartgol:
What's the issue with gcc 9.x? Do you get a warning without `std::`?

Note: I'm always in favor of using `std::blah` instead of `blah`, especially since std is such a short namespace name. I'm just curious on what the issue with gcc 9.x is.

Btw, there is _a ton_ of usage of std stuff without namespace, due to usage of `using std::blah`. Do those cause issues as well? If not, then I'm curious about what's peculiar with the changes in your commit that doesn't apply to other std stuff usage.

koomie:
No, it's just a garden-variety compile failure. Presumably not related to this version of gcc but just wanted to mention what I was building with locally.  

I didn't look thru the header files included by Albany_StateManager.cpp but I assume they don't call out using the `std` namespace.  Maybe other src files do - I only encountered a build issue with this .cpp file.

Anyway, I was guessing this was just an oversight since the majority of the file makes copious use of `std::blah`. A local build with `ALBANY_VERBOSE` enabled completed ok for me after making this change.

bartgol:
Oh, I see the issue now. I wonder if anyone ever compiled with ALBANY_VERBOSE on. There was no way that file compiled otherwise. Unless due to some odd configuration, this file included some header which had `using std::cout` or something analogous, which is unlikely.

Anyhow, thanks for the catch. You can merge whenever you like.

bartgol (approved):

PR #651, opened 2021-01-08
Pyalbany: update the parallelEnv and add more examples 

kliegeois:
This PR updates PyAlbany to avoid cases where Kokkos initialize was called twice.

Moreover, this PR includes a performance analysis example and the examples which will be discussed during the live demo/tutorial of next week.

@mperego @ikalash


mperego (approved):
Looks good! Thanks.

kliegeois:
Thanks @mperego !

PR #654, opened 2021-01-26
StokesFO Kokkos Refactor and Boundary Layout Update

mcarlson801:
This PR contains updates to the boundary conditions of the StokesFO problem. All sideset evaluators used in StokesFO have been updated to use the new collapsed sidesets layout while still allowing other problems to use these evaluators with the old sideset layouts. Additionally, the remaining performance intensive evaluators used in StokesFO have been ported to Kokkos. A summary of the changes in this PR is as follows:

- StokesFO sideset evaluators have been updated to allow new or old sideset layouts
- StokesFO evaluators have been fully ported to Kokkos and the total fill time for GPU builds has been reduced significantly
- Added a 'localDOFview' structure to PHAL_Workset that is initialized in STKDiscretization.cpp::computeSideSets(). This structure is a kokkos view that is used in GatherVerticallyContractedSolution to index into the underlying Thyra vector in a coalesced fashion on GPU.
- Added control flow to StokesFOBase so that other problems that extend StokesFOBase use the old sideset layout until they can be updated.
- Views have been added to DiscretizationUtils so that computeBasisFunctionsSide has access to the appropriate data when running on GPU.


src/evaluators/response/PHAL_ResponseSquaredL2DifferenceSide_Def.hpp
```diff
@@ -118,55 +120,88 @@ evaluateFields(typename Traits::EvalData workset)
...
 
-  std::vector<ScalarT> diff_1;
-  std::vector<std::vector<ScalarT>> diff_2;
-  if (fieldDim==1)
-    diff_1.resize(dims[3]);
-  else if (fieldDim==2)
-    diff_2.resize(dims[3],std::vector<ScalarT>(dims[3]));
+  size_t diffDims = useCollapsedSidesets ? dims[2] : dims[3];
+  Kokkos::View<ScalarT*,  PHX::Device> diff_1("diff_1", diffDims);
+  Kokkos::View<ScalarT**, PHX::Device> diff_2("diff_@", diffDims, diffDims);
```
jewatkins, line 125: "diff_2"

src/evaluators/response/PHAL_ResponseSquaredL2DifferenceSide_Def.hpp
```diff
@@ -118,55 +120,88 @@ evaluateFields(typename Traits::EvalData workset)
...
+
+              for (size_t i=0; i<dims[3]; ++i)
+                for (size_t j=0; j<dims[3]; ++j)
+                  for (size_t k=0; k<dims[3]; ++k)
+                    for (size_t l=0; l<dims[3]; ++l)
+                      sq += metric(sideSet_idx,qp,k,i)*diff_2(i,j) * metric(sideSet_idx,qp,j,l)*diff_2(l,k);
+              break;
+          }
+          sum += sq * w_measure(sideSet_idx,qp);
+        }
```
jewatkins, line 170: Need to discuss creating kernels for these

src/evaluators/utility/PHAL_ComputeBasisFunctionsSide_Def.hpp
```diff
@@ -566,7 +431,12 @@ evaluateFields(typename Traits::EvalData workset)
       for (std::size_t iCell=0; iCell < numCells_; ++iCell)
         for (unsigned int icoor=0; icoor<numCellDims; ++icoor)
           for (unsigned int qp=0; qp<numSideQPs; ++qp)
-            normals(sideSet.cellsOnSide(side,iCell),side,qp, icoor) = normals_view(iCell,qp,icoor);
+          {
+            if (useCollapsedSidesets)
+              normals(sideSet.sideSetIdxOnSide(side,iCell),qp, icoor) = normals_view(iCell,qp,icoor);
```
jewatkins, line 436: I forgot, what was the reason for not making this a kernel?

src/evaluators/utility/PHAL_FieldFrobeniusNorm_Def.hpp
```diff
@@ -198,66 +200,119 @@ void FieldFrobeniusNormBase<EvalT, Traits, ScalarT>::evaluateFields (typename Tr
...
+                norm += std::pow(field(sideSet_idx,i,dim),2);
+              }
+              field_norm(sideSet_idx,i) = std::sqrt(norm + regularization);
+            }
+          }
+        }
+        break;
+      default:
+        TEUCHOS_TEST_FOR_EXCEPTION(true, std::logic_error, "Error! Invalid field layout.\n");
+    }
```
jewatkins, line 257: need to discuss if kernels are needed for these too

src/evaluators/utility/PHAL_MapToPhysicalFrameSide_Def.hpp
```diff
@@ -76,32 +79,55 @@ postRegistrationSetup(typename Traits::SetupData d,
...
-void MapToPhysicalFrameSide<EvalT, Traits>::evaluateFields(typename Traits::EvalData workset)
-{
-  if (workset.sideSets->find(sideSetName)==workset.sideSets->end()) {
-    return;
+KOKKOS_INLINE_FUNCTION
+void MapToPhysicalFrameSide<EvalT, Traits>::
+operator() (const MapToPhysicalFrameSide_Tag& tag, const int& sideSet_idx) const {
+  
+  // Get the local data of side and cell
+  const int cell = sideSet.elem_LID(sideSet_idx);
```
jewatkins, line 90: I think cell is not needed in this one

src/LandIce/evaluators/LandIce_GatherVerticallyContractedSolution_Def.hpp
```diff
@@ -89,67 +100,134 @@ GatherVerticallyContractedSolution(const Teuchos::ParameterList& p,
...
+  TEUCHOS_TEST_FOR_EXCEPTION(workset.sideSetViews.is_null(), std::logic_error,
                              "Side sets defined in input file but not properly specified on the mesh.\n");
 
-  const Albany::SideSetList& ssList = *(workset.sideSets);
-  Albany::SideSetList::const_iterator it = ssList.find(this->meshPart);
+  // Get total number of layers
+  const Albany::LayeredMeshNumbering<GO>& layeredMeshNumbering = *workset.disc->getLayeredMeshNumbering();
+  this->numLayers = layeredMeshNumbering.numLayers;
+
+  // Compute quadWeights
```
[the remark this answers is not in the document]
mcarlson801: This was my plan originally but I don't know how I can get access to the LayeredMeshNumbering in the constructor or postRegSetup. I'll look into it to see if there is a way I can get it during discretization.

src/LandIce/problems/LandIce_StokesFOBase.cpp
```diff
@@ -165,8 +167,11 @@ void StokesFOBase::buildProblem (Teuchos::ArrayRCP<Teuchos::RCP<Albany::MeshSpec
       unsigned int numSideNodes    = sideBasis[ssName]->getCardinality();
       unsigned int numSideQPs      = sideCubature[ssName]->getNumPoints();
 
+      if(Teuchos::GlobalMPISession::getRank() == 0) printf("Layout %s: worksetSize = %d, sideMeshSpecs.worksetSize = %d\n", ssName.c_str(), worksetSize, sideMeshSpecs.worksetSize);
+
```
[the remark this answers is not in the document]
mcarlson801: Just for debugging. Thanks for catching that. It did confirm that sideMeshSpecs.worksetSize is not enough to determine the actual worksetSize.

src/disc/stk/Albany_STKDiscretization.cpp
```diff
@@ -2024,12 +2029,71 @@ STKDiscretization::computeSideSets()
...
 
       current_local_index[ss_key]++;
 
       ss_it++;
     }
   }
+
+  // 6) Populate localDOFViews for GatherVerticallyContractedSolution
+  const int maxSideNodes = 6;
+  const Albany::LayeredMeshNumbering<GO>& layeredMeshNumbering = *(stkMeshStruct->layered_mesh_numbering);
```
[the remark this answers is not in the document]
mcarlson801: I do actually check if layered_mesh_numbering exists but a few lines later for some reason. I'll move that up to avoid some unnecessary cycles and see if there is a way to check if it is needed by any evaluators.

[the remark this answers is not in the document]
jewatkins: The latter might be a bit difficult. I don't see a clear parameter that would turn this on so it's probably okay to only construct this if the layered mesh numbering exists. It's probably not a lot of memory compared to the other memory issues we have.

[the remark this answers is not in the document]
bartgol: I don't think the disc has any way to know if the LMN is indeed needed.

src/evaluators/utility/PHAL_ComputeBasisFunctionsSide_Def.hpp
mcarlson801: I think there was a reason at one point but the way I've rewritten it, it would now be easy to turn it into a kernel. I'll also add one for populating physPointsCell.

jewatkins: okay great thanks

src/LandIce/evaluators/LandIce_BasalFrictionCoefficientGradient_Def.hpp
```diff
@@ -171,73 +169,137 @@ void BasalFrictionCoefficientGradient<EvalT, Traits>::evaluateFields (typename T
     power  = powerParam(0);
   }
```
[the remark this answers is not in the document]
mcarlson801: BasalFrictionCoefficient and BasalFrictionCoefficientGradient now have Kokkos kernels. The response kernels that you mentioned in other comments here are easy enough to add kernels for so I'm going to add those as well when I have some time tomorrow. Once all kernels are in, I'll do a performance comparison and post the results.

src/Albany_StateManager.cpp
```diff
@@ -575,14 +575,9 @@ Albany::StateManager::registerSideSetStateVariable_collapsed(
...
-  } else {
-
-    TEUCHOS_TEST_FOR_EXCEPTION(
-        true,
-        std::logic_error,
-        "Error! The given layout does not appear to be that of a collapsed side set "
-        "field.\n");
-
+  }
+  else {
```
bartgol, line 579: Why not erroring out? It seemed appropriate, if the input was unexpected.

src/LandIce/evaluators/LandIce_EffectivePressure_Def.hpp
```diff
@@ -77,6 +79,34 @@ postRegistrationSetup(typename Traits::SetupData /* d */,
...
+  for (int pt=0; pt<numPts; ++pt) {
+    // N = P_o-P_w
+    N (sideSet_idx,pt) = (1-alpha)*P_o(sideSet_idx,pt);
+  }
+
+}
+
+template<typename EvalT, typename Traits, bool IsStokes, bool Surrogate>
+KOKKOS_INLINE_FUNCTION
+void EffectivePressure<EvalT, Traits, IsStokes, Surrogate>::
```
bartgol, line 100: This is fine, but for the record, given the small size, I believe keeping both cases in a single kernel (with an if to determine which for loop to execute) would have been good enough. The branch predictor will likely hit the correct branch.

src/LandIce/evaluators/LandIce_GatherVerticallyContractedSolution_Def.hpp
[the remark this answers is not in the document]
bartgol: I think the struct LayeredMeshNumbering is probably putting a lot of info in the same struct. Some, belongs to the MeshStruct itself (which is where the LayeredMeshNumbering ptr is currently stored), but the number of layers is more of a mesh specs kind of info (much like the workset size).

Perhaps we could add a couple of vars to the MeshSpecs:
```
bool is_layered;
int num_layers;
```
Or even just `num_layers`, with the convention that, say, a negative number means not-layered. The number of layers can also be retained in the LayeredMeshNumbering struct, for ease of use (only one struct to lookup layered mesh info). But having this one bit of info in the mesh specs might help to do this kind of setup in the ctor phase.

src/disc/stk/Albany_STKDiscretization.cpp
```diff
@@ -2024,12 +2029,71 @@ STKDiscretization::computeSideSets()
...
+      lssList[ss_key].sideSetIdxOnSide    = Kokkos::subview(globalSideSetViews[ss_key].sideSetIdxOnSide,    current_index, Kokkos::ALL(), Kokkos::ALL() );
 
       current_local_index[ss_key]++;
 
       ss_it++;
     }
   }
+
+  // 6) Populate localDOFViews for GatherVerticallyContractedSolution
+  const int maxSideNodes = 6;
```
bartgol, line 2041: Do you set '6' cause you're thinking about quadratic triangles? I don't think we support quadratic elements in Albany. But then, if we did, 6 would not fit quadratic Quads, which would show up when using prisms.

You could be more tight, and loop over the cell_topo sides, and count their nodes. That will tell you exactly what 'max_side_nodes' should be. I suspect 6 is an overshoot for most of our cases. A tighter upper bound might help with caching.

src/evaluators/interpolation/PHAL_DOFCellToSide.hpp
```diff
@@ -59,6 +59,7 @@ class DOFCellToSideBase : public PHX::EvaluatorWithBaseImpl<Traits>,
   enum LayoutType
   {
     CELL_SCALAR = 1,
+    CELL_SCALAR_SIDESET,
```
bartgol, line 62: So you are only producing an output with collapsed layout for the "P0-scalar" and P1-scalar/vector cases, is that correct?

That's fine, for now, but it's worth making a TODO action item, to propagate collapsed layout also to P0 vector/tensor and P1-tensor quantities.

Note: this can be done when the non-collapsed side layouts are completely removed.

src/problems/Albany_Layouts.cpp
```diff
@@ -201,14 +201,24 @@ Albany::Layouts::Layouts (int worksetSize, int numVertices, int numNodes, int nu
   useCollapsedSidesets = collapsed_sidesets;
 
   // Collapsed sideset layouts to ensure contiguous memory access for efficient GPU evaluation
+  // TODO: using meshspecs struct to set sideset workset size doesn't work for all cases,
+  //        so we are using the naive lower bound here until we can find a better bound
+  // sidesetWorksetSize = (sidesetWorksetSize > 0) ? sidesetWorksetSize : worksetSize*numSides;
+  sidesetWorksetSize = worksetSize*numSides;
```
bartgol, line 207: I was just thinking, that maybe we could switch to `PHX::Layout` instead of `PHX::MDALayout`. The former allows to defer the specification of the extents to a later time. For us, that'd be after the whole mesh has been created, so that we know the exact number of sides on the sideset, and can allocate smaller side layouts.

The main drawback is that with dynamic layouts, you _need_ to call `setFieldData` for each required/evaluated field in `postRegistrationSetup`, while now fields are allocated at construction time, so the `setFieldData` step is superfluous (and sometimes I don't even do it in the evaluators I write). Also, you would not be able to get the number of nodes or quad points from a layout at construction time inside the evaluators (like we sometimes do).

Just a thought, for later.

P.s: obviously, `PHX::Layout` still inherits from `PHX::DataLayout`, so the modifications would be minimal.

bartgol (approved):
Monstruous PR! Great work. I had some comments, but I believe they are just thoughts, for future development and/or to spark discussions.

src/Albany_StateManager.cpp
mcarlson801: The exception that I removed here was copy and pasted from the non-collapsed implementation and only throws if there is a layout with only one field. Since the old sideset layouts have to have two fields (Cell, Side) at minimum, there was no way it could have only one field. However, with the collapsed layout, it is fine to have a single field (Side). From what I could gather, for a single field layout, all we need to do here is get the dimensions.

src/disc/stk/Albany_STKDiscretization.cpp
mcarlson801: Yeah, this was a bandaid thing that I didn't get around to fixing. I'll loop over the sides to get the tightest bound here.

src/Albany_StateManager.cpp
bartgol: I'm just weary of 'else' statements, cause they catch whatever didn't fit in the previous if(s). Maybe we can put a test for exception, ensuring that the rank is indeed 1? Right now, you didn't add, say, `qp_tensor3` as a collapsed layout to Albany::Layout. But if tomorrow someone did it, and used it, adding a check in here might save hours of debugging.

bartgol: Note: alternatively, you could add a check on the layout rank at the beginning of the fcn, and support only rank<=4 for now.

mcarlson801: That's a good point. I'll rewrite this to make sure nothing gets through the cracks. I assume that means an else-if for rank 1 and then throw an exception in the else but I'll double check to make sure that covers all cases.

mperego:
@mcarlson801 @jewatkins  I'm going to start working on some evaluators modified by this PR. Do you suggest that I base my work on this branch?

jewatkins:
> @mcarlson801 @jewatkins I'm going to start working on some evaluators modified by this PR. Do you suggest that I base my work on this branch?

If they're sideset evaluators, I would say yes. Although if you need things in master soon maybe not. I think there's still a few things Max needs to fix and it would be good to have the GPU performance tests up before merging this. See https://github.com/SNLComputation/Albany/issues/661 and https://github.com/trilinos/Trilinos/issues/8727.

mperego:
> > @mcarlson801 @jewatkins I'm going to start working on some evaluators modified by this PR. Do you suggest that I base my work on this branch?
> 
> If they're sideset evaluators, I would say yes. Although if you need things in master soon maybe not. I think there's still a few things Max needs to fix and it would be good to have the GPU performance tests up before merging this. See #661 and [trilinos/Trilinos#8727](https://github.com/trilinos/Trilinos/issues/8727).

I see. Yes, I'm going to work on the basal friction evaluator.

mperego:
@mcarlson801 when I run the testsuit using this branch, a lot of tests fail with this error:
```
3: p=3: *** Caught standard std::exception of type 'std::logic_error' :
3: 
3:  /ascldap/users/mperego/Workspace/albany/sources/albany-src/src/disc/stk/Albany_STKDiscretization.hpp:210:
3:  
3:  Throw number = 1
3:  
3:  Throw test that evaluated to true: true
3:  
3:  Local DOF view was requested but it has not been created.
```
Have you seen that?

mcarlson801:
> @mcarlson801 when I run the testsuit using this branch, a lot of tests fail with this error:
> 
> ```
> 3: p=3: *** Caught standard std::exception of type 'std::logic_error' :
> 3: 
> 3:  /ascldap/users/mperego/Workspace/albany/sources/albany-src/src/disc/stk/Albany_STKDiscretization.hpp:210:
> 3:  
> 3:  Throw number = 1
> 3:  
> 3:  Throw test that evaluated to true: true
> 3:  
> 3:  Local DOF view was requested but it has not been created.
> ```
> 
> Have you seen that?

@mperego Hmm, I must not have run unit tests since adding that exception. I'll look into this and give you an update shortly.

mcarlson801:
> @mcarlson801 when I run the testsuit using this branch, a lot of tests fail with this error:
> 
> ```
> 3: p=3: *** Caught standard std::exception of type 'std::logic_error' :
> 3: 
> 3:  /ascldap/users/mperego/Workspace/albany/sources/albany-src/src/disc/stk/Albany_STKDiscretization.hpp:210:
> 3:  
> 3:  Throw number = 1
> 3:  
> 3:  Throw test that evaluated to true: true
> 3:  
> 3:  Local DOF view was requested but it has not been created.
> ```
> 
> Have you seen that?

@mperego I figured out what the issue is. I added that exception to "getLocalDOFViews" so that we knew if a workset was grabbing that view if it didn't need it. However, in Albany_Application, "getLocalDOFViews" is always called for every workset regardless if it is needed.

I've reverted this change temporarily so that I'm not interrupting your workflow but a proper fix is coming later today in addition to a fix for an issue that only shows up on GPU builds.

mcarlson801:
> @mcarlson801 when I run the testsuit using this branch, a lot of tests fail with this error:
> 
> ```
> 3: p=3: *** Caught standard std::exception of type 'std::logic_error' :
> 3: 
> 3:  /ascldap/users/mperego/Workspace/albany/sources/albany-src/src/disc/stk/Albany_STKDiscretization.hpp:210:
> 3:  
> 3:  Throw number = 1
> 3:  
> 3:  Throw test that evaluated to true: true
> 3:  
> 3:  Local DOF view was requested but it has not been created.
> ```
> 
> Have you seen that?

@mperego It looks like this change wasn't enough to pass unit tests. I'm working on a more comprehensive fix and I'll let you know as soon as this branch is testing clean. Sorry for the inconvenience, I'll try to get this working ASAP.

mperego:
@mcarlson801  Thanks. Do you mind if I cherry-pick your last commit on the my version where I just merged the master and force-push? Otherwise I need to do the merging again

mcarlson801:
> @mcarlson801 Thanks. Do you mind if I cherry-pick your last commit onto my version where I just merged the master and force-push? Otherwise I need to do the merging again

@mperego Yeah, if that works best for you, go for it. I'll work around whatever you decide to do, shouldn't be a problem.

mperego:
@mcarlson801 It's probably best if I hold off while you are fixing things. I won't mess up with this branch

src/evaluators/utility/PHAL_FieldFrobeniusNorm_Def.hpp
```diff
@@ -88,8 +90,8 @@ FieldFrobeniusNormBase (const Teuchos::ParameterList& p,
     TEUCHOS_TEST_FOR_EXCEPTION (!dl->isSideLayouts, Teuchos::Exceptions::InvalidParameter,
                                 "Error! The layouts structure does not appear to be that of a side set.\n");
 
-    field      = decltype(field)(fieldName, dl->node_vector);
-    field_norm = decltype(field_norm)(fieldNormName, dl->node_scalar);
+    field      = decltype(field)(fieldName, useCollapsedSidesets ? dl->node_vector_sideset : dl->node_vector);
+    field_norm = decltype(field_norm)(fieldNormName, useCollapsedSidesets ? dl->node_scalar_sideset : dl->node_scalar);
 
     dl->node_vector->dimensions(dims);
```
mperego, line 96: @mcarlson801  I think this should be changed to `dl->node_vector_sideset` when `useCollapsedSideset` is `true`. Same in the next two if statemets

mcarlson801: Yeah, this does seem to be a mistake on my part. The fix will come in with my next commit.

src/LandIce/evaluators/LandIce_BasalFrictionCoefficient_Def.hpp
```diff
@@ -228,6 +235,258 @@ postRegistrationSetup (typename Traits::SetupData d,
...
+void BasalFrictionCoefficient<EvalT, Traits, EffPressureST, VelocityST, TemperatureST>::
+operator() (const ExpGivenFieldParam_Nodal_Tag& tag, const int& cell) const {
+  for (unsigned int ipt=0; ipt<dim; ++ipt) {
+    beta(cell,ipt) = std::exp(given_field_param(cell,ipt));
+  }
+}
+template<typename EvalT, typename Traits, typename EffPressureST, typename VelocityST, typename TemperatureST>
+KOKKOS_INLINE_FUNCTION
+void BasalFrictionCoefficient<EvalT, Traits, EffPressureST, VelocityST, TemperatureST>::
+operator() (const ExpGivenFieldParam_Tag& tag, const int& cell) const {
```
mperego, line 315: @mcarlson801  I think you need to swap the Tag types of this and the following function.

src/evaluators/state/PHAL_SaveSideSetStateField_Def.hpp
```diff
@@ -59,6 +59,8 @@ SaveSideSetStateField (const Teuchos::ParameterList& p,
 
   savestate_operation = Teuchos::rcp(new PHX::Tag<ScalarT>(fieldName, dl->dummy));
 
+  useCollapsedSidesets = dl->useCollapsedSidesets;
+
   this->addDependentField (field.fieldTag());
   this->addEvaluatedField (*savestate_operation);
```
mperego, line 66: @mcarlson801  The two TEUCHOS_TEST_FOR_EXCEPTION below, need to be modified for collapsed side sets (`.size() < 2` and `.name(1) `)

src/evaluators/state/PHAL_SaveSideSetStateField_Def.hpp
```diff
@@ -382,12 +441,12 @@ saveNodeState(typename Traits::EvalData workset)
...
+    sideSet = workset.sideSetViews->at(sideSetName);
+    for (int sideSet_idx = 0; sideSet_idx < sideSet.size; ++sideSet_idx)
     {
       // Get the data that corresponds to the side
-      const int cell     = it_side.elem_LID;
-      const int side     = it_side.side_local_id;
+      const int cell = sideSet.elem_LID(sideSet_idx);
+      const int side = sideSet.side_local_id(sideSet_idx);
 
       switch (dims.size())
```
mperego, line 451: @mcarlson801 this `switch` needs to be modified for collapsed side sets (`dims.size()` is either 2 or 3), also when you access `dims`, you need to reduce the index by one.

src/evaluators/state/PHAL_SaveSideSetStateField_Def.hpp
```diff
@@ -332,12 +391,12 @@ saveNodeState(typename Traits::EvalData workset)
...
+    sideSet = workset.sideSetViews->at(sideSetName);
+    for (int sideSet_idx = 0; sideSet_idx < sideSet.size; ++sideSet_idx)
     {
       // Get the data that corresponds to the side
-      const int cell     = it_side.elem_LID;
-      const int side     = it_side.side_local_id;
+      const int cell = sideSet.elem_LID(sideSet_idx);
+      const int side = sideSet.side_local_id(sideSet_idx);
 
       // Notice: in the following, we retrieve the id of the stk node using the 3d mesh.
```
mperego, line 401: @mcarlson801 this `switch` needs to be modified for collapsed side sets (`dims.size()` is either 2 or 3), also when you access `dims`, you need to reduce the index by one.

mperego:
@mcarlson801 I notice that the `StokesFOThermoCoupled` problem is still using the non collapsed sidesets. Are you planning to change that in this PR or is it separated by this? 
Sorry for keeping pinging you. As I mentioned I'm working at different evaluators that are based on sideset and I'd rather do that only once using the new layout and Kokkos implementation.

mcarlson801:
> @mcarlson801 I notice that the `StokesFOThermoCoupled` problem is still using the non collapsed sidesets. Are you planning to change that in this PR or is it separated by this?
> Sorry for keeping pinging you. As I mentioned I'm working at different evaluators that are based on sideset and I'd rather do that only once using the new layout and Kokkos implementation.

@mperego The only problem that will be updated in this PR is StokesFO. The other problems that inherit from StokesFOBase will be changed in a later PR.

Once I have tested this next (and hopefully final) commit for this PR, I will take a look at the StokesFOThermoCoupled problem to see what is needed to get it working with the collapsed sidesets. It may be that all of the evaluators for StokesFOThermoCoupled are ready for the new layout and in that case I could just switch a flag and it would use the new layouts. I'll update you when I know more.

mperego:
OK, thanks!  I'll start working on that. For sure `w_Resid` evaluator needs to be converted to have StokesFOThermo coupled to work.

mperego:
@mcarlson801 B.t.w. I confirm that  `w_Resid` is the only evaluator that needs to be changed in order to use the collapsed layout in `StokesFOThermoCopled`.

mperego:
@mcarlson801 thanks for all this work. I've merged master into the branch. @mcarlson801 @jewatkins is anything preventing this from being merged? More testing?

jewatkins:
> @mcarlson801 thanks for all this work. I've merged master into the branch. @mcarlson801 @jewatkins is anything preventing this from being merged? More testing?

I think there's still some lingering issues that Max was going to address today.

mcarlson801:
> > @mcarlson801 thanks for all this work. I've merged master into the branch. @mcarlson801 @jewatkins is anything preventing this from being merged? More testing?
> 
> I think there's still some lingering issues that Max was going to address today.

The PR now has all the changes I was planning to make and is currently passing all unit tests, the only thing left is testing for GPU builds and then I think Jerry wants to look over it once more for a final review.

jewatkins:
I think the trilinos gpu fix is in develop so I'll go ahead and do a final review/test and push if everything looks good.

src/evaluators/utility/PHAL_ComputeBasisFunctionsSide_Def.hpp
```diff
@@ -228,77 +228,22 @@ operator() (const ComputeBasisFunctionsSide_Collapsed_Tag& tag, const int& sideS
...
-        GradBF(cell,side,node,qp,ider)=0;
-        for (unsigned int jder=0; jder< numSideDims; ++jder) {
-          GradBF(cell,side,node,qp,ider) += inv_metric(cell,side,qp,ider,jder)*grad_at_cub_points(node,qp,jder);
-        }
-      }
-    }
-  }
+  for (unsigned int icoor=0; icoor<numCellDims; ++icoor)
+    for (unsigned int qp=0; qp<numSideQPs; ++qp)
+      normals(sideSet.sideSetIdxOnSide(currentSide,iCell),qp, icoor) = normals_view(iCell,qp,icoor);
```
jewatkins, line 246: this is a race condition if multiple cells touch a single "sideset index on side". I'm not sure if that's the case here.

jewatkins (changes requested):
A few more things. I think we're close. All performance tests pass except one (I think it's a known stability issue which I need to look into).

mperego:
@mcarlson801 what's the ETC for this PR? We have enother PR that is ready to be merged in. My understanding is that they are complitely orthogonal. Should we merge the other PR first?

jewatkins:
> @mcarlson801 what's the ETC for this PR? We have enother PR that is ready to be merged in. My understanding is that they are complitely orthogonal. Should we merge the other PR first?

There's a few minor edits and Max is seeing nondeterministic behavior in a performance test that needs to be tracked down. I don't see an issue with merging the other PR first.

src/LandIce/evaluators/LandIce_BasalFrictionCoefficient_Def.hpp
```diff
@@ -228,6 +235,258 @@ postRegistrationSetup (typename Traits::SetupData d,
...
+operator() (const GivenFieldParam_Tag& tag, const int& cell) const {
+  for (unsigned int ipt=0; ipt<dim; ++ipt)
+    beta(cell,ipt) = given_field_param(cell,ipt);
+}
+template<typename EvalT, typename Traits, typename EffPressureST, typename VelocityST, typename TemperatureST>
+KOKKOS_INLINE_FUNCTION
+void BasalFrictionCoefficient<EvalT, Traits, EffPressureST, VelocityST, TemperatureST>::
+operator() (const GivenField_Tag& tag, const int& cell) const {
+  for (unsigned int ipt=0; ipt<dim; ++ipt)
+    beta(cell,ipt) = given_field(cell,ipt);
```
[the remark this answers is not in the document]
bartgol: I'm not sure. The scalar types of the two fields are different, so deep_copy might either not compile, or do a memcpy (which would be even worse). I guess we could try and though.

src/LandIce/evaluators/LandIce_GatherVerticallyContractedSolution_Def.hpp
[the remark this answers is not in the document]
mcarlson801: Since this change is okay for now, I'm going to resolve this one.

mperego:
> > @mcarlson801 what's the ETC for this PR? We have enother PR that is ready to be merged in. My understanding is that they are complitely orthogonal. Should we merge the other PR first?
> 
> There's a few minor edits and Max is seeing nondeterministic behavior in a performance test that needs to be tracked down. I don't see an issue with merging the other PR first.

@jewatkins OK! I merged the other PR. Sorry for the nondeterministic behavior.. those are painful to fix.

src/LandIce/evaluators/LandIce_BasalFrictionCoefficient_Def.hpp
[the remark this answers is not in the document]
jewatkins: Good point, I did not see the different scalar type. I'm not sure if sacado has the overloads for this case. I would hope there's something in place to prevent a direct memcpy but maybe there's no such guard.

src/evaluators/interpolation/PHAL_DOFCellToSide.hpp
mcarlson801: I'll add a TODO here just to be safe but yes, you have that right and my plan is to address this during the removal of the non-collapsed layouts.

src/evaluators/utility/PHAL_ComputeBasisFunctionsSide_Def.hpp
mcarlson801: The sideset indices for each side are guaranteed to be disjoint sets so this one is safe.

mperego: Maybe it's worth adding a comment stating that.

src/Albany_StateManager.cpp
mcarlson801: I revisited this section of the code to see what needed to be done here. The else here is covering the case where stateRef.entity is not nodal. There isn't really anything in this if-else that depends on layout size so I'm not sure it actually makes sense to do an exception here based on layout rank beyond checking for the bad case rank == 1 which is no longer relevant for collapsed layouts. 

It seems like to me that there are already exceptions earlier in the function to make sure we aren't getting unexpected input so as far as I can tell nothing needs to be added here. Let me know if I am missing something here since I am not super familiar with the stateRef entity stuff.

src/LandIce/evaluators/LandIce_BasalFrictionCoefficient_Def.hpp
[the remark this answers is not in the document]
mcarlson801: It looks like we will have to mess with this a bit more to get it to compile so I'm marking resolved for now and we can revisit in the future.

mperego:
@mcarlson801 thanks!!

PR #655, opened 2021-01-27
Use PHX::print<Entity>() to get Entity name

mperego:
In parts of the code we use the strings `Node`, `Cell`, etc. for comparisons with the `layout` names. However the `layout` names  are based on the implementation of `typeid` that depends on the compiler. In particular   `PHX::print<Node>()` returns `Node` when compiled with gcc and `4Node` when compiled with intel. In order to fix this, we replace the string `Node` with   `PHX::print<Node>()` throughout the code. We also change `Albany` implementation of `Node::name()` to return  `PHX::print<Node>()`, for consistency. Moreover, following Panzer, we overload `PHX::print<Node>()` to return `Node`, so that the value does not change using different compilers.
Thanks @bartgol for figuring out what the issue was.

### DIFF
--- a/src/evaluators/state/PHAL_LoadSideSetStateField_Def.hpp
+++ b/src/evaluators/state/PHAL_LoadSideSetStateField_Def.hpp
@@ -99,7 +99,7 @@ evaluateFields(typename Traits::EvalData workset)
   int size = dims.size();
   // Check the tag of the first extent after (cell,side,) to determine if field is nodal
   const std::string& leading_field_tag = size>2 ? field.fieldTag().dataLayout().name(2) : "";
-  TEUCHOS_TEST_FOR_EXCEPTION (size>2 && leading_field_tag!="Node" && leading_field_tag!="Dim" && leading_field_tag!="VecDim", std::logic_error,
+  TEUCHOS_TEST_FOR_EXCEPTION (size>2 && leading_field_tag!=PHX::print<Node>() && leading_field_tag!=PHX::print<Dim>() && leading_field_tag!=PHX::print<VecDim>(), std::logic_error,
                               "Error! Invalid field layout in LoadSideSetStateField.\n");
 
   // For collapsed layouts, the tag to check is of the first extent after (side,) to determine if field is nodal
@@ -149,7 +149,7 @@ evaluateFields(typename Traits::EvalData workset)
           break;
 
         case 2:
-          if (leading_field_tag_sideset=="Node")
+          if (leading_field_tag_sideset==PHX::print<Node>())
           {
             // side set node scalar
             for (unsigned int node=0; node<dims[1]; ++node)
@@ -168,7 +168,7 @@ evaluateFields(typename Traits::EvalData workset)
           break;
 
         case 3:
-          if (leading_field_tag_sideset=="Node")
+          if (leading_field_tag_sideset==PHX::print<Node>())
           {
             // side set node vector/gradient
             for (unsigned int node=0; node<dims[1]; ++node)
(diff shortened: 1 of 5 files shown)


src/PHAL_Dimension.hpp
```diff
@@ -85,4 +85,17 @@ namespace PHX {
   template<> struct is_extent<Dummy> : std::true_type {};
 }
 
+namespace PHX {
```
bartgol, line 88: Are these specialization declaration necessary? I thought ETI-ing the fcn in the cpp file was enough (much like we do for the evaluators)...

src/evaluators/state/PHAL_LoadSideSetStateField_Def.hpp
```diff
@@ -99,7 +99,7 @@ evaluateFields(typename Traits::EvalData workset)
   int size = dims.size();
   // Check the tag of the first extent after (cell,side,) to determine if field is nodal
   const std::string& leading_field_tag = size>2 ? field.fieldTag().dataLayout().name(2) : "";
-  TEUCHOS_TEST_FOR_EXCEPTION (size>2 && leading_field_tag!="Node" && leading_field_tag!="Dim" && leading_field_tag!="VecDim", std::logic_error,
+  TEUCHOS_TEST_FOR_EXCEPTION (size>2 && leading_field_tag!=PHX::print<Node>() && leading_field_tag!=PHX::print<Dim>() && leading_field_tag!="PHX::print<VecDim>()", std::logic_error,
```
bartgol, line 102: I think you left some spurious quote marks for VecDim.

bartgol (changes requested):
Looks good. I think there's one typo to fix.

src/PHAL_Dimension.hpp
mperego: I get this error if I don't  (Panzer does the same):
```
/ascldap/users/mperego/Workspace/albany/sources/albany-src/src/PHAL_Dimension.cpp:62:37: error: specialization of ‘std::string PHX::print() [with T = Dim; std::string = std::__cxx11::basic_string<char>]’ after instantiation
   62 |   template<> std::string print<Dim>(){return "Dim";}
```

src/evaluators/state/PHAL_LoadSideSetStateField_Def.hpp
mperego: Good catch!

mperego:
@bartgol I fixed the typo. Let me know if I can merge.

src/PHAL_Dimension.hpp
bartgol: Ah, that's because you should put all the specializations at the top of the cpp file, before they are used in all the `name` methods below.

Btw, when I compile, all is good, but when I open the file in my editor, my background clang compiler issues a warning for all the `name()` implementations:
```
returning address of local temporary object
```

I think the reason is that `PHX::print<T>()` returns a string, which should get destroyed when `name()` returns, and yet, we return the pointer inside the std::string. I am not sure this is wrong, but I am also not sure it's right...

One way could be to do
```
static auto s = PHX::print<Node>();
return s.c_str();
``` 
It's a big cloggy, but definitely safe

bartgol: Ok, I confirmed this is indeed buggy with this:
```
std::string foo () {
  std::string a = "test me";
  return a;
}

const char* bar () {
  static auto s = foo();
  return s.c_str();
}

int main(int ,char** ){
  std::cout << "s: " << bar() << "\n";
  return 0;
}

```
Output with `static` in the impl of `bar`:

```
s: test me
```
Output without `static` in the impl of `bar`:
```
s:
```

mperego:
@bartgol Thanks for findin the bug with the `name()` implementation. I fixed it as you suggested. I've also removed the declaration of the specializations.

bartgol (approved):

bartgol:
Looks good, thanks!

PR #657, opened 2021-01-29
Use Kokkos::sort instead of std::sort

### LINKED ISSUE #656: CUDA build errors in nightly
It looks like the cuda builds are down:
```
/.../build/TrilinosInstall/include/Cuda/Kokkos_Cuda_Parallel.hpp(480): error: calling a __host__ function("std::sort<int *> ") from a __device__ function("Kokkos::Impl::ParallelFor< ::,  ::Kokkos::RangePolicy< ::Kokkos::Cuda > ,  ::Kokkos::Cuda> ::operator () const") is not allowed
/.../build/TrilinosInstall/include/Cuda/Kokkos_Cuda_Parallel.hpp(480): error: identifier "std::sort<int *> " is undefined in device code
2 errors detected in the compilation of "/tmp/tmpxft_0000ef94_00000000-6_Albany_Hessian.cpp1.ii".
```
@mperego @kliegeois 
It looks like a std::sort is used which can't be used on device. If you want that bit of code on device, I think you can try using Kokkos::sort.

kliegeois:
This PR should fix #656.

I tested it on Blake, the tests passed.
However, I did not test it on a CUDA build.

@jewatkins

### DIFF
--- a/src/utility/Albany_Hessian.cpp
+++ b/src/utility/Albany_Hessian.cpp
@@ -3,20 +3,11 @@
 #include "Albany_TpetraTypes.hpp"
 #include "Albany_StateInfoStruct.hpp"
 #include "Albany_Hessian.hpp"
+#include "Albany_KokkosUtils.hpp"
 
 #include <Kokkos_UnorderedMap.hpp>
 #include <Kokkos_Sort.hpp>
 #include <math.h>
-
-int min(int a, int b)
-{
-    return (a <= b) ? a : b;
-}
-
-int max(int a, int b)
-{
-    return (a >= b) ? a : b;
-}
 
 Teuchos::RCP<Tpetra_CrsGraph> Albany::createHessianCrsGraph(
     Teuchos::RCP<const Tpetra_Map> p_owned_map,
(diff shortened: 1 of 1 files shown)


jewatkins:
I'm getting the following error on weaver:
```
/home/projects/albany/nightlyCDashWeaver/build/TrilinosInstall/include/Cuda/Kokkos_Cuda_Parallel.hpp(480): error: calling a __host__ function("Kokkos::sort< ::Kokkos::View<int *,  ::Kokkos::LayoutLeft,  ::Kokkos::Device< ::Kokkos::Cuda,  ::Kokkos::CudaUVMSpace> ,  ::Kokkos::MemoryTraits<(unsigned int)0u>  > > ") from a __device__ function("Kokkos::Impl::ParallelFor< ::,  ::Kokkos::RangePolicy< ::Kokkos::Cuda > ,  ::Kokkos::Cuda> ::operator () const") is not allowed

/home/projects/albany/nightlyCDashWeaver/build/TrilinosInstall/include/Cuda/Kokkos_Cuda_Parallel.hpp(480): error: identifier "Kokkos::sort< ::Kokkos::View<int *,  ::Kokkos::LayoutLeft,  ::Kokkos::Device< ::Kokkos::Cuda,  ::Kokkos::CudaUVMSpace> ,  ::Kokkos::MemoryTraits<(unsigned int)0u>  > > " is undefined in device code
```
I think Kokkos::sort is only used as a threaded sort outside of a parallel_for. I don't think kokkos currently has a device sort. See:
https://github.com/kokkos/kokkos/issues/645 for options

An easy fix might be to not thread on rows and do a threaded sort for each row sequentially - i.e. get rid of the parallel_for.

kliegeois:
@jewatkins thanks for having tested it with CUDA!

I fixed it as suggested: using a threaded sort called sequentially.
It is the easiest fix and allows the CUDA tests to pass correctly as soon as possible.
If needed, I can revisit this to have a more scalable strategy.

jewatkins:
Looks like there's one more lingering error:
```
/ascldap/users/jwatkin/LandIce/Albany-test/src/utility/Albany_Hessian.cpp(11): error: function "min(int, int)" has already been defined
/ascldap/users/jwatkin/LandIce/Albany-test/src/utility/Albany_Hessian.cpp(16): error: function "max(int, int)" has already been defined
```
I think you can safely get rid of those functions and use the ones provided in `src/utility/Albany_KokkosUtils.hpp` using the `KU` namespace. Otherwise you could rename them and add `KOKKOS_INLINE_FUNCTION` to make them device capable (I think they currently conflict with the cuda lib implementations).

kliegeois:
Thanks again @jewatkins !
I just updated the file as suggested and verified that it compiled and that the tests passed on Blake.

jewatkins:
This is working now thanks!

Issue #663, opened 2021-02-17
 SteadyHeatConstrainedOpt2D_Scalar_And_Dist_Param_Tpetra failing on Blake in GCC Warning-free build


ikalash:
With this nightly test harness being back up, it appears the  SteadyHeatConstrainedOpt2D_Scalar_And_Dist_Param_Tpetra  test is failing in the Blake GCC warning-free build: 

https://sems-cdash-son.sandia.gov/cdash/testDetails.php?test=878186&build=12561 

The error is a bit cryptic:

```
 		total iterations: 39
  
resp<Tangent>: 0.221462 [ 0 0 0 ]
resp<Tangent>: 0.15 [ 0 0 0 ]
resp<DistParamDeriv>: 0.221462 [ 0 0 0 0 ]
resp<DistParamDeriv>: 0.15 [ 0.075 0.075 0.075 0.075 ]

p=0: *** Caught an unknown exception

p=1: *** Caught an unknown exception

p=3: *** Caught an unknown exception

p=2: *** Caught an unknown exception
```

@mperego, @kliegeois : any idea on what might be going on?  It could conceivably be a fluke, in which case we can give it another day and see what happens tomorrow.

mperego:
@ikalash I'm not sure. @kliegeois found out that it failed when Sacado is built with debug checks. Anyway, @kliegeois is going to modify the implementation using `SharedParameter`,  so instead of looking into fixing this, I think it's best to wait for the new implementation that should happen next week at the latest.

kliegeois:
@ikalash this looks very similar to the issues that I faced. As proposed by @mperego , it is better to wait for the use of SharedParameter as this might solve the issue.

ikalash:
Thanks @kliegeois .  Do you know what is the approximate timeline for the SharedParameter fix?

kliegeois:
@ikalash, I do not know yet. I am on it but it requires some refactoring. Hopefully, I should be able to estimate the timeline soon.

kliegeois:
@ikalash the issue is resolved on my branch.

Before creating the PR (which includes the Hess-vec computation with scalar parameters), I still need a bit of time to: 
- fix 3 tests that failed,
- build with warnings as errors,
- finalize the dox.

The PR should be created early next week at the latest.

kliegeois:
@ikalash quick update: the PR should be ready on Monday!

ikalash:
Great thanks @kliegeois !  I will keep an eye out for it next week.

[...]

PR #664, opened 2021-02-23
Hessian-vector product related to 

kliegeois:
This PR covers several improvements:

1. The Hessian-vector product computations related to scalar parameters are added both for responses and residuals,
2. Solves #663,
3. Solves a bug related to the zero-out of the Lagrange multipliers associated to Dirichlet boundary conditions when using more than 1 MPI process; this is now tested in PyAlbany using the steadyHeat.py python test,
4. An unused option of the steadyHeat.py python test has been removed,
5. It is now possible to specify the AD support for the Hessian-vector product differently for each contribution in the yaml file; some conditions are required to have meaningful computations, if one of those is not verified, an error is thrown explaining what should be updated in the .yaml file.

This PR has been tested on Blake with and without `MESH_DEPENDS_ON_PARAMETERS` enabled (with [PR #8806](https://github.com/trilinos/Trilinos/pull/8806)).
The code builds _warning free_, i.e., no errors are thrown while building with warning treated as errors with the previously defined considered set of warnings.

@mperego @ikalash


src/evaluators/utility/PHAL_Source_Def.hpp
```diff
@@ -17,6 +17,9 @@
 #include "Teuchos_Array.hpp"
 #include "Teuchos_TestForException.hpp"
 
+#include "PHAL_SharedParameter.hpp"
+#include "PHAL_Spatial.hpp"
```
mperego, line 21: The name `PHAL_Spatial` is a bit obsucre. We could call it `PHAL_Field_Source` or `Source Functions`

mperego (approved):
@kliegeois  A lot of work!
I didn't really try to understand every single line of code, but overall looks great.
I added a couple of comments. They are just suggestions, no need to modify them.

PR #666, opened 2021-02-26
Solve a bug which prevents from computing the Hess-vec of the residual w.r.t extruded parameters with MPI

kliegeois:
This PR solves a bug which prevented from computing the Hessian-vector product of the residual w.r.t extruded parameters with MPI.

@mperego


mperego (approved):
Looks good. Thanks for the fix @kliegeois !

PR #667, opened 2021-03-04
Remove Use Old Reduced Space Interface and add observeResponse

kliegeois:
This PR removes the deprecated `Use Old Reduced Space Interface` option and allows to print the response values during the optimization solve.
In particular, `CumulativeScalarResponseFunction` responses are printed such that both the value of the sum and the value of all the terms are printed.

The tests passed on Blake and the code built warning free.

@mperego


mperego (approved):
Looks good! Thanks.
I've made a few changes to
1. Change Response output message from "Iteration" to "Optimization Iteration" to
   make it easy to search in output file
2. Fix response for iteration 0

PR #672, opened 2021-03-15
PyAlbany: add performAnalysis and setParameter functions

kliegeois:
This PR updates PyAlbany to add the following capabilities:

1. Add the possibility to create a PyAlbany problem using a `Teuchos::ParameterList` illustrated as follows (and as discussed with @bartgol ):
```
        filename = "input_dirichlet_mixed_paramsT.yaml"
        parameter = Utils.createParameterList(
            filename, parallelEnv
        )

        parameter.sublist("Discretization").set("1D Elements", 10)
        parameter.sublist("Discretization").set("2D Elements", 10)

        problem = Utils.createAlbanyProblem(parameter, parallelEnv)
```

2. Add the possibility to solve the inverse problem as follows:
```
        problem.performAnalysis()
```
It is then possible to access the parameter values as follows:
```
        para_0 = problem.getParameter(0)
```
Or even to compute sensitivity values and reduced Hessian evaluated at the last point of the optimization problem using:
```
        problem.performSolve()
```
3. Add the possibility to specify the values of the parameter where the response, the sensitivity values, and the reduced Hessian vector-product must be evaluated:
```
        problem.setParameter(0, para_0_new)
```
This allows to loop over samples and evaluate the model without creating a new one as follows:
```
        for i in range(0, n_values):
            para_0_new[:] = para_0_values[i]
            problem.setParameter(0, para_0_new)

            problem.performSolve()

            response = problem.getResponse(0)
            responses[i] = response.getData()[0]
```
There is, however, one known limitation, if the inverse problem of an instance of a PyAlbany problem has been solved,  we cannot manually change the values of the parameter. If needed, it is required to create a new instance of a PyAlbany problem and set the parameter for this second instance. If a user tries to set the parameter values after the solve of the inverse problem, an error is thrown.
 
All those capabilities are tested using ctest.

Finally, a Monte Carlo example has been added in PyAlbany/examples to illustrate the capability of looping over samples.
This example draws 200 samples of a uniform distribution, evaluates the forward problem for every of those samples, and draws a histogram of the random parameter, plots the samples and their corresponding quantity of interest value, and a histogram of the quantity of interest:

UQ

@mperego @ikalash

### DIFF
--- a/PyAlbany/examples/UQ/.gitignore
+++ b/PyAlbany/examples/UQ/.gitignore
@@ -1,0 +1,2 @@
+steady2d*
+*.jpeg
--- a/PyAlbany/examples/UQ/MC_example.py
+++ b/PyAlbany/examples/UQ/MC_example.py
@@ -1,0 +1,73 @@
+from PyTrilinos import Tpetra
+from PyTrilinos import Teuchos
+
+from mpi4py import MPI
+import numpy as np
+from PyAlbany import Utils
+import os
+import sys
+
+try:
+    import matplotlib as mpl
+    mpl.use('Agg')
+    import matplotlib.pyplot as plt
+    printPlot = True
+except:
+    printPlot = False
+
+def main(parallelEnv):
+    comm = MPI.COMM_WORLD
+    myGlobalRank = comm.rank
+
+    # Create an Albany problem:
+    filename = "input_dirichletT.yaml"
+    parameter = Utils.createParameterList(
+        filename, parallelEnv
+    )
+
+    problem = Utils.createAlbanyProblem(parameter, parallelEnv)
+
+    parameter_map_0 = problem.getParameterMap(0)
+    parameter_0 = Tpetra.Vector(parameter_map_0, dtype="d")
+
+    N = 200
+    p_min = -2.
+    p_max = 2.
+
+    # Generate N samples randomly chosen in [p_min, p_max]:
+    p = np.random.uniform(p_min, p_max, N)
+    QoI = np.zeros((N,))
+
+    # Loop over the N samples and evaluate the quantity of interest:
+    for i in range(0, N):
+        parameter_0[0] = p[i]
+        problem.setParameter(0, parameter_0)
+
+        problem.performSolve()
+
+        response = problem.getResponse(0)
+        QoI[i] = response.getData()[0]
+
+    if myGlobalRank == 0:
+        if printPlot:
+            f, (ax1, ax2, ax3) = plt.subplots(1, 3, figsize=(16,6))
+
+            ax1.hist(p)
+            ax1.set_ylabel('Counts')
+            ax1.set_xlabel('Random parameter')
+
+            ax2.scatter(p, QoI)
+            ax2.set_ylabel('Quantity of interest')
+            ax2.set_xlabel('Random parameter')
+
+            ax3.hist(QoI)
+            ax3.set_ylabel('Counts')
+            ax3.set_xlabel('Quantity of interest')
+
+            plt.savefig('UQ.jpeg', dpi=800)
+            plt.close()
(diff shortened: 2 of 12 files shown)


kliegeois:
This PR requires [PR #8885](https://github.com/trilinos/Trilinos/pull/8885) to work.

bartgol:
> There is, however, one known limitation, if the inverse problem of an instance of a PyAlbany problem has been solved, we cannot manually change the values of the parameter. If needed, it is required to create a new instance of a PyAlbany problem and set the parameter for this second instance. If a user tries to set the parameter values after the solve of the inverse problem, an error is thrown

I'm curious, what's the reason for this? I mean, which class throws an error?

Anyhow, this is not a "big" limitation, since probably one is not interested in testing the model at different data points and doing an optimization problem in the same run. Usually, you do sampling when you don't want to do an optimization (or you cannot). So probably this won't be a big deal.

kliegeois:
> > There is, however, one known limitation, if the inverse problem of an instance of a PyAlbany problem has been solved, we cannot manually change the values of the parameter. If needed, it is required to create a new instance of a PyAlbany problem and set the parameter for this second instance. If a user tries to set the parameter values after the solve of the inverse problem, an error is thrown
> 
> 
> 
> I'm curious, what's the reason for this? I mean, which class throws an error?
> 
> 
> 
> Anyhow, this is not a "big" limitation, since probably one is not interested in testing the model at different data points and doing an optimization problem in the same run. Usually, you do sampling when you don't want to do an optimization (or you cannot). So probably this won't be a big deal.

The error is thrown deliberately in the Python interface. I did that because it appears that some computations of the PeformSolve are reused from the PerformAnalysis. Those computations are correct only if the parameters have not changed. As you said, this basically makes less sense to change the parameter values after a PeformAnalysis. This is why I did not investigate that too much and decided to throw an error. If needed, I can investigate that more.

bartgol:
> The error is thrown deliberately in the Python interface. I did that because it appears that some computations of the PeformSolve are reused from the PerformAnalysis.

I wonder if [this](https://github.com/trilinos/Trilinos/blob/1a61be2dc0378de70f515792f215889d6f449004/packages/piro/src/Piro_NOXSolver_Def.hpp#L150) is why. We might have to purge/update the parameter list, as done [here](https://github.com/trilinos/Trilinos/blob/1a61be2dc0378de70f515792f215889d6f449004/packages/piro/src/Piro_NOXSolver_Def.hpp#L135).

kliegeois:
> > The error is thrown deliberately in the Python interface. I did that because it appears that some computations of the PeformSolve are reused from the PerformAnalysis.
> 
> 
> 
> I wonder if [this](https://github.com/trilinos/Trilinos/blob/1a61be2dc0378de70f515792f215889d6f449004/packages/piro/src/Piro_NOXSolver_Def.hpp#L150) is why. We might have to purge/update the parameter list, as done [here](https://github.com/trilinos/Trilinos/blob/1a61be2dc0378de70f515792f215889d6f449004/packages/piro/src/Piro_NOXSolver_Def.hpp#L135).
> 
> 

It could be the case, I will try that. Thanks for the advice!

mperego (approved):
Looks good @kliegeois. Thanks for adding this. The Montecarlo problem is cool! I'm sure we'll make good use of this at some point.

kliegeois:
Thanks @mperego for your review!

@bartgol, after investigating what you suggested, I added a second commit which sets the `"Compute State"` to true when setting the parameter value. I had to do so using `RCP<Teuchos::ParameterList>` instead of `Teuchos::ParameterList` to propagate the change up to the`Piro::NOXSolver`.
I updated the `PyAlbany_Inv_SteadyHeat` test to test the capability. There is no known limitation anymore.

PR #673, opened 2021-03-18
Block discretization: computation of the graph of the blocks

kliegeois:
This PR is a work in progress.

This PR is intended to introduce the block discretization with more than one block.

List of capabilities already available in the branch:
- Possibility to create a `BlockedSTKDiscretization` with more than one `AbstractSTKMeshStruct`,
- Possibility to create the graph of the diagonal blocks using the underlying `STKDiscretization` objects,
- Possibility to access the FAD length and offset of the blocks using `getBlockFADLength` and `getBlockFADOffset` respectively.

Those capabilities are tested in `Discretization_Serial_Unit_Test`.

List of missing capabilities:
- Computation of the off diagonal graph,
- Creation of the transposed operator associated to off diagonal matrices,
- Reforming of the option to specify the block structure (@mperego do you have a preference regarding the new format?).

The tests passed correctly on Blake.

@bartgol @mperego


bartgol:
For block structure specs format, my preference are in order:

1) array with brackets-delimited blocks, like `blocks: [ velocity, temperature]`, and, if we foresee nesting, `blocks: [ velocity, [N, h]]`.
2) sublists:
```
 Blocks:
    Number: 2
       Block 0: velocity
       Block 1: temperature
```
and, if we foresee nesting,
```
  Blocks:
    Number: 2
       Block 0:
          Type: simple
           Field: velocity
        Block 2:
           Type: blocked
            Number 2:
               Block 0:
                  Type: simple
                  Field: N
               Block 1:
                   Type: simple
                    Field: h
```

I think the brackets format is more compact, without giving up readability (especially for nested blocks). However, the sublists one allows to specify other specs in the sublist, like the FE type. The brackets approach would require more syntax (or more lists) for FE type:
```
  blocks: [ velocity, temperature, [N, h] ]
  FE types: [ P2, P1, [P1, P0] ]
```
or
```
blocks: [ velocity:P2, temperature:P1, [N:P1, h:P0] }
``` 
I think separate `blocks` and `FE types` lists (whose structure must be matched) is more readable than the single one.

Another advantage of the sublist approach is that we can simply grab a block sublist, and forward it to the dof mgr, or to the discretization factory.

ikalash:
@bartgol : I have a few questions about the proposed syntax.  What would the subnesting mean for the blocks?  In your example, what would '[N, h]' mean?  This is a single block for two sets of equations, one for `N' and one for `h'?  Regarding the FE types: it seems to me what you are proposing might not be consistent with the way the discretization works now, potentially.  If you are reading in an input mesh, it will have a finite element type in it already, and the discretization class will infer the element order from that.  In this case, specifying the FE type will not make sense I think.  I could see other instances where it might make sense (with the mesh extrusion or if we're generating the mesh inside the code, though that currently doesn't support any elements other than linear quads/hexes).

bartgol:
@ikalash 

The notation `[u, [v, w]]` would mean that the solution would have 2 blocks, and the second block would be itself a blocked vector (or a Thyra::ProductVector, I should say). It might help if we get into the case where there are multiple physics coupled together. E.g., say you couple Navier-Stokes with some other physics (Temperature or maybe Structural solver). On one level, you want to differentiate NS from the other physics, but you might also want to see the NS pb as a block problem. So an FSI block structure might be `[ [u,p], d]', with `d` displacement. This would make it easier to cook up a blocked preconditioner for the FSI pb, whose velocity block is itself a block preconditioner.

As for the FE, sure, the mesh will have an associated FE order. But we probably will have more blocks than meshes/discretizations, so we need to find a way to tell which discretization each block belongs to. You are right, though, saying that what I proposed won't work, cause a) it doesn't say on which mesh each DOF lives, and b) there might be two meshes, both with P1. Maybe we should break the concept of mesh from the concept of discretization (which I see as a mesh+FE_Type pair), and do something like this:

```
Meshes:
  Num Meshes: 2
  Mesh 0:
    Name: fluid
    Type: Gmsh
     ...
  Mesh 1:
     Name: structure
     Type: Ioss
     ...
```
Or assume that there's always one (and only one) volume mesh (ruling out stuff like schwarz or 3D-3D FSI kind of setup):
```
Mesh:
  Name: ice_mesh
  Type: Extruded
  Side Meshes:
     Sidesets: [basal,upper]
     basal:
        Name: hydro_mesh
        Type: Ioss
     upper:
        Name: surf_mesh
        Type: SidesetSTK
```
Then create blocked discretization:
```
Discretization:
   Num Blocks: 3
   Block 0:
      Name: vol_P1
      Mesh: ice_mesh
      FE Type:  P1
    Block 1:
      Name:  basal_P1
      Mesh: hydro_mesh
      FE Type: P1
    Block 2:
       Name: basal_P0
       Mesh: hydro_mesh
       FE Type: P0
```
Finally:
```
Solution:
  blocks names: [ velocity, [N,h] ]
  blocks discretizations: [ vol_P1, [basal_P1, basal_P0] ]
...

**Comment by bartgol** [comment]
I do believe that we ought to split our current concept of discretization into two: geometry and FE. That's because we might have different blocks (e.g., different orders) defined on the same mesh.

**Comment by ikalash** [comment]
@bartgol : I like your proposed solution with the blocked mesh and sub-mesh entries there.  This is similar to how Teko sub-block preconditioners are specified.  

I see what you are saying about splitting out the discretization/finite element type, but I worry that this may cause some confusion...  How does a user know what is a valid discretization, i.e. that vol_P1 is valid in block 0 in your example?  If the mesh fed in is something different like a biquadratic hex mesh, that element won't make sense - would the code throw an error?  Maybe we can keep this design in mind for the future but hold off on putting it in for now, since at this point it's not possible to have different order elements in different blocks on the same mesh (e.g. P1/P0 elements for velocity/pressure)?  Once we have the capability, we can add it.

Basically, my two cents is I think we should be a little bit careful to not make this unnecessarily complicated, like the SS mesh syntax (it's really complicated IMO though the complexity may be unavoidable in that case).

**Comment by kliegeois** [comment]
@bartgol @mperego @ikalash 

I have a somehow related comment/question.

Currently, in the `BlockedSTKDiscretization` class, we are using:
`Teuchos::Array<Teuchos::RCP<disc_type>> m_blocks;` where `disc_type` is `STKDiscretization`.
I assumed that the choice of using a `Teuchos::Array` was motivated by the capability to be able to store a different discretization for the diagonal blocks (I might be wrong as the class was limited to be used with one block first).

From that, I _temporarily_ added a constructor which takes as an argument `Teuchos::Array<Teuchos::RCP<AbstractSTKMeshStruct>> &stkMeshStruct,` where one STK mesh structure is specified for every block diagonal.

I think that, instead, we should have only one STK mesh structure set as input and that, we could then, during the construction of `BlockedSTKDiscretization`, create a new `AbstractSTKMeshStruct` per block as follows:
1. Use the  input STK mesh structure and create a new one which only includes the `stk::mesh::Part` related to the current block (it seems that it can be done using `stk::mesh::Selector` ),
2. Set the `neq` of the `AbstractSTKMeshStruct` of the current block.

This requires a `IossSTKMeshStruct` constructor, among others, which takes as an argument an existing `IossSTKMeshStruct` and a `stk::mesh::PartVector` which specifies the list of parts related to the current block.
Obviously, this will require to define `stk::mesh::Part` parts of the used mesh and we will need, in the parameter list, to specify on which mesh part an equation is defined.

Does it make sense?

**Comment by bartgol** [comment]
@ikalash What is preventing us from separating the mesh order from the FE one? I can build a P2  FE on a linear mesh (and viceversa, which would implicitly discard the edges/faces curvature). Looking at albany, it all boils down to the element coordinates vector (which later is used in ComputeBasisFunctions). Currently, we store it in the discretization as `coords[num_ws][num_cells][num_nodes][3]`. We _do_ fill this by grabbing the stk `coordinates_field`, which would have the same layout as what the mesh element is (so, say, a TET_10 would have 10 nodes), so we'd have to do a check on that (but it would be a check only at the disc construction time, so not too bad). We could do it by hand (check the stk topo and compare with the FE Type, to see if we need to interpolate or slice (recall that stk stores vertices before edges, before faces)). But perhaps there's some tool in Intrepit that can help to do this. @mperego is [this stuff](https://github.com/trilinos/Trilinos/blob/master/packages/intrepid2/src/Projection/Intrepid2_ProjectionTools.hpp) something that could help here?

This would take care of the 'online' phase only, though. There is the issue of saving the results to exodus. If the stk/exodus metadata/bulkdata store, say, P2 elements, we cannot save only the P1 nodes and leave the P2 nodes to 0. The correct value of the P2 nodes should be interpolated from the P1 ones.

So the easiest strategy would be to have 1 stk mesh per FE Type, but that seems silly. And I can't believe that stk folks are forcing CFD apps (which are very common) to save u and p on different meshes. There's got to be a way to save a P1 field on a P2 mesh without interpolating it up. Unless the stk devs assume the user will always set the output mesh to be linear (the P1 part of a P2 field comes first).

Long story short:

- I think we can make the online calculation work for multiple FE types on the same mesh. We "just" need to pay attention when grabbing the coord field from the mesh (and possibly also the stored states?).
- It seems odd that one cannot store a P1 and P2 field on the same exodus mesh. Perhaps we ought to always use a linear mesh as output, and drop all the solution dofs at the quadratic nodes? Probably ok for plotting, but not great for restarts (lowers accuracy).
- If we lock FE type with stk mesh, we are making it impossible for a user to take a P1 mesh with IC to start/restart a simulation with P2 elements.

**Comment by bartgol** [comment]
@kliegeois I think I like your idea. It would allow to specify a discretization with FE order and mesh part name. 

I see two limitations, though:
- It limits ourselves to the case where we only have 1 "volumetric" mesh. Arguably, that will always be enough for ALI, but we need to be ok with hard coding this limitation deep into our discretization stack.
- Right now, it is possible to load a 3d mesh and the corresponding 2d basal mesh (from a previous simulation). It is natural to use the 2d mesh to store basal quantities, rather than big 3d fields in the 3d mesh, which will be mostly filled with 0's. If we follow your approach, however, the basal mesh would be rebuilt from the 3d one, while I'd like to load the existing 2d one (which has all the data, btw).

Maybe we could use a hybrid approach. What you suggest is not "that much" different from how a SideSetSTKMeshStruct is currently built (usually for the upper side): specify a side set, and create a stk mesh with the entities on that side set (which is just one particular type of Part). Perhaps we could:

- allow the user to specify side meshes (e.g., if they intend to load them); this way, one can create an extruded mesh from a loaded 2d one, or even load 3d and 2d meshes (they must "match", of course).
- when parsing the discretizations (FE,part name) pairs, if a stk mesh is present for that part, use it, otherwise proceed to build it.

Imho, for most ALI problems, the user will want to pass in the basal mesh. But having the ability to build a stk mesh on the fly without having the user write out all its specs on the input file (as it's done now for SideSetSTK) might be useful. The tricky part is to make sure that the MeshSpecs for all the meshes built from Part's are built before the Problem (which needs them). I guess we can do that by parsing the blocks discretization types before calling the problem ctor.

**Comment by kliegeois** [comment]
@bartgol I am not sure about the limitation of having only one volume.
I am not a STK expert at all but it seems that you can have as many `stk::mesh::Part` as you want (those can even have overlaps).

Therefore, I think that we can define as many volumes as we want (with different `stk::mesh::Part`). The mapping between `stk::mesh::Part` and blocks can be treated using `panzer::BlockedDOFManager` and `Albany::STKConnManager` (a child class of `panzer::ConnManager`).

Concerning the second limitation, we could consider having the possibility to load the 3D mesh and 0,1, or more 2D mesh(es) as you suggested in your comment:

> - allow the user to specify side meshes (e.g., if they intend to load them); this way, one can create an extruded mesh from a loaded 2d one, or even load 3d and 2d meshes (they must "match", of course).

I still try to understand STK better and how to create correctly the `AbstractSTKMeshStruct ` based on a `Selector`.

**Comment by bartgol** [comment]
> I still try to understand STK better and how to create correctly the `AbstractSTKMeshStruct ` based on a `Selector`.

I think SideSetSTKMeshStruct can probably be rebranded as `STKPartMeshStruct` or something.  After all, in the impl of SideSetSTKMeshStruct there's no place where we _explicitly_ use the fact that it's a side. I mean, there are some, like when we loop on the parent mesh's sides and add them to this mesh with `declare_element`, but we can use 'entity-rank-agnostics' methods (like `declare_entity`), and treat each part agnostically.

**Comment by mperego** [comment]
@bartgol I think it is possible to store multiple fields in the STK mesh. I think they do that in Drekar/Empire, but I'll double check on that.  Yes Intrepid2 (local) projections can be used to map fields into a particular fiinte element space, with some  limitations. E.g. you cannot project a dscontinuous FE into a Hgrad FE. If you want to do that you can use global L2 projections.

**Comment by bartgol** [comment]
> @bartgol I am not sure about the limitation of having only one volume.
> I am not a STK expert at all but it seems that you can have as many `stk::mesh::Part` as you want (those can even have overlaps).

> Therefore, I think that we can define as many volumes as we want (with different `stk::mesh::Part`). The mapping between `stk::mesh::Part` and blocks can be treated using `panzer::BlockedDOFManager` and `Albany::STKConnManager` (a child class of `panzer::ConnManager`).

Yes, we can define as many volumes as we want, provided they all come from the same, single, parent mesh. What I meant is that we cannot do something like an overlapping Schwarz method where the two meshes have different elements. That kind of stuff.

TBC, I am fine with this limitation. I just want us to be aware of what roads we're precluding ourselves.

**Comment by kliegeois** [comment]
> > I still try to understand STK better and how to create correctly the `AbstractSTKMeshStruct ` based on a `Selector`.
> 
> I think SideSetSTKMeshStruct can probably be rebranded as `STKPartMeshStruct` or something. After all, in the impl of SideSetSTKMeshStruct there's no place where we _explicitly_ use the fact that it's a side. I mean, there are some, like when we loop on the parent mesh's sides and add them to this mesh with `declare_element`, but we can use 'entity-rank-agnostics' methods (like `declare_entity`), and treat each part agnostically.

Just to be sure, you suggest that we replace `SideSetSTKMeshStruct` by `STKPartMeshStruct` such that we can use `STKPartMeshStruct` on any defined `stk::mesh::Part` which can be a part define on a volume or surface (or even edges).
We would have a 3D parent mesh (and possibly some 2D parent meshes) and all equations have their own `STKPartMeshStruct` (even if the equation is in volume).

I like this possibility. I can move forward on that direction if we are fine with the known limitation such as:

> In other words, we can only load a single volume mesh, and all the volume meshes must be created from parts of that.

**Comment by bartgol** [comment]
> We would have a 3D parent mesh (and possibly some 2D parent meshes) and all equations have their own `STKPartMeshStruct` (even if the equation is in volume).

Well, two equations might still have the same mesh, assuming they are defined on the same Part and have the same FE Type. We should check if their mesh already exists, before creating a whole new one...

**Comment by bartgol** [comment]
> I like this possibility. I can move forward on that direction if we are fine with the known limitation such as:
> 
> > In other words, we can only load a single volume mesh, and all the volume meshes must be created from parts of that.

Perhaps we can punt on the STKPartMeshStruct for now. It's a nice generalization, but I think we can do away with it for now, as we work on the actual block structure.

**Comment by bartgol** [comment]
> @bartgol I think it is possible to store fields w/ different discretizations in the STK mesh. I think they do that in Drekar/Empire, but I'll double check on that. Yes Intrepid2 (local) projections can be used to map fields into a particular fiinte element space, with some limitations. E.g. you cannot project a dscontinuous FE into a HGrad FE. If you want to do that you can use global L2 projections.

That's ok. Also, I'm not 100% sure we want to map the coordinates_field to whatever FE type the current block has. I _think_ the coordinate fields always need to be a lagrangian field. How does Intrepid use coordinates to build the FE basis info? I'm assuming it always assumes the coord field to be a lagrangian field (e.g., a P1 or P2 one)?

**Comment by kliegeois** [comment]
> > We would have a 3D parent mesh (and possibly some 2D parent meshes) and all equations have their own `STKPartMeshStruct` (even if the equation is in volume).
> 
> Well, two equations might still have the same mesh, assuming they are defined on the same Part and have the same FE Type. We should check if their mesh already exists, before creating a whole new one...

Correct, sorry, I misspoke. 
I meant:
- 1 `STKDiscretization` per equation which has a RCP to a `STKPartMeshStruct` (possibly shared with another equation),
- 1 `STKPartMeshStruct` per part(or group of parts)/FE type.

**Comment by ikalash** [comment]
> @ikalash What is preventing us from separating the mesh order from the FE one? I can build a P2 FE on a linear mesh (and viceversa, which would implicitly discard the edges/faces curvature). Looking at albany, it all boils down to the element coordinates vector (which later is used in ComputeBasisFunctions). Currently, we store it in the discretization as `coords[num_ws][num_cells][num_nodes][3]`. We _do_ fill this by grabbing the stk `coordinates_field`, which would have the same layout as what the mesh element is (so, say, a TET_10 would have 10 nodes), so we'd have to do a check on that (but it would be a check only at the disc construction time, so not too bad). We could do it by hand (check the stk topo and compare with the FE Type, to see if we need to interpolate or slice (recall that stk stores vertices before edges, before faces)). But perhaps there's some tool in Intrepit that can help to do this. @mperego is [this stuff](https://github.com/trilinos/Trilinos/blob/master/packages/intrepid2/src/Projection/Intrepid2_ProjectionTools.hpp) something that could help here?
> 
> This would take care of the 'online' phase only, though. There is the issue of saving the results to exodus. If the stk/exodus metadata/bulkdata store, say, P2 elements, we cannot save only the P1 nodes and leave the P2 nodes to 0. The correct value of the P2 nodes should be interpolated from the P1 ones.
> 
> So the easiest strategy would be to have 1 stk mesh per FE Type, but that seems silly. And I can't believe that stk folks are forcing CFD apps (which are very common) to save u and p on different meshes. There's got to be a way to save a P1 field on a P2 mesh without interpolating it up. Unless the stk devs assume the user will always set the output mesh to be linear (the P1 part of a P2 field comes first).
> 
> Long story short:
> 
> * I think we can make the online calculation work for multiple FE types on the same mesh. We "just" need to pay attention when grabbing the coord field from the mesh (and possibly also the stored states?).
> * It seems odd that one cannot store a P1 and P2 field on the same exodus mesh. Perhaps we ought to always use a linear mesh as output, and drop all the solution dofs at the quadratic nodes? Probably ok for plotting, but not great for restarts (lowers accuracy).
> * If we lock FE type with stk mesh, we are making it impossible for a user to take a P1 mesh with IC to start/restart a simulation with P2 elements.ds

I think there is nothing stopping us, we just don't have that capability now.  I do see the output as a problem, as I don't think you can store e.g. P1 and P2 element solutions on the same (P2) mesh.  One would have to figure out how to handle this - perhaps create a separate output file for the P1 mesh field?

**Comment by bartgol** [comment]
@mperego seems convinced we can store P1 and P2 on the same mesh. That would be obviously optimal. If not possible, we'd have to choose whether we want to have separate exo files for different FE types, or if we prefer to interpolate everything onto one output mesh. Mauro, let us know what the Drekar/Empire folks tell you.

Assuming it's not possible (otherwise we're good), I'd go for interpolating one field onto the other mesh, for ease of visualization. But again, I don't foresee this happening soon, except maybe in the hydrology: one could choose `h` (the water thickness) to be P0, and the integrals in the weak formulation would be well defined (as for the weak pb to be well posed, that is an open question though).

**Comment by mperego** [comment]
> @mperego seems convinced we can store P1 and P2 on the same mesh. That would be obviously optimal. If not possible, we'd have to choose whether we want to have separate exo files for different FE types, or if we prefer to interpolate everything onto one output mesh. Mauro, let us know what the Drekar/Empire folks tell you.
> 
> Assuming it's not possible (otherwise we're good), I'd go for interpolating one field onto the other mesh, for ease of visualization. But again, I don't foresee this happening soon, except maybe in the hydrology: one could choose `h` (the water thickness) to be P0, and the integrals in the weak formulation would be well defined (as for the weak pb to be well posed, that is an open question though).

Below is the answer from Roger. It's probably better for the moment to assume we can store mixed finite elements but without excercising that capability, until we need to do so. By that time things might be a bit better:
_"You can store quadratic nodal elements but nothing higher as far as I know. It has to be in the topology library to be stored in stk. They don’t use intrepid basis. But a few years ago they dropped shards topology for their own implementation. So far they have kept ordering consistent. I believe there is support for lowest order curl and div bases in the exodus format – it is used by alegra. We just recently added support through the entire panzer stack but it is still experimental. We are trying to get the paraview team to visualize lowest order curl and div before extending to higher order. So you should be able to store mixed order, but it may have many bugs. In empire we are just dumping the cell centered values for vis to exodus. For restarts, we have our own format for kokkos views that avoids exodus all together."_

**Comment by bartgol** [comment]
@mperego So basicallly what they are doing is "interpolate the given FE to P0, then save to exodus", right? We could do something like that for now: assume the mesh is P1 (in fact, we could enforce this), then if the DOF is not P1 or P0, interpolate before saving. I believe mixing P0 and P1 is fine (exodus can handle cell and node values).

 I think in the short term (at least), P0 and P1 are all we might be dealing with, so this might be an acceptable compromise. As stk/exodus handling of mixed FE types improves (thank you, Panzer, for being our guinea pig), we can remove the dof interpolation.

**Comment by kliegeois** [comment]
@bartgol there is, however, an issue with what I proposed: to have 1 `STKDiscretization` per block and reuse the same `MeshStruct` for different blocks.
Currently, the number of equations `neq` is set as a data member in the `MeshStruct` and we will probably want to have different number of equations for different blocks of the block discretization (this number of equations is required by `fieldContainer` too).
If we want to move forward with that solution, we have to decide if we want to remove `neq` from the `MeshStruct` classes, this could require some refactoring work.
I think that the mesh structure should not be aware of the number of equations as this number has no influence on the geometric mesh.

Alternatively, we can consider using one `MeshStruct` which stores an array of `neq`, one per block associated with the current `MeshStruct`.

Another thought is that I think that the option where we consider multiple meshes would potentially not be compatible with the `panzer::BlockedDOFManager`.

**Comment by mperego** [comment]
> @mperego So basicallly what they are doing is "interpolate the given FE to P0, then save to exodus", right? We could do something like that for now: assume the mesh is P1 (in fact, we could enforce this), then if the DOF is not P1 or P0, interpolate before saving. I believe mixing P0 and P1 is fine (exodus can handle cell and node values).
> 
> I think in the short term (at least), P0 and P1 are all we might be dealing with, so this might be an acceptable compromise. As stk/exodus handling of mixed FE types improves (thank you, Panzer, for being our guinea pig), we can remove the dof interpolation.

It makes sense to me, bu what about restart?

**Comment by bartgol** [comment]
> > @mperego So basicallly what they are doing is "interpolate the given FE to P0, then save to exodus", right? We could do something like that for now: assume the mesh is P1 (in fact, we could enforce this), then if the DOF is not P1 or P0, interpolate before saving. I believe mixing P0 and P1 is fine (exodus can handle cell and node values).
> > I think in the short term (at least), P0 and P1 are all we might be dealing with, so this might be an acceptable compromise. As stk/exodus handling of mixed FE types improves (thank you, Panzer, for being our guinea pig), we can remove the dof interpolation.
> 
> It makes sense to me, bu what about restart?

Yes, there may be an inconsistency if we do, say, P2->P1 to save, then P1->P2 when restarting. 

An alternative would be to allow any kind of mesh. Simply, if dof_fe!=mesh_fe, do interpolation. This would allow to have a pb with P2 fields, and do an exact restart.

**Comment by bartgol** [comment]
> @bartgol there is, however, an issue with what I proposed: to have 1 `STKDiscretization` per block and reuse the same `MeshStruct` for different blocks.
> Currently, the number of equations `neq` is set as a data member in the `MeshStruct` and we will probably want to have different number of equations for different blocks of the block discretization (this number of equations is required by `fieldContainer` too).
> If we want to move forward with that solution, we have to decide if we want to remove `neq` from the `MeshStruct` classes, this could require some refactoring work.
> I think that the mesh structure should not be aware of the number of equations as this number has no influence on the geometric mesh.
> 

You're preaching to the choir. I've disliked how we mix geometry and numerical appx stuff into the same struct (the mesh struct) for a long time.

> Alternatively, we can consider using one `MeshStruct` which stores an array of `neq`, one per block associated with the current `MeshStruct`.
> 
> Another thought is that I think that the option where we consider multiple meshes would potentially not be compatible with the `panzer::BlockedDOFManager`.

So, looking at the code, it seems that `neq` is needed cause we store a field called `solution`, of dimension `neq`. Imho, we could do away with this. Here are my thoughts.

- The field container is currently part of the mesh struct. That's not great. I'd move it inside the Discretization. Fields have a FE type associated, and the mesh doesn't (shouldn't) know about FE. The mesh should only provide mesh-related info.
- Having a big field called `solution` in the exo file is not really helpful. You often end up extracting the different components, just to print what you're after. E.g., when solving ice+hydro,  `solution=[u, v, N, h]`, but I always end up picking a component to plot. The only benefit (maybe...maybe) is that a restart only has to look for one field to load.
- The MultiSTKFieldContainer might end up being what we should use from now on: it requires an array of names for the solution fields. This would allow to create separate fields with meaningful names, and do away with `solution`. I believe that MultiSTKFieldContainer, at least in the case where the user _does_ provide names for the sol vectors, doesn't _really_ need neq. It's used to setup a NodalDOFManager (merely a struct to basically just get correct local/global indices) when loading/saving the solution. But I believe that if our Thyra vector is blocked, it's not needed there either.

Long story short: yes, I think we should have Mesh structs that _only_ store information on the mesh, without knowledge on the solution field.

Note: we can still access the stk fields from the mesh, if need be (they are in the stk MetaData, after all). E.g., we may want to access the coordinate fields (to move the mesh maybe?). Or, more important for us, the thickness/surface_height fields when building an extruded mesh. That's fine, cause those are "geometry" fields, and the mesh can assume they are defined on all mesh nodes (whatever the mesh element is).

**Comment by kliegeois** [comment]
@bartgol I thought about the `STKFieldContainer` and how to remove it from the `STKMeshStruct`.
I understand that it could be totally moved to the `STKDiscretization` but I think that it could be more efficient both in terms of implementation work and memory usage to divide the `STKFieldContainer` into two field containers one related to the mesh and one to the solution. This way, when we have multiple `STKDiscretization` associated to a same mesh, those `STKDiscretization` can reuse/share the same `STKFieldContainer` related to the mesh and each one can have a `STKFieldContainer` related to the solution.

**Comment by bartgol** [comment]
@kliegeois the two containers would be of the same type though, right? Like both `STKFieldContainer` or something. If so, than I think what you say makes sense. Otherwise, I would advocate for incorporating the interface to access mesh fields directly in the mesh struct...

**Comment by kliegeois** [comment]
@bartgol I was more thinking of having two different types one related to the solution and one to the mesh.
You are suggesting having only one type and one instantiation at the mesh structure level and multiple instantiations at the discretization level which would have a reference to the one at the mesh structure level?

What do you mean by accessing mesh fields directly in the mesh struct? Do you mean accessing the fields stored in the container associated to the discretization? If so, wouldn't it be a loop dependency?

**Comment by bartgol** [comment]
> @bartgol I was more thinking of having two different types one related to the solution and one to the mesh.
> You are suggesting having only one type and one instantiation at the mesh structure level and multiple instantiations at the discretization level which would have a reference to the one at the mesh structure level?

Yes, so each disc would store a `RCP<STKFieldContainer>` with the fields that have that disc type. The mesh would also store a `RCP<STKFieldContainer>`, with only mesh-related fields. All discs should have a ptr to their stk mesh, so if they need, they could also access the mesh's field container. I am assuming that the same field container interface would work for both FE field and Mesh fields. Do you think it's a bad assumption? Or, put it this way, what differences do you think there should be in the interface of a FE field container and a Mesh field container?

> 
> What do you mean by accessing mesh fields directly in the mesh struct? Do you mean accessing the fields stored in the container associated to the discretization? If so, wouldn't it be a loop dependency?

I meant this for the STK mesh only. So in `GenericSTKMeshStruct` (or the abstract one, maybe), put something like this:

```
using ScalarFieldType = stk::mesh::Field<double>;
using VectorFieldType = stk::mesh::Field<double,PHAL::VecDim>;

ScalarFieldType* get_cell_scalar_field (const std::string& name) const;
VectorFieldType* get_cell_vector_field (const std::string& name) const;

ScalarFieldType* get_node_scalar_field (const std::string& name) const;
VectorFieldType* get_node_vector_field (const std::string& name) const;
// Shortcut for the previous one, with name=="coordinates"
VectorFieldType* get_coordinates_field () const { return get_node_vector_field("coordinates"); }
```
The first four methods would simply query the undelrying mesh. E.g.:
```
ScalarFieldType* get_node_scalar_field(const std::string& name) const {
  return metaData->get_field<ScalarFieldType>(stk::topology::NODE_RANK,name);
}
```
I guess it's basically what a field container would do, so probably it's ok to just store a STKFieldContainer for the mesh fields...

**Comment by kliegeois** [comment]
@bartgol I think that what you suggest should work. I can try it tomorrow. 

What I try today is the following:

For the mesh related field container use the following constructor interface:
```
template <DiscType Interleaved>
class MultiSTKFieldContainer : public GenericSTKFieldContainer<Interleaved>
{
 public:
  MultiSTKFieldContainer(
      const Teuchos::RCP<Teuchos::ParameterList>&               params_,
      const Teuchos::RCP<stk::mesh::MetaData>&                  metaData_,
      const Teuchos::RCP<stk::mesh::BulkData>&                  bulkData_,
      const int                                                 numDim_,
      const Teuchos::RCP<Albany::StateInfoStruct>&              sis,
      const Teuchos::Array<Teuchos::Array<std::string>>&        solution_vector,
      const int                                                 num_params);
```

For the discretization related field container use the following constructor interface:
```
 public:
  MultiSTKSolutionFieldContainer(
    const Teuchos::RCP<Teuchos::ParameterList>& params_,
    const int numDim_,
    const int neq_,
    const Teuchos::RCP<MultiSTKFieldContainer<Interleaved>>& fieldContainer_,
    const Teuchos::Array<Teuchos::Array<std::string>>& solution_vector,
    const int                                          num_params_);
```

The `MultiSTKSolutionFieldContainer` stores, as a private member data, a rcp to the field container associated to the mesh which allows to access the `metaData` and the `bulkData` of the `MultiSTKFieldContainer` from the `MultiSTKSolutionFieldContainer` level.

I suppose that it would be possible to have only one class where we are using some logic to check how we should access the `metaData` or if we have to add the solution field for the current instantiation.

Alternatively, as you said, we can remove the field container on the mesh structure level and use the `metaData` in the mesh structure.

Honestly, I am not a fan of my solution with 2 classes as it adds more classes and more files to compile.

**Comment by bartgol** [comment]
A couple of comments:

- why storing the mesh field container in the FE one? Do you think you need it? If the only reason is so you can access meta/bulk data, then simply pass metaData and bulkData, imho.
- I don't think I want the mesh container to handle the states. I think it should be done by the disc one. But I see how that would complicate things: we would need the StateManager to separate the states based on their discretization type. It would be much cleaner, but might require a few more impl steps... Up to you.
- I see that num_params is not even used in MultiSTKFieldContainer, but only in the Ordinary one. I suspect we simply didn't add the feature to the Multi container, since we don't really use it much. @mperego @ikalash thoughts?
- The ctor for the mesh version doesn't need the `solution_vector` input, I think.
- The ctor for the FE version needs more info than just the solution names. In particular, it has to provide info on "where" the field lives (nodes, cell, edges,...). I think we should create a mini-helper struct, like so
```
struct FieldSpec {
  std::string name;
  stk::topology::topology_t entity;
  int rank;
  std::vector<int> dims; // only if rank>0
};
```
Then, we can pass a `Teuchos::Array<FieldSpec>`. I don't like that right now we pass an array of arrays (for x, xdot, xdotdot). I think we should have a single array, with all the solution fields that need to be registered in the stk data structures. After all, the MultiSTKFieldContainer does not make any assumption on what the 1st array is, vs the 2nd or 3d, so we might as well collapse all of them in one.

**Comment by kliegeois** [comment]
Since the end of last week, I tried both approaches; 1 class ([branch combinedcontainer](https://github.com/kliegeois/Albany/tree/combinedcontainer) ) and 2 classes ([branch solutionfieldcontainer](https://github.com/kliegeois/Albany/tree/solutionfieldcontainer) ).
Up to now, they lead to the following failing tests:

- 1-class implementation:

        The following tests FAILED:
                107 - LandIce_FO_Dome_Ascii (Failed)
                108 - LandIce_FO_Dome_Restart (Failed)
                119 - FO_MMS_FO_GLF_QUAD (Failed)
                120 - FO_MMS_FO_GLF_TRI (Failed)
                121 - FO_MMS_FO_GLF_TRI_CORNER_CASE (Failed)
                122 - FO_MMS_FO_GLF_TRI_NONCONST_VELOCITY (Failed)
                123 - FO_MMS_FO_SMB_QUAD (Failed)
                124 - FO_MMS_FO_SMB_TRI (Failed)
                139 - FO_GIS_PopulateMeshes (Failed)
                141 - FO_GIS_Unstructured_Epetra (Not Run)
                142 - FO_GIS_Unstructured_Restart (Not Run)
                143 - FO_GIS_Unstructured_Tpetra (Not Run)
                144 - FO_GIS_Unstructured_Memoization (Not Run)
                145 - FO_GIS_Gis20km_Epetra (Failed)
                147 - FO_GIS_CoupledThickness_Epetra (Failed)
                148 - FO_GIS_CoupledThickness_Tpetra (Failed)
                149 - FO_GIS_AdjointSensitivity_Epetra (Failed)
                150 - FO_GIS_AdjointSensitivity_Tpetra (Failed)
                151 - FO_GIS_ForwardSensitivity_Tpetra (Failed)
                152 - FO_GIS_AdjointSensitivity_Wedge (Failed)
                153 - FO_GIS_AdjointSensitivity_BasalFriction_Epetra (Failed)
                154 - FO_GIS_AdjointSensitivity_BasalFriction_Tpetra (Failed)
                155 - FO_GIS_Analysis_BasalFriction (Failed)
                156 - FO_GIS_Analysis_BasalFriction_Memoization (Failed)
                157 - FO_GIS_Analysis_BasalFriction_Hessian (Failed)
                158 - FO_GIS_AdjointSensitivity_StiffeningBasalFriction_Epetra (Failed)
                159 - FO_GIS_AdjointSensitivity_StiffeningBasalFriction_Tpetra (Failed)
                160 - FO_GIS_Analysis_StiffeningBasalFriction (Failed)
                161 - FO_GIS_Analysis_StiffeningBasalFriction_Memoization (Failed)
                162 - FO_GIS_SensSMBwrtBeta_Epetra (Failed)
                163 - FO_GIS_SensSMBwrtBeta_Restart (Failed)
                164 - FO_GIS_SensSMBwrtBeta_Tpetra (Failed)
                165 - FO_GIS_Manifold (Failed)
                166 - FO_GIS_LaplacianPriorSampling (Failed)
                167 - FO_Thermo_Dry_Bed (Failed)
                168 - FO_Thermo_Wet_Bed (Failed)
                169 - FO_Thermo_Humboldt_fluxDiv (Failed)
                189 - Enthalpy_Humboldt (Failed)
                191 - Enthalpy_Kleiner_B (Failed)
                192 - Enthalpy_Kleiner_A (Failed)
                203 - Discretization_Serial_Unit_Test (Failed)
        Errors while running CTest

- 2-class implementation:

        The following tests FAILED:
                166 - FO_GIS_LaplacianPriorSampling (Failed)
        Errors while running CTest

However, I should be able to make the 1-class implementation behave like the 2-class implementation and which fails only on (I discuss that later in this comment) `FO_GIS_LaplacianPriorSampling`.
The tests that are failing with the 1-class implementation and that are not failing with the 2-class implementation are related to the management of side sets which I did not implement correctly yet.

@bartgol the proposed prototype constructor for the 2-class implementation is not yet cleaned; some arguments are not used at all. I will do that if we consider that option which is less likely.

I run into several issues with both of the 2 implementations:
1. As stated, the goal is to have one field container at the mesh structure level and possibly more than one field container at the discretization level (one field container per block of the block discretization). Therefore, I did something as follows:

        Teuchos::RCP<AbstractDiscretization>
        DiscretizationFactory::createDiscretization(
                unsigned int neq, const std::map<int, std::vector<std::string> >& sideSetEquations,
                const Teuchos::RCP<StateInfoStruct>& sis,
                const std::map<std::string, Teuchos::RCP<StateInfoStruct> >& side_set_sis,
                const AbstractFieldContainer::FieldContainerRequirements& req,
                const std::map<std::string, AbstractFieldContainer::FieldContainerRequirements>& side_set_req,
                const Teuchos::RCP<RigidBodyModes>& rigidBodyModes) 
        {
            TEUCHOS_TEST_FOR_EXCEPTION(meshStruct == Teuchos::null,
                    std::logic_error,
                    "meshStruct accessed, but it has not been constructed" << std::endl);
        
            setupInternalMeshStruct(sis, side_set_sis, req, 
                                    side_set_req);
            Teuchos::RCP<AbstractDiscretization> result =
                    createDiscretizationFromInternalMeshStruct(neq, sideSetEquations, rigidBodyModes, req, sis);
        
            meshStruct->loadRequiredInputFields(req,commT);
            return result;
        }
    Concretely, first the mesh struct is filled, then, the discretization is built where the number of blocks will be deduced from a parameter list and the field container at the block level are constructed.
Finally, those field containers are filled.

    The issue is that those block-level field containers share the same metaData pointer as the mesh-structure-level field container for which the metaData has already been committed (`metaData->commit()`).

    Therefore, `enable_late_fields()` is required (with the above-mentioned code), I do not know if this can have a performance impact on our tests.

2. As the solution field are added to the metaData at the block-level field container, there are not yet added when we read the input file. This implies that, for the `FO_GIS_LaplacianPriorSampling` the code reads the exodus file and observes that the solution has 2 components per entry (dimension of the solution stored in the file) and an error is thrown when the code tries to redefine the solution field to be scalar:

        166:  Expr '!(restr->num_scalars_per_entity() != tmp.num_scalars_per_entity())' eval'd to true, throwing.
        166:  Error occured at: stk_mesh/stk_mesh/baseImpl/FieldBaseImpl.cpp:282
        166:  Error: std::mesh::MetaData::declare_field_restriction FAILED for FieldBaseImpl<double,Cartesian3d>[ name = "solution" , #states = 1 ] FieldRestriction[ selector: "{UNIVERSAL}", dimension: 2, scalars per entity: 2 ] WITH INCOMPATIBLE REDECLARATION FieldRestriction[ selector: "{UNIVERSAL}", dimension: 1, scalars per entity: 1 ]

    Reading an exodus file which does not store the solution would be sufficient to make this test pass.


Therefore, I think that the block-level field containers should be created concurrently with the mesh-structure-level field container.
@bartgol do you have comments regarding that?
A side comment is that we can consider using Panzer to add the fields on different blocks.

**Comment by bartgol** [comment]
Ah, the meta data commit issue is annoying. I have no idea what the perf implications are. An exception msg in stk says
```
          ThrowRequireMsg(m_meta_data->are_late_fields_enabled(),    
                          "Attempting to register Field '" << m_name << "' after MetaData is" << std::endl <<    
                          "committed. If you are willing to accept the performance implications, call" << std::endl <<    
                          "MetaData::enable_late_fields() before adding these Fields.");    
```
It doesn't really help to figure out what this means though. So I guess we could start by trying to avoid using the feature.

What if we create the disc _before_ completing the mesh setup? We could have all the containers created, and then close all meta data. So something like this:

```
DiscretizationFactory::createDiscretization(...) {
    TEUCHOS_TEST_FOR_EXCEPTION(meshStruct == Teuchos::null,...);

    // Create bare discretization
    auto disc = createDiscretizationFromInternalMeshStruct(...);
    
    // Register all fields in the different containers
    disc->setFieldData (...);
    
    // Close metaData and complete mesh setup
    setupInternalMeshStruct(...);
    
    // basically just calls updateMesh, if meshSpecsType==STK_MS, but for the sake of 'generality', delegate to helper fcn.
    // Otherwise, simply call disc->updateMesh(), possibly enclosed in `if(meshStruct->meshSpecsType()==STK_MS)`.
    completeDiscSetup (...);
}
```

**Comment by bartgol** [comment]
Regarding the restart issue (2d solution in input file, and scalar solution1 solution2 in the disc): if we register the fields before completing the mesh setup (as suggested above), we should have all fields registered before we load the mesh. This, however, would not get rid of the vector-solution vs scalar-solution-blocks issue.

IMHO, getting a `[u v]` problem to load the solution from a `[horiz_velocity]` file is an overkill. If need be, we could create a helper problem, that loads a vector field, and saves individual components to file with given names (and viceversa), which can be used as a pre-processing tool in case one wants to restart a simulation changing the blocks structure. But for now, I think it's fine to assume we restart with the same block structure as in the input mesh. What do you think?

**Comment by kliegeois** [comment]
@bartgol thanks! I think that I agree with what you suggest; it is consistent with what I meant with "should be created concurrently".

I doubt that we will need to change the order of the blocks when restarting the simulation.

**Comment by kliegeois** [comment]
@bartgol, @mperego  I think that I achieved to correctly compute the graph of the off-diagonal blocks associated to a same mesh and with potentially different number of equations per block but with a P1 discretization for each block.

However, I still need to implement those off-diagonal graphs for:
- the volume/side set interaction,
- the side set/side set interaction,
- when we use P0/P1 discretization (do we have tests/problems which use `HVOL_C0` discretization? ).

As previously discussed, the `computeGraph()` function computes the graph of all blocks located on the lower block triangle.
When creating a strictly upper triangular block of the blocked operator, the graph of the corresponding lower triangular block is used to create a Thyra operator which is then transposed before being set as the corresponding block of the operator.

How easy will it be to fill the local data of this transposed block knowing that it cannot be cast to a Tpetra crs matrix directly?

**Comment by bartgol** [comment]
> How easy will it be to fill the local data of this transposed block knowing that it cannot be cast to a Tpetra crs matrix directly?

That's a very good point. And even if we extracted the [T|E]petra matrix that is stored inside the transpose op, we'd be doing assembly by column, which is definitely not easy/desirable.

I guess it is best to save a RCP<ThyraCrsMatrixFactory> for _all_ blocks. We can, of course, be smart and recycle pointers if the domain and range disc are the same. E.g, if we have a block problem with blocks `[u, v]`, and both u and v are, say, P1, then we can store just one RCP<ThyraCrsMatrixFactory>.

But in general, I think the point you raised is forcing us to store a factory for each block.

**Comment by kliegeois** [comment]
@bartgol I agree, it will be easier.

Do you have thoughts related to the P0/P1 case?
Do you know if we have tests/problems which use HVOL_C0 discretization?

**Comment by bartgol** [comment]
Re:P0/P1. It should be fairly easy to implement, right? Unfortunately we do not have any P0/P1 test case. But I might want to use it for the hydrology, where the eff. pressure `N` is in H1, while the water thickness `h` can be in L2.

I am not sure how easy it'd be to cook up a small toy problem to test this. Even if we created a block op for, say, a Stokes pb with P1 velocity and P0 pressure (not LBB stable, but we're not interested in that here, just the Jac graph), we would not be able to compare with a non-blocked graph, since Albany does not do P1-P0 coupling.

My best idea for a test is to create a small STK1D discretization with 3 elements, create a Jac for a P1-P0 pb, and save the jac to file (using `writeMatrixMarket` utility). Then, create a 2nd test (dependent on the 1st) that simply runs a `diff` against a reference file (which we can create by hand and inspect in matlab/python/octave to make sure it has the correct structure).

**Comment by ikalash** [comment]
You could use one of the incompressible Navier stokes tests in Albany to verify p1/p0.  Those run now with stabilization and p1/p1.

**Comment by bartgol** [comment]
> You could use one of the incompressible Navier stokes tests in Albany to verify p1/p0. Those run now with stabilization and p1/p1.

Do they assemble p1/p1 or p1/p0?

**Comment by ikalash** [comment]
It’s P1/p1 with stabilization.  You can’t do p1/p0 currently in Albany.

**Comment by bartgol** [comment]
> It’s P1/p1 with stabilization. You can’t do p1/p0 currently in Albany.

Right, so I don't know how else we can test a P1/P0 block assembly other than just use a very small mesh, save the jac to file, and then diff against a gold file.

**Comment by ikalash** [comment]
Maybe I’m missing something but wouldn’t the goal be to run the problem with P1/p0 without stabilization when the block assembly is ready and check that the result is close to the expected one (say the current stabilized one)?

**Comment by bartgol** [comment]
Well, P1/P0 is not inf-sup stable, so we wouldn't be using it for NS. I think the goal here is to see that we can assemble a block jacobian using two different FE spaces. It's the next thing to try after verifying we can assemble a P1/P1 block jac.

What you suggest is I think way ahead of where we are. :) To do that, we'd need to have all the pieces of block disc/assembly/solver done. But we should keep this kind of test in mind for when our block structure is all in place.

**Comment by kliegeois** [comment]
@bartgol I think that P1/P0 is relatively straightforward to implement but we need to discuss one implementation detail. 

P0 elements have one dof per element. However, the STKDiscretization class relies on the list of nodes to create the vector spaces and the graph. My question was more whether we agree that the STKDiscretization (for one block) should be revisited for P0 elements or if we already support that capability by doing some operations that I am not aware of.

**Comment by bartgol** [comment]
Yeah, we need to revisit a bit. For P1/P0, instead of retrieving the cell's nodes via

```
stk::mesh::Entity const* node_rels = bulkData.begin_nodes(e);
```
we can do
```
auto dofs = bulkData.begin_entities(dof_entity_rank);
```
For higher order elements, this does not work, cause we can have dofs on entities of multiple ranks (e.g., P2 on tri have node and edge dofs).

mperego:
I think we can limit ourselves to low order elements in STK. If we have high-ordr FE we can always project them to a lower order. The only issue is restart.. but we can think about that another time

bartgol:
> I think we can limit ourselves to low order elements in STK. If we have high-ordr FE we can always project them to a lower order. The only issue is restart.. but we can think about that another time

Well, here we are talking about the DOFs, not the output. We use the entity_id of the stk entity where the dof lives to give it a GID. 

Actually, that approach will be buggy for high order FE, since there might be 2+ dofs on the same geo entity.

I agree, for now we can limit ourselves to Hgrad_C1 and Hvol_C0. But if we go higher, we need to find a way to associate gids to the dofs that does something more than simply picking the gid of the stk entity where the dof lives...

mperego:
> > I think we can limit ourselves to low order elements in STK. If we have high-ordr FE we can always project them to a lower order. The only issue is restart.. but we can think about that another time
> 
> Well, here we are talking about the DOFs, not the output. We use the entity_id of the stk entity where the dof lives to give it a GID.
> 
> Actually, that approach will be buggy for high order FE, since there might be 2+ dofs on the same geo entity.
> 
> I agree, for now we can limit ourselves to Hgrad_C1 and Hvol_C0. But if we go higher, we need to find a way to associate gids to the dofs that does something more than simply picking the gid of the stk entity where the dof lives...

I think that's is the job of the DofManager. Basically the STK connectivity manager gives the mesh connectibity and vertices, and based on that and on the finite element, the DofManager gives back a local and global ID numbers. Sorry if I'm sayng somethig obvious or not that relevant... I haven't followed all the discussion in detail

bartgol:
It's not obvious, nor irrelevant. What you say makes sense. I haven't looked too much in detail at the DofManager. But if it provides that capability (and it makes sense that it does), then we can also simplify a bit our stk disc class, since we don't have to bother doing all those manual loops on stk entities, and just rely on dof mgr and the conn mgr inside it.

mperego:
Yeah, and that will help separating the mesh from the discretization.
I don't know  whether Glen got to the point of hooking it up correctly with the STK connectivity manager though.

kliegeois:
Thanks for your comments! I will look into that.

kliegeois:
@bartgol @mperego I looked today into the use of the `DofManager` to compute the graphs and the vector spaces.

It looks like it allows simplifying a lot the code for the computation of the graph of the blocks.

There is, however, one drawback, only scalar fields can be added to the `DofManager`. This means that, for the velocity in 3D for example, we have to add 3 fields. This has two impacts:

1. The `block names` string is longer: `"velocity"` becomes `"[Ux, Uy, Uz]"` for example,
2. If we want to keep the logic of the Panzer field ordering, it seems that we cannot nest blocks as previously proposed `"[ [Ux, Uy, Uz], [N, h]]"`. To be more precise, we can ask to have two blocks (one for the velocity and one for `N` and `h`) but it seems that we cannot specify the block structure of the second block to be blocked too. It appears that if the subblocks share the same block element and the same pattern ( as it is for `"Ux"` and `"Uy"` for example) the DOF are stored with an interleaved indexing (`ux0, uy0, uz0, ux1, ...`) otherwise, the IDs are blocked.

I am fine with both points, if we really want to enforce that `N` and `h` are not interleaved (if it applies) or if we want to access their block separately for preconditioning purpose, we can use a 3x3 block operator.

mperego:
> 
> 
> @bartgol @mperego I looked today into the use of the `DofManager` to compute the graphs and the vector spaces.
> 
> It looks like it allows simplifying a lot the code for the computation of the graph of the blocks.
> 
> There is, however, one drawback, only scalar fields can be added to the `DofManager`. This means that, for the velocity in 3D for example, we have to add 3 fields. This has two impacts:
> 
>     1. The `block names` string is longer: `"velocity"` becomes `"[Ux, Uy, Uz]"` for example,
> 
>     2. If we want to keep the logic of the Panzer field ordering, it seems that we cannot nest blocks as previously proposed `"[ [Ux, Uy, Uz], [N, h]]"`. To be more precise, we can ask to have two blocks (one for the velocity and one for `N` and `h`) but it seems that we cannot specify the block structure of the second block to be blocked too. It appears that if the subblocks share the same block element and the same pattern ( as it is for `"Ux"` and `"Uy"` for example) the DOF are stored with an interleaved indexing (`ux0, uy0, uz0, ux1, ...`) otherwise, the IDs are blocked.
> 
> 
> I am fine with both points, if we really want to enforce that `N` and `h` are not interleaved (if it applies) or if we want to access their block separately for preconditioning purpose, we can use a 3x3 block operator.

I'm also fine with that.

bartgol:
But we would still get U=[u,v,w] and Nh=[N,h] to be two blocks, right? If so, then I guess it's ok. If this becomes a restrictive limitation later on, we can work around it then. One step at a time, I guess. =)

kliegeois:
Yes, if we use: [[u,v,w],[N,h]] or even 3 blocks if we use: [[u,v,w], N, h].

kliegeois:
@mperego @bartgol this PR can be reviewed and merged.
The branch is as discussed during our last meeting:
- The computation of the graph of the block uses the connection manager and the DOF manager,
- Documentation has been added inside the function which computes the graph to explain the method,
- The branch is warning free,
- The tests passed on my computer.

mperego:
@kliegeois thanks! it looks good to me but I didn't reviewed it carefully. @bartgol should we merge this?

bartgol (approved):
@mperego Yes, I'm gonna merge it now, so that the nightlies have a chance to test this tonight, and Kim has a chance to look at cdash if something is wrong.

mperego:
@ikalash I just commit a fix from @kliegeois for Cism, although we didn't test it w/ Cism. Hopefully tomorrow the Dashboard will be clean.

ikalash:
Thanks @mperego .  So I just built the codes, and they built but all the CISM tests are seg faulting...

mperego:
argh.. worst possible outcome

mperego:
@ikalash I'm seeing issues in MPAS as well.

ikalash:
@mperego : lets open a separate issue.  Often seg faults are pretty easy to hunt down - something's not allocated and accessed, or something like this.  I need to find time to create a debug build and run through gdb.  I'm glad it's not just CISM.  It's likely the same problem affecting both codes.

PR #693, opened 2021-05-27
Reduced sideset allocation size for single workset case

jewatkins:
This reduces the sideset allocation size in the case of a single workset by setting the correct full workset sizes in mesh specs. It's controlled by a mesh specs parameter called `singleWorksetSizeAllocation`. It currently only works for `IossSTKMeshStruct`, `ExtrudedSTKMeshStruct` and `SideSetSTKMeshStruct` but it should be easy to extend to other mesh structs as long as the mesh struct knows the exact sideset sizes. If this is not known, the sidesets will be set to the original larger size.

Some notable changes:
1. if a single workset is used in the top level discretization, all sideset structs will also use a single workset
2. initializing the sideset mesh specs is now required before initializing the sideset mesh structs (i.e. it will throw). I think this is what was originally intended anyways?

@mcarlson801

### DIFF
--- a/src/disc/Albany_DiscretizationFactory.cpp
+++ b/src/disc/Albany_DiscretizationFactory.cpp
@@ -105,12 +105,14 @@ DiscretizationFactory::createMeshStruct(Teuchos::RCP<Teuchos::ParameterList> dis
         int extruded_ws_size = disc_params->get("Workset Size", 50);
         int basal_ws_size = -1;
         if(extruded_ws_size != -1) {
-          basal_ws_size =  extruded_ws_size/ (disc_params->get<int>("NumLayers") * ((disc_params->get<std::string>("Element Shape") == "Tetrahedron") ? 3 : 1));
+          const int num_cells_per_layer = disc_params->get<std::string>("Element Shape") == "Tetrahedron" ? 3 : 1;
+          const int num_cells_per_column = num_cells_per_layer * disc_params->get<int>("NumLayers");
+          basal_ws_size = extruded_ws_size / num_cells_per_column;
           basal_ws_size = std::max(basal_ws_size,1); //makes sure is at least 1.
         }
         if (disc_params->isSublist("Side Set Discretizations") && disc_params->sublist("Side Set Discretizations").isSublist("basalside")) {
             basal_params = Teuchos::rcp(new Teuchos::ParameterList(disc_params->sublist("Side Set Discretizations").sublist("basalside")));
-            if(!disc_params->sublist("Side Set Discretizations").isParameter("Workset Size"))
+            if(!disc_params->sublist("Side Set Discretizations").isParameter("Workset Size") || extruded_ws_size == -1)
               basal_params->set("Workset Size", basal_ws_size);
         } else {
             // Backward compatibility: Ioss, with parameters mixed with the extruded mesh ones
--- a/src/disc/Albany_DiscretizationFactory.cpp
+++ b/src/disc/Albany_DiscretizationFactory.cpp
@@ -105,14 +105,17 @@ DiscretizationFactory::createMeshStruct(Teuchos::RCP<Teuchos::ParameterList> dis
         int extruded_ws_size = disc_params->get("Workset Size", 50);
         int basal_ws_size = -1;
         if(extruded_ws_size != -1) {
-          const int num_cells_per_layer = disc_params->get<std::string>("Element Shape") == "Tetrahedron" ? 3 : 1;
-          const int num_cells_per_column = num_cells_per_layer * disc_params->get<int>("NumLayers");
+          const int num_cells_per_column_per_layer = disc_params->get<std::string>("Element Shape") ==
+              "Tetrahedron" ? 3 : 1; // number of 3D cells in one extruded 2D cell
+          const int num_cells_per_column = num_cells_per_column_per_layer * disc_params->get<int>("NumLayers");
           basal_ws_size = extruded_ws_size / num_cells_per_column;
           basal_ws_size = std::max(basal_ws_size,1); //makes sure is at least 1.
         }
         if (disc_params->isSublist("Side Set Discretizations") && disc_params->sublist("Side Set Discretizations").isSublist("basalside")) {
             basal_params = Teuchos::rcp(new Teuchos::ParameterList(disc_params->sublist("Side Set Discretizations").sublist("basalside")));
-            if(!disc_params->sublist("Side Set Discretizations").isParameter("Workset Size") || extruded_ws_size == -1)
+            if (extruded_ws_size == -1)
+              basal_params->set("Workset Size", -1);
+            else if (!disc_params->sublist("Side Set Discretizations").isParameter("Workset Size"))
               basal_params->set("Workset Size", basal_ws_size);
         } else {
             // Backward compatibility: Ioss, with parameters mixed with the extruded mesh ones
--- a/src/disc/Albany_DiscretizationFactory.cpp
+++ b/src/disc/Albany_DiscretizationFactory.cpp
@@ -100,30 +100,32 @@ DiscretizationFactory::createMeshStruct(Teuchos::RCP<Teuchos::ParameterList> dis
     }
     else if (method == "Extruded") {
         Teuchos::RCP<AbstractMeshStruct> basalMesh;
+
+        // Get basal_params
         Teuchos::RCP<Teuchos::ParameterList> basal_params;
-        //compute basal Workset size starting from Discretization
-        int extruded_ws_size = disc_params->get("Workset Size", 50);
-        int basal_ws_size = -1;
-        if(extruded_ws_size != -1) {
-          const int num_cells_per_column_per_layer = disc_params->get<std::string>("Element Shape") ==
-              "Tetrahedron" ? 3 : 1; // number of 3D cells in one extruded 2D cell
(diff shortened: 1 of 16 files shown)


src/disc/Albany_DiscretizationFactory.cpp
```diff
@@ -105,12 +105,14 @@ DiscretizationFactory::createMeshStruct(Teuchos::RCP<Teuchos::ParameterList> dis
         int extruded_ws_size = disc_params->get("Workset Size", 50);
         int basal_ws_size = -1;
         if(extruded_ws_size != -1) {
-          basal_ws_size =  extruded_ws_size/ (disc_params->get<int>("NumLayers") * ((disc_params->get<std::string>("Element Shape") == "Tetrahedron") ? 3 : 1));
+          const int num_cells_per_layer = disc_params->get<std::string>("Element Shape") == "Tetrahedron" ? 3 : 1;
```
bartgol, line 108: The name is a bit misleading, since it sounds like it's the number of cells over a whole horizontal layer, rather than the number of 3d cells in one layer of an extruded single 2d cell. That said, I think finding a self-explanatory name is quite hard (unless one is fine with 100chars-long names). But maybe we could clarify it with a comment?

src/disc/Albany_DiscretizationFactory.cpp
```diff
@@ -105,12 +105,14 @@ DiscretizationFactory::createMeshStruct(Teuchos::RCP<Teuchos::ParameterList> dis
...
+          const int num_cells_per_column = num_cells_per_layer * disc_params->get<int>("NumLayers");
+          basal_ws_size = extruded_ws_size / num_cells_per_column;
           basal_ws_size = std::max(basal_ws_size,1); //makes sure is at least 1.
         }
         if (disc_params->isSublist("Side Set Discretizations") && disc_params->sublist("Side Set Discretizations").isSublist("basalside")) {
```
bartgol, line 113: Question: shouldn't we _always_ have a "Side Set Discretization"->"basalside" sublist for Extruded meshes?

src/disc/Albany_DiscretizationFactory.cpp
```diff
@@ -105,12 +105,14 @@ DiscretizationFactory::createMeshStruct(Teuchos::RCP<Teuchos::ParameterList> dis
...
             basal_params = Teuchos::rcp(new Teuchos::ParameterList(disc_params->sublist("Side Set Discretizations").sublist("basalside")));
-            if(!disc_params->sublist("Side Set Discretizations").isParameter("Workset Size"))
+            if(!disc_params->sublist("Side Set Discretizations").isParameter("Workset Size") || extruded_ws_size == -1)
```
bartgol, line 115: Did you mean to access "Side Set Discretizations"->"basalside"->"Workset Size" instead? Or is the idea to set the ws size for all side meshes?

src/disc/stk/Albany_ExtrudedSTKMeshStruct.cpp
```diff
@@ -195,6 +198,51 @@ Albany::ExtrudedSTKMeshStruct::ExtrudedSTKMeshStruct(const Teuchos::RCP<Teuchos:
   // Create a mesh specs object for EACH side set
   this->initializeSideSetMeshSpecs(comm);
 
+  // Get upper bound on lateral/upper workset sizes by using Ioss element counts on side blocks
+  if (worksetSizeMax == -1 && basalMeshSpec->singleWorksetSizeAllocation) {
+    // Set lateral workset sizes based on basal sidesets
+    for (auto bssName : basalMeshSpec->ssNames) {
+      // Get maximum workset size of basalside sideset
+      const auto& basalSideSetMeshSpecs = basalMeshSpec->sideSetMeshSpecs;
```
bartgol, line 206: I'm a bit confused. Are you trying to get the mesh specs of the side of the basal mesh, i.e., a 1d mesh? I wasn't aware we stored those. I thought we stored mesh specs only when we built a mesh, and I thought we didn't build a mesh for the side of the side mesh.

bartgol (commented):
Looks good. I have some inline comments.

src/disc/Albany_DiscretizationFactory.cpp
jewatkins: Yeah, I was trying to make this more clear but I failed. Will add the comment.

src/disc/stk/Albany_ExtrudedSTKMeshStruct.cpp
mcarlson801: When I was digging around the mesh specs it looked like there are two sidesets called lateralside. One is the 2d boundary of the fully extruded 3d mesh and the other is attached to basalside's mesh specs and appears to be the 1d boundary of the basalside mesh. So it looks like it gets stored but I don't know if it gets used at all by any evaluators since it has the same sideset name as lateralside.

src/disc/Albany_DiscretizationFactory.cpp
jewatkins: I don't know of any other cases but maybe @mperego knows. If not, I could clean this up.

src/disc/Albany_DiscretizationFactory.cpp
jewatkins: The idea here was to set the workset size of basalside to be -1 if the extruded workset size is -1. That's not very clear here, I'll be more explicit.

src/disc/stk/Albany_ExtrudedSTKMeshStruct.cpp
jewatkins: Yes, I'm using the workset size from the 1D sides of the basal mesh to set the lateral workset sizes. In the case that the basalside uses Ioss, we always stored the mesh specs of the 1D sides. I just added the correct workset sizes. I'm not exactly sure if the mesh is built on the side but I think it is since I see prints for STKDisc: sideset [1D sideset name]. I was also wondering if that was even correct...

bartgol: Mind=blown.

mperego: Note sure it's what you are talking about, but we do that to give the possibility of having different lateral side sets, e.g. a lateral side where the ice is floating, one where it's grounded, one where we prescribe Dirichlet conditions. So you'll have sides like `extruded_boundary_side_set_1`, `extruded_boundary_side_set_2`, etc. We always have `lateralside` that contains all the lateral sides.

mperego (commented):
It looks good. What happens with multiple worksets?
It would be good to set in input file the max workset size for sidesets needed. Would it be hard for the code to print to screen what would be the minimum workset size needed?

src/disc/stk/Albany_ExtrudedSTKMeshStruct.cpp
bartgol: What I'm saying is that I did not know that we built a mesh specs for the basal mesh side sets. I thought we only stored those sidesets as side sets, without the extra info coming with a mesh specs.

jewatkins:
> It looks good. What happens with multiple worksets?
Everything is skipped when there's multiple worksets so it will use the large allocation - i.e. whatever we usually do. So the top level discretization `worksetSize` * `numSides` on a 3D element. Max and I spent a long time trying to get the workset size and bucket sizes to be consistent so that we could get this case to work too but we kept getting errors. We talked to Roger and there's a way to delay the layout sizes until after we construct the discretization (which would solve all issues) but it might be intrusive. Max is looking into it now.

> It would be good to set in input file the max workset size for sidesets needed.
Do you mean setting the maximum size manually? e.g. If the basalside has 100 cell total, you want to be able to set `Workset Size: 100` in the input file and still get the slim memory allocation?

> Would it be hard for the code to print to screen what would be the minimum workset size needed?
Do you mean, print the maximum size of sidesets all the time? Yes, the code already prints the maximum sizes when it prints all the `STKDisc: sideset...`.

Hmm, one extra thing I thought about is that currently, you have to set the top extruded discretization workset size to -1 in order to get the slim allocations on the lateral/upper sides. I think I can relax that so that you can have multiple worksets on the top discretization but a single workset on all sidesets. Would that be useful?

mperego:
> > It would be good to set in input file the max workset size for sidesets needed.

> Do you mean setting the maximum size manually? e.g. If the basalside has 100 cell total, you want to be able to set Workset Size: 100 in the input file and still get the slim memory allocation?

Yes. Since we know the maximum size after running it once, it would be nice to set it manually.

mperego:
> Hmm, one extra thing I thought about is that currently, you have to set the top extruded discretization workset size to -1 in order to get the slim allocations on the lateral/upper sides. I think I can relax that so that you can have multiple worksets on the top discretization but a single workset on all sidesets. Would that be useful?

Not sure.. I would say to not bother about that.

src/disc/stk/Albany_ExtrudedSTKMeshStruct.cpp
jewatkins: Yeah, I just realized this too as I was working this. I imagine what happened was the general initialization of sidesets was put into Ioss for restart and 3D meshes but the consequence of that is that sideset mesh specs are constructed for any sideset that uses Ioss too.

jewatkins:
Okay I think I can check to see if the maximum size is set too. And I'll just go ahead and relax the checks for a single workset on sidesets too. I think it makes more sense to do that.

PR #694, opened 2021-05-28
Add test for StokesFOHydrology problem

bartgol:
This PR fixes a few leftover issues with the StokesFOHydrology problem, and adds a unit test for it.

Among the issues fixed/added (more info in the proper commits):

* Bug in ScatterSideEqnResidual: the side nodes were not gathered correctly
* Ensured flowRate/iceSoftness is consistent between ice and hydrology (both in code and input files specs)
* Store basal nodesets in Extruded meshes (for hydro DBC)
* Bug fix (plus cleanup) in STK meshes, related to passing parameter lists
* Simplified STK field container helpers, by using Kokkos View's instead of Shard Arrays.
* Uniform the way sideset name is attached to field names: always use fname + "_" + sideName, even for BFs and other common fields.
* When doing side fields gradients, name them as fname + "_gradient_" + sideName. This allows to request them as output fields.
* Refactor Save/Load side state: they assume stk mesh now, and use directly stk utils to retrieve data (rather than passing through the disc stateArrays structs).
* Improved regularization of WaterDischarge: like FrobeniusNorm ev, can now regularize in different ways.
* Removed "Homotopy" param from LandIce_ParamEnum.hpp: always use "GLHomotopyParam". We might rename this to 'HomoropyParam" later, since it starts to be used everywhere (not just in Glen's Law).
* SideSet layouts-related fixes in hydrology-related classes.
* Improved handling of Albany_config.h: allows to avoid rebuilding all  Albany when the file doesn't change after a CMake run.
* Added helper functions to StokesFOBase, to handle name of fields on a side.


bartgol:
@mperego I think this PR is ready to be merged. I'm running the SHMIP tests to take a look at the solution, and ensure it still looks right, then I'm setting up regression tests for those (as you suggested). But meanwhile, I think you can already review this.

Sorry, it's a relatively big PR.

bartgol:
@mperego  Ok, I found a mini bug, then checked again the solutions, and they all look right.

src/evaluators/scatter/PHAL_ScatterSideEqnResidual_Def.hpp
```diff
@@ -113,29 +113,84 @@ postRegistrationSetup(typename Traits::SetupData d,
   d.fill_field_dependencies(this->dependentFields(),this->evaluatedFields());
 }
 
+template<typename EvalT, typename Traits>
```
mperego, line 116: It's probably safer to error out if the mesh is not extruded.

tests/small/LandIce/FO_Hydrology/input_slab_steadyT.yaml
```diff
@@ -181,6 +180,10 @@ ANONYMOUS:
             Field Name: sliding_velocity
             Field Type: Node Scalar
             Field Usage: Output
+          Field 10:
+            Field Name: water_pressure
+            Field Type: Node Scalar
+            Field Usage: Output
   Piro:
```
mperego, line 187: No Regression sublist in this test, can you add it?

tests/small/LandIce/SHMIP/input_bueler_unsteady.yaml
```diff
@@ -52,7 +52,8 @@ ANONYMOUS:
         Water Thickness Exponent: 1.0
         Potential Gradient Norm Exponent: 2.0
         Transmissivity: 1.0e-06
-        Regularize With Continuation: false
+        Regularization:
```
mperego, line 55: Can you add a regression sublist?

tests/small/LandIce/SHMIP/input_suite_A6.yaml
```diff
@@ -57,7 +57,9 @@ ANONYMOUS:
         Lump Mass: false
         Use Melting: false
       Darcy Law:
-        Regularize With Continuation: true
+        Regularization:
```
mperego, line 60: Can you add a Regression test?

mperego (commented):
Lot of work! Thanks @bartgol! Looks good overall. 
I have a few comments. 
DISCLAIMER: I didn't check carefully the part related to the mesh and side sets.. it would take me too much time to fully understand the changes.

Also, can you update the PR description  summurizing what you do? (copying and pasting from commit messages would is fine)

CMakeLists.txt
```diff
@@ -444,51 +436,69 @@ IF (MueLu_List_ID GREATER -1)
...
 ENDIF()
 
-# Check if Trilinos has Tempus
-LIST(FIND Trilinos_PACKAGE_LIST Tempus Tempus_List_ID)
-IF (Tempus_List_ID GREATER -1)
-  MESSAGE("-- Looking for Tempus (required) in Trilinos:     Found")
-  SET(ALBANY_TEMPUS TRUE)
+# CHeck if Trilinos has Thyra-Epetra adapters
+LIST(FIND Trilinos_PACKAGE_LIST ThyraEpetraAdapters ThyraEpetraAdapters_ID)
+IF (ThyraEpetraAdapters_ID GREATER -1)
```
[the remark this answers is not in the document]
bartgol: I figured we'd put all the checks for trilinos packages in one place. But I don't feel really strongly about this, so I can roll it back (i.e., put the check inside the check for albany epetra.

CMakeLists.txt
```diff
@@ -507,11 +517,8 @@ IF (ENABLE_ALBANY_EPETRA)
   MESSAGE("-- ALBANY_EPETRA                           Enabled.")
   SET(ALBANY_EPETRA TRUE)
 
-  LIST(FIND Trilinos_PACKAGE_LIST ThyraEpetraAdapters ThyraEpetraAdapters_ID)
-  IF (ThyraEpetraAdapters_ID GREATER -1)
-    MESSAGE("  -- Looking for ThyraEpetraAdapters:      Found.")
-  ELSE()
-    MESSAGE("  -- Looking for ThyraEpetraAdapters:  NOT found.")
+  # If Epetra build is requested, the ThyraEpetraAdapter in Trilinos is no longer optional
```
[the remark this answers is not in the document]
bartgol: Yeah, but I moved the check for the epetra adapters inside the optional trilinos packages. I think it was probably a bad mod. I will bring the check for the adapters back into this section.

src/LandIce/evaluators/hydrology/LandIce_HydrologyResidualCavitiesEqn_Def.hpp
```diff
@@ -174,16 +172,15 @@ evaluateFieldsSide (typename Traits::EvalData workset)
 
   if (workset.sideSets->find(sideSetName)==workset.sideSets->end()) return;
 
+  // Note: the '1e9' is to convert the ice softness in kPa^-3 s^-1, so that the kPa
```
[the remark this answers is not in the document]
bartgol: True. Then I should also change 1e9 to `pow(1000,n)`.

src/LandIce/problems/LandIce_StokesFOBase.hpp
```diff
@@ -1463,16 +1457,20 @@ void StokesFOBase::constructSMBEvaluators (PHX::FieldManager<PHAL::AlbanyTraits>
...
-    std::string ice_thickness_side_name = ice_thickness_name + "_" + basalSideName;
-    std::string ice_thickness_side_name_planar = ice_thickness_name + "_" + basalSideNamePlanar;
-    std::string surface_height_side_name = surface_height_name + "_" + basalSideName;
-    std::string apparent_mass_balance_side_name = "apparent_mass_balance_" + basalSideName;
-    std::string apparent_mass_balance_RMS_side_name = "apparent_mass_balance_RMS_" + basalSideName;
-    std::string stiffening_factor_side_name = stiffening_factor_name + "_" + basalSideName;
-    std::string effective_pressure_side_name = effective_pressure_name + "_" + basalSideName;
-    std::string vertically_averaged_velocity_side_name = vertically_averaged_velocity_name + "_" + basalSideName;
-    std::string bed_roughness_side_name = "bed_roughness_" + basalSideName;
+    auto bname = [&] (const std::string& fname) -> std::string {
```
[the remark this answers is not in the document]
bartgol: Yeah, sname and bname are probably not very verbose. I will make them more verbose. And I can turn bname (renamed) into a method, since derived classses might use it too. And add a `surf_fname` method too.

src/evaluators/scatter/PHAL_ScatterSideEqnResidual_Def.hpp
bartgol: True.

tests/small/LandIce/FO_Hydrology/input_slab_steadyT.yaml
bartgol: Ah, this input file got committed, but there's not actual test added. Since I haven't really checked this input file "works", I might simply remove it.

tests/small/LandIce/SHMIP/input_bueler_unsteady.yaml
bartgol: Ah, again, not tested. Since this is the only unsteady input file, I should probably make sure it works (i.e., the solution after a long time is similar to the steady state one), and then add a test.

tests/small/LandIce/SHMIP/input_suite_A6.yaml
bartgol: This test did not converge (and according to Matt, this test case kind of requires channels, so it's ok that it didn't). I will remove the input file, to avoid confusion.

mperego (approved):
@bartgol  Looks good to me!

PR #695, opened 2021-06-02
Use SI units for Glen's Law A parameter

bartgol:
The value of A in the input files for LandIce was in some funky units: `[k^-1 kPa^-n yr^-1]`. The reason was that these units perfectly canceled the ones coming from the fact that our meshes are in km, and our velocities in m/yr.

However, A is also needed in the Subglacial Hydrology problem, and the natural units there are different. In particular, the natural units for A in the hydrology would be `[ kPa^-n s^-1]`.

To keep things easy to understand, after talking with @mperego , we decided to settle for SI units: `[ Pa^-n s^-1]`. All evaluators can convert to whatever is needed in the particular equation to cancel out units and scaling factors.

I modified all input files to use these units, and a handful of evaluators to do the conversion back to what they were expecting before. Also, I modified input files to eliminate double quotes. E.g., I turned `'Glen''s Law A': xyz` into `Glen's Law A: xyz`.

### DIFF
--- a/src/LandIce/evaluators/LandIce_FlowRate_Def.hpp
+++ b/src/LandIce/evaluators/LandIce_FlowRate_Def.hpp
@@ -79,13 +79,10 @@ void FlowRate<EvalT, Traits, TempST>::evaluateFields (typename Traits::EvalData 
       constexpr double switchingT (263.15);   // [K]
       constexpr double arrmlh (1.733e3);      // [Pa-3 s-1]
       constexpr double arrmll (3.613e-13);    // [Pa-3 s-1]
-      constexpr double k4scyr (3.1536e19);    // [s y-1]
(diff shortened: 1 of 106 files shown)


mperego (approved):
Looks good. Thanks!!

bartgol:
I get some failures in the full Stokes tests. Once I fix those, I'll merge.

mperego:
@bartgol can you fix also the mpas interface? It should be easy. Just scale approproately when `Glen's Law A` is set.

bartgol:
Ok, all tests now pass on my laptop. I also modified the MPAS interface @mperego can you double check that at some point? Otherwise, I can merge to master, and let the nightlies test this (I don't recall if we have MALI nightlies though...).

mperego:
Go ahed and merge it. I'll double check that with MPAS together with other changes I'm doing later on.

PR #696, opened 2021-06-03
Refactor Of Basal BCs

mperego:
(Edited after review)
The Basal BCs have been refactored to harmonize the different boundary conditions and to add new capabilites.
There are now four main basal BCs: `Constant`, `Field`, `Power Law` and `Regularized Coulomb`.  As before, for `Constant`, beta is a constant read input file.
In the  `Power Law` case, `beta = mu N |u|^{q-1}`, where `N` is the effective pressure. The linear condition is now a special case of `Power Law` with `N =1` and `q=1`.
The `Field` case is implemented as `Power Law`, qith `q=1`, `N=1`. 
In the `Regularized Coulomb`, `beta = mu N |u|^{q-1} /(|u| + lambda A N^n)^q`, where `lambda` is the bed roughness. `mu`, `N` and `q` are shared by the `Power Law` and `Regularized Coulomb`. 
`mu` and `lambda` can be `Constant`, `Field`, `Exponent Of Field` and `Exponent Of Field At Nodes`. When `Exponent Of Field At Nodes` is selected, the Field is always exponentiated at the nodes, and then interpolated at the Quad Points if needed.
`N` can be `Constant`, `Field`, `Hydrostatic`, `Hydrostatic Computed At Nodes`. In the hydrostatic case, Albany computes N using the formula `N = max( rho_i H + min(rho_w b, 0.0), 0.0)`, where `rho_i` and `rho_w` are the ice and water densities, `H` the ice thickness and `b` the bed topography.
We have two options to set beta to zero on floating ice, hopefully we'll be able to settle on one in the future. One is `Zero Effective Pressure On Floating Ice At Nodes`, which does what it says, the other is the usual `Zero Beta On Floating Ice` which zeros beta either at nodes (in case of a nodal field) or at quad points.

Typical cases
Linear:
```
        Basal Friction Coefficient:
          Type: Field
          Beta Field Name: basal_friction
```
Power Law with Effective pressure computed by Albany
```
        Basal Friction Coefficient:
          Type: Power Law
          Mu Field Name: mu
          Effective Pressure Type: Hydrostatic
          Power Exponent: 0.33333333333333
```
Schoof law with given Effective Pressure given 
```
        Basal Friction Coefficient:
          Type: Regularized Coulomb
          Mu Field Name: mu
          Bed Roughness Type: Uniform
          Bed Roughness: 1.0
          Effective Pressure Type: Field
          Power Exponent: 0.33333333333333
```

There are also a bunch of random modifications, in particular after chatting with Luca I simplified the way SharedParameter work and make sure we register only the "real" parameters (those we want to compute sensitivities w.r.t.). 

Consolidated some of the melting pressure evaluators (we have one less now) and removed the BasalFrictionGradient evaluator as we don't really need it.


jewatkins:
```
Basal Friction Coefficient:
  Type: Constant
  Beta: 1.0
```
makes sense to me.

jewatkins:
Changing `Uniform` to `Constant` also would be nice but not necessary.

mperego:
> Changing `Uniform` to `Constant` also would be nice but not necessary.

That's easy enough

src/LandIce/evaluators/LandIce_StokesFOBasalResid_Def.hpp
```diff
@@ -113,19 +103,24 @@ operator() (const StokesFOBasalResid_Tag& tag, const int& sideSet_idx) const {
...
-      for (unsigned int qp=0; qp<numSideQPs; ++qp) {
-        local_res[dim] += (ff + beta(sideSet_idx,qp)*u(sideSet_idx,qp,dim))*BF(sideSet_idx,node,qp)*w_measure(sideSet_idx,qp);
-      }
-      Kokkos::atomic_add(&residual(cell,sideNodes(side,node),dim), local_res[dim]);
+
+    for (unsigned int qp=0; qp<numSideQPs; ++qp) {
+      MeshScalarT bx = -normals(sideSet_idx,qp,0)/normals(sideSet_idx,qp,2);
+      MeshScalarT by = -normals(sideSet_idx,qp,1)/normals(sideSet_idx,qp,2);
+      // removed the following line after checking performance
+      bx=0.0; by=0.0;
```
jewatkins, line 116: Remove?

src/LandIce/interface_with_mpas/Interface.cpp
```diff
@@ -223,12 +204,13 @@ void velocity_solver_solve_fo(int nLayers, int globalVerticesStride,
...
-    if (il == 0 && basalFrictionField!=nullptr) {
-      double* beta = stk::mesh::field_data(*basalFrictionField, node);
-      beta[0] = betaData[ib];
-    }
 
-    if (!muData.empty() && (muField != nullptr)) {
+    //if (il == 0 && muField!=nullptr) {
+    //  double* mu = stk::mesh::field_data(*muField, node);
+    //  mu[0] = betaData[ib];
+    //}
```
jewatkins, line 211: Remove?

src/LandIce/problems/LandIce_StokesFOBase.cpp.orig
```diff
@@ -0,0 +1,752 @@
...
+#include "Albany_GeneralPurposeFieldsNames.hpp"
+#include "Teuchos_CompilerCodeTweakMacros.hpp"
+
+#include <string.hpp>               // For util::upper_case (do not confuse this with <string>! string.hpp is an Albany file)
+#include <Albany_ProblemUtils.hpp>  // For 'getIntrepidwBasis'
+
+namespace LandIce {
+
+StokesFOBase::
+StokesFOBase (const Teuchos::RCP<Teuchos::ParameterList>& params_,
```
jewatkins, line 11: Remove this file?

src/LandIce/problems/LandIce_StokesFOBase.hpp
```diff
@@ -30,7 +30,7 @@
 #include "LandIce_ResponseUtilities.hpp"
 
 #include "LandIce_BasalFrictionCoefficient.hpp"
-#include "LandIce_BasalFrictionCoefficientGradient.hpp"
+//#include "LandIce_BasalFrictionCoefficientGradient.hpp"
```
jewatkins, line 33: Remove?

jewatkins (commented):
Looks good, probably just needs cleanup.

src/LandIce/interface_with_mpas/Interface.cpp
mperego: will do

src/LandIce/evaluators/LandIce_StokesFOBasalResid_Def.hpp
mperego: I'd like to check whether requiring the normal computations significantly impacts the performances. Removing that line will make several tests fail and we'll have reset the regression results.

src/LandIce/problems/LandIce_StokesFOBase.cpp.orig
mperego: Oh, good catch!

src/LandIce/problems/LandIce_StokesFOBase.hpp
mperego: sure

src/LandIce/evaluators/LandIce_StokesFOBasalResid_Def.hpp
jewatkins: Ah okay got it.

src/LandIce/evaluators/LandIce_BasalFrictionCoefficient_Def.hpp
```diff
@@ -242,115 +303,130 @@ KOKKOS_INLINE_FUNCTION
...
-            beta(cell,ipt) = std::exp(given_field_param(cell,ipt));
-        } else {
-          for (unsigned int ipt=0; ipt<dim; ++ipt)
-            beta(cell,ipt) = std::exp(given_field(cell,ipt));
+
+      switch (mu_type) {
+      case FIELD_TYPE::FIELD:
+        muValue = muField(cell,ipt);
+        break;
+      case FIELD_TYPE::EXPONENT_OF_FIELD_AT_NODES:
```
bartgol, line 367: Is this option really something we want? I think it's easier (and probably more accurate) to interpolate mu at the quadrature points, and then exponentiate, no?

src/LandIce/evaluators/LandIce_BasalFrictionCoefficient_Def.hpp
```diff
@@ -242,115 +303,130 @@ KOKKOS_INLINE_FUNCTION
...
+        else {
+          MeshScalarT thickness(0), bed_topo(0);
+          for (unsigned int node=0; node<numNodes; ++node) {
+            thickness += thickness_field(cell,node)*BF(cell,node,ipt);
+            bed_topo += bed_topo_field(cell,node)*BF(cell,node,ipt);
           }
+          Nval = g*KU::max(rho_i*thickness+KU::min(rho_w*bed_topo,0.0),0.0);
+        }
+        break;
+      case EFFECTIVE_PRESSURE_TYPE::HYDROSTATIC_AT_NODES:
```
bartgol, line 352: Why computing N at nodes and then interpolating at qp's? Why not computing N directly at the qp's? It seems the latter is more accurate in general.

src/LandIce/evaluators/LandIce_BasalFrictionCoefficient_Def.hpp
```diff
@@ -242,115 +303,130 @@ KOKKOS_INLINE_FUNCTION
...
-            ScalarT q = u_norm(cell,ipt) / ( u_norm(cell,ipt) + lambda*ice_softness(cell)*std::pow(Nval,n) );
-            beta(cell,ipt) = mu * Nval * std::pow( q, power) / u_norm(cell,ipt);
+        break;
+      case EFFECTIVE_PRESSURE_TYPE::HYDROSTATIC:
+        if(nodal)
+          Nval = g*KU::max(rho_i*thickness_field(cell,ipt)+KU::min(rho_w*bed_topo_field(cell,ipt),0.0),0.0);
+        else {
+          MeshScalarT thickness(0), bed_topo(0);
+          for (unsigned int node=0; node<numNodes; ++node) {
+            thickness += thickness_field(cell,node)*BF(cell,node,ipt);
```
bartgol, line 346: We almost surely already compute these fields at the quadrature points, so why not simply asking for thickness and bed topo as QuadPoints MDField's?

src/LandIce/evaluators/LandIce_BasalFrictionCoefficient_Def.hpp
bartgol: Ah, wait: is the reasoning that mu might be the log of the field, and you would like to get the interpolation of the actual nodal mu, rather than `exp(interp(log(mu)))` ? If that's the case, then I would get rid of EXPONENT_OF_FIELD. I mean, what's the use of it?

src/LandIce/evaluators/LandIce_BasalFrictionCoefficient_Def.hpp
```diff
@@ -242,115 +303,130 @@ KOKKOS_INLINE_FUNCTION
...
+          if(nodal)
+            lambdaValue = std::exp(lambdaField(cell,ipt));
+          else {
+            lambdaValue = 0;
             for (unsigned int node=0; node<numNodes; ++node)
-              beta(cell,ipt) += std::exp(given_field(cell,node))*BF(cell,node,ipt);
+              lambdaValue += std::exp(lambdaField(cell,node))*BF(cell,node,ipt);
           }
+          break;
+        case FIELD_TYPE::EXPONENT_OF_FIELD:
```
bartgol, line 404: Same consideration as for mu: we could keep just one between EXPONENT_OF_FIELD and EXPONENT_OF_FIELD_AT_NODES, and always do interp(exp(f_nodal)).

src/LandIce/evaluators/LandIce_BasalFrictionCoefficient_Def.hpp
```diff
@@ -242,115 +303,130 @@ KOKKOS_INLINE_FUNCTION
...
-        beta(cell,ipt) *=  isGrounded;
-      }
-    } else {
-      for (unsigned int ipt=0; ipt<dim; ++ipt) {
-        ParamScalarT isGrounded = rho_i*thickness_field(cell,ipt) > -rho_w*bed_topo_field(cell,ipt);
-        beta(cell,ipt) *=  isGrounded;
+    for (unsigned int ipt=0; ipt<dim; ++ipt) {
+      MeshScalarT isGrounded;
+      if(nodal)
+        isGrounded = rho_i*thickness_field(cell,ipt) > -rho_w*bed_topo_field(cell,ipt);
```
bartgol, line 420: I'm confused on how this would work with Fad types: isGrounded is morally a bool, but how does `op >` handle the derivatives? If they are zero, then we might be better off doing
```
if (rhoi*H>-rhow*z_b)
   beta = 0.0;
```
rather than use `op *=`, which has to do more stuff (b/c of product rule in the Fad dx).

src/LandIce/evaluators/LandIce_BasalFrictionCoefficient_Def.hpp
```diff
@@ -242,115 +303,130 @@ KOKKOS_INLINE_FUNCTION
...
+      else {
+        MeshScalarT thickness(0), bed_topo(0);
+        for (unsigned int node=0; node<numNodes; ++node) {
+          thickness += thickness_field(cell,node)*BF(cell,node,ipt);
```
bartgol, line 424: As I wrote in another comment, can't we assume we have H and z_b already at qp's?

src/LandIce/evaluators/LandIce_BasalFrictionCoefficient_Def.hpp
```diff
@@ -373,15 +449,15 @@ evaluateFields (typename Traits::EvalData workset)
   if (memoizer.have_saved_data(workset,this->evaluatedFields()))
     return;
 
-  if (beta_type == POWER_LAW) {
+  if ((beta_type == POWER_LAW)||(beta_type==REGULARIZED_COULOMB) ) {
```
bartgol, line 452: You might just add a `|| beta_type==GIVEN_CONSTANT` in the previous if statement, and keep the code below less indented.

src/LandIce/CMakeLists.txt
```diff
@@ -8,7 +8,6 @@
 
 set(SOURCES
   evaluators/LandIce_BasalFrictionCoefficient.cpp
-  evaluators/LandIce_BasalFrictionCoefficientGradient.cpp
```
bartgol, line 11: This evaluator was used for the regularization term inside SurfaceVelocityMismatch. Are we sure it's not needed anymore? (Note: I'm happy if it can be removed, just double checking).

src/LandIce/evaluators/LandIce_LiquidWaterFraction.hpp
```diff
@@ -26,13 +26,13 @@ namespace LandIce
   This evaluator computes the liquid water fraction in temperate ice
  */
 
-template<typename EvalT, typename Traits, typename Type>
+template<typename EvalT, typename Traits, typename EntalpyType>
```
bartgol, line 29: Minor: `Entalpy`->`Enthalpy`. Also, we could be more verbose, and use `EnthalpyST` or `EnthalpyScalarT`, to emphasizes the template arg is the _scalar type_ of the enthalpy (not the "type of enthalpy", which I'm not sure what would mean). Also, `BlahST` is somewhat the common name used in other evaluators for the scalar type of field Blah.

src/LandIce/problems/LandIce_StokesFOThermoCoupled.hpp
```diff
@@ -303,9 +302,7 @@ constructEnthalpyEvaluators (PHX::FieldManager<PHAL::AlbanyTraits>& fm0,
   //Output
   p->set<std::string>("Basal Vertical Velocity Variable Name", "basal_vert_velocity_" + basalSideName);
   p->set<std::string>("Basal Melt Rate Variable Name", "basal_melt_rate_" + basalSideName);
-  ev = createEvaluatorWithTwoScalarTypes<BasalMeltRate,EvalT>(p,dl->side_layouts[basalSideName],
-                                                                         FST::Scalar,
-                                                                         get_scalar_type(melting_enthalpy_name));
+  ev = createEvaluatorWithOneScalarType<BasalMeltRate,EvalT>(p,dl->side_layouts[basalSideName], FST::Scalar);
```
bartgol, line 305: Since the scalar type is known at compile time, you might be better off doing
```
ev = Teuchos::rcp(new BasalMeltRate<EvalT, PHAL::AlbanyTraits, typename EvalT::ScalarT>(*p,dl);
```
like you do below for LiquidWaterFraction. It should be faster to compile.

src/LandIce/problems/LandIce_StokesFOThermoCoupled.hpp
```diff
@@ -430,7 +412,7 @@ constructEnthalpyEvaluators (PHX::FieldManager<PHAL::AlbanyTraits>& fm0,
   //Output
   p->set<std::string>("Residual Variable Name", resid_names[2]);
 
-  ev = createEvaluatorWithTwoScalarTypes<EnthalpyResid,EvalT>(p,dl,FST::Scalar,get_scalar_type(melting_temperature_name));
+  ev = createEvaluatorWithOneScalarType<EnthalpyResid,EvalT>(p,dl,FST::Scalar);
```
bartgol, line 415: Here too: if the scalar type is known at compile time, you save compilation time by not using the `createEvaluatorWithOneScalarType` helper function.

src/evaluators/utility/PHAL_SharedParameter.hpp
```diff
@@ -32,8 +32,61 @@ class SharedParameter : public PHX::EvaluatorWithBaseImpl<Traits>,
...
     Teuchos::RCP<ParamLib> paramLib = p.get<Teuchos::RCP<ParamLib>>("Parameter Library");
-    this->registerSacadoParameter(param_name, paramLib);
     this->setName("Shared Parameter " + param_name + PHX::print<EvalT>());
+
+    const Teuchos::ParameterList* paramsList = p.get<const Teuchos::ParameterList*>("Parameters List");
+
+    // Find the parameter in the Paramter list,
+    // register as a Sacado Parameter and set the Nominal value
+    bool nominalValueSet = false;
+    if((paramsList != NULL) && paramsList->isParameter("Number Of Parameters"))
```
bartgol, line 42: Very minor: in C++ is more apt to use `nullptr` instead of the C-style `NULL`.

tests/small/LandIce/FO_AIS/inputML1.yaml
```diff
@@ -20,8 +20,9 @@ ANONYMOUS:
         Cubature Degree: 3
         Side Set Name: basalside
         Basal Friction Coefficient:
-          Type: Given Field
-          Given Field Variable Name: basal_friction
+          Type: Power Law
```
bartgol, line 23: I wonder if we should keep `Given Field` (or 'Prescribed Field' maybe) as a valid option. We could simply have:
```
Basal Friction Coefficient:
  Type: Prescribed Field
  Field Name: basal_friction
```
Internally, we could still treat it as a Power Law with exponent=1.0, N=One, and muType=FIELD. But the input file would probably be less confusing for those interested in prescribing a field. It is kind of silly to have to specify a power exponent just so that we get rid of the power law altogether...

tests/small/LandIce/SHMIP/input_suite_A1.yaml
```diff
@@ -65,6 +65,8 @@ ANONYMOUS:
         Type: Given Field
     LandIce Basal Friction Coefficient: 
       Type: Regularized Coulomb
+      Mu Type: Uniform
```
bartgol, line 68: Off topic: this is a somewhat silly thing I did: the Hydrology pb always builds the BasalFrictionCoefficient evaluator, so this sublist needs to be there. However, these tests are not using the evaluator. So maybe I should (at some point) change the Hydrology pb, so that it doesn't build the ev if basal melting is not used (so that I can remove these sublists).

bartgol (commented):
Nice work! I especially like how you reduced the number of template args in a few LandIce evaluators. That will definitely help with compilation times.

I have a few questions/suggestions, see inline comments.

src/LandIce/evaluators/LandIce_BasalFrictionCoefficient_Def.hpp
mperego: That's exactly right. And I'd like to get rid of EXPONENT OF FIELD, just wanted to keep backword compatibility for now, as that's the one we use in most of the tests

src/LandIce/evaluators/LandIce_BasalFrictionCoefficient_Def.hpp
mperego: I'd like that if you compute the hydrostatic effective pressure in MPAS, on nodes, and then pass it to Albany, you get the same result as computing that with Albany using `HYDROSTATIC_AT_NODES`. Also, if N is computed at quadrature points, it will be a nonsmooth field at the GL, which makes it hard to be integrated.

src/LandIce/evaluators/LandIce_BasalFrictionCoefficient_Def.hpp
mperego: To avoid storing both thickness_field and thickness_field_at_nodes (and same for bed_topo), given that in some cases you would need both.

src/LandIce/evaluators/LandIce_BasalFrictionCoefficient_Def.hpp
mperego: In this case we could get rid of it, as we are not breaking any backward compatibility (lambda has never been used as exponent of a field)

src/LandIce/evaluators/LandIce_BasalFrictionCoefficient_Def.hpp
mperego: That's my guess to. OK I'll rewrite that as if condition.

src/LandIce/problems/LandIce_StokesFOThermoCoupled.hpp
mperego: It makes sense, will do.

src/LandIce/problems/LandIce_StokesFOThermoCoupled.hpp
mperego: OK

src/evaluators/utility/PHAL_SharedParameter.hpp
mperego: OK, I'll fix that.

tests/small/LandIce/FO_AIS/inputML1.yaml
mperego: We could, but we'll be adding yet another case. I think Matt and Trevor are fine always using muField also in the linear case.

mperego: As you say it would be trivial to implement. But it would be used only in standalone Albany anyway, as that condition is hardcoded as the default one in the MPAS Interface as well.

src/LandIce/evaluators/LandIce_BasalFrictionCoefficient.hpp
```diff
@@ -55,14 +55,15 @@ class BasalFrictionCoefficient : public PHX::EvaluatorWithBaseImpl<Traits>,
...
+  double beta_val;   // beta value [kPa yr/m] (for CONSTANT only)
+  MeshScalarT N_val; // effective pressure value [kPa]
 
   // Input:
   PHX::MDField<const RealType>          BF;
   PHX::MDField<const VelocityST>        u_norm;          // [m yr^{-1}]
   PHX::MDField<const ParamScalarT>      lambdaField;     // [km], characteristic length
   PHX::MDField<const ParamScalarT>      muField;         // [yr^q m^{-q}] or [adim], Power Law with exponent q, Coulomb Friction
-  PHX::MDField<const EffPressureST>     N;               // [kPa]
+  PHX::MDField<const MeshScalarT>       N;               // [kPa]
```
bartgol, line 66: For hydrology-coupled simulations, you do need N to be ScalarT. That's why I had the EffPressureST template arg. Does this code compile for you?

mperego:
@bartgol @jewatkins I have addressed your comments. I also added the case `Type: Field` that is implemented as `Power Law` with `q=1` and `N=1`.

src/LandIce/evaluators/LandIce_BasalFrictionCoefficient.hpp
mperego: Oh.. right.. yeah, let me fix this. It does compile though

bartgol: Ah, I wonder how/why. This line in LandIce_Hydrology.hpp should cause an error:

```
ev = Teuchos::rcp(new BasalFrictionCoefficient<EvalT,PHAL::AlbanyTraits,ScalarT,RealType,RealType>(*p,dl));
```

jewatkins (approved):

src/LandIce/evaluators/LandIce_BasalFrictionCoefficient.hpp
mperego: I think I had changed that line. Anyway, I reverted the change

PR #699, opened 2021-06-16
Removing 'is_adjoint' options

ikalash:
Per discussion with @mperego , removing "is_adjoint" options, as these are obsolete / not used, and just cause confusion.  @mperego , can you please approve?


mperego (approved):
It looks good Irina. Thanks for the cleaning!

Issue #703, opened 2021-07-07
BlockDiscretization_Jacobian_Unit_Test failing on machines w/o numpy/scipy


ikalash:
It looks like the recently-added BlockDiscretization_Jacobian_Unit_Test is failing on machines w/o numpy/scipy (blake, skybridge, mockba): https://sems-cdash-son.sandia.gov/cdash/index.php?project=Albany .   One fix is to disable the test if numpy/scipy are not found.  @kliegeois is looking into this.

Question for @jewatkins : is there a way to get numpy/scipy through modules on blake (and similarly skybridge)?  It seems strange to me that there is a python module but not one for numpy/scipy.  Maybe I'm just not finding it?

jewatkins:
`module load python/3.7.3` seems to already have numpy, scipy on blake but I might have installed it locally. You could try:
```
module load python/3.7.3
pip install numpy --user
pip install scipy --user
```
I'm not sure about the other clusters. It sometimes won't work (even locally) due to permissions depending on the package. @mcarlson801 is going to look into getting blake/weaver working with other python packages for autotuning so he may already have more information

kliegeois:
I just finished the modification. I am testing in on Blake and I create the PR.

kliegeois:
> `module load python/3.7.3` seems to already have numpy, scipy on blake but I might have installed it locally. 

@jewatkins I just verified, it comes with numpy but not scipy (which can be installed locally either with pip as you said or with  easy_instal).

jewatkins:
okay, @ikalash you could probably add something like what's suggested [here ](https://askubuntu.com/a/1211956) to the blake module file to ensure these packages are always installed.

ikalash:
@jewatkins : thanks, your suggestion worked!

ikalash:
Just for documentation (i.e., so I don't forget), to install libs using pip on some of the SRN machines and bypass proxies/SSL certificates, one needs to execute the following command:

```pip3 install numpy --trusted-host=pypi.org --trusted-host=files.pythonhosted.org --user```

kliegeois:
If I may add, on IBM machines (at least weaver), you need to specify a bit older numpy (such as 1.19.2) otherwise, when building up to date numpy, it fails to compile.

jewatkins:
Thanks for the suggestions. It should help for future python workflow.

kliegeois:
Just for documentation, on Weaver, to install matplotlib we need to install pillow as follows:
[...]

PR #704, opened 2021-07-07
Add checks not to run BlockDiscretization_Jacobian_Unit_Test

kliegeois:
If the PYTHON_EXECUTABLE is not sufficiently new or does not support sufficiently new numpy or scipy, these CMAKE modifications prevent from running the test BlockDiscretization_Jacobian_Unit_Test.

This is related to issue #703.

EDIT: I should specify that the versions of python, scipy, and numpy listed are the oldest ones with which I tried the test. The test can work with older versions but it has not been tested. If needed, those can be changed.


ikalash (approved):
This does exactly what I was thinking - thanks for getting this in so fast @kliegeois !

PR #709, opened 2021-07-09
Fix coordinates passed into ML libs and used to compute nullspace

jewatkins:
The coordinates used by ML/MueLu are translated by the centroid (see subtractCentroid()). This function had a bug where the local number of nodes was used instead of the global number of nodes:
https://github.com/SNLComputation/Albany/compare/jewatkins/fix-ml-coords?expand=1#diff-fba75aeba9c71d8c94b90f32551a058c6e45012508a5ab2399a379066f9b0cedR102
It looks like its been incorrect for about 2 years and was changed during @bartgol thyra refactor.
Before: https://github.com/SNLComputation/Albany/blame/7e7871ebbe48533be1817e26d82a94f242b0212c/src/Albany_NullSpaceUtils.cpp#L190
After: https://github.com/SNLComputation/Albany/blame/7912e8626a1080f1c6869a6b1fc676f551b5ac12/src/Albany_NullSpaceUtils.cpp#L205
I added some unit tests so hopefully we'll actually be able to pick up issues next time.

Thanks @rstumin for pointing this out! @mperego Ray mentioned that the effect might be modest: rebalance issues that would effect convergence/performance.

Also I restored the coordinate write feature which will dump in mm format by adding "Write Coordinates to MatrixMarket" to the input under "Discretization".

### DIFF
--- a/src/Albany_NullSpaceUtils.cpp
+++ b/src/Albany_NullSpaceUtils.cpp
@@ -33,7 +33,7 @@ void ComputeNullSpace(
   auto data = getLocalData(coordMV.getConst());
 
   // At least component x should be there
-  const int numNodes = data[0].size(); // length of each vector in the multivector
+  const int numNodes = data[0].size(); // local length of each vector
 
   Teuchos::ArrayRCP<const ST> x, y, z;
   x = data[0];
@@ -84,7 +84,7 @@ void ComputeNullSpace(
 void subtractCentroid(const Teuchos::RCP<Thyra_MultiVector> &coordMV)
 {
   auto spmd_vs = getSpmdVectorSpace(coordMV->range());
-  const int nnodes = spmd_vs->localSubDim(); // local length of each vector
+  const int numNodes = spmd_vs->localSubDim(); // local length of each vector
   const int ndim = coordMV->domain()->dim(); // Number of multivectors are the dimension of the problem
 
   auto data = getNonconstLocalData(coordMV);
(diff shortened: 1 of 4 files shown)


src/Albany_NullSpaceUtils.cpp
```diff
@@ -351,14 +354,11 @@ setCoordinatesAndComputeNullspace(const Teuchos::RCP<Thyra_MultiVector>& coordMV
...
 
-      TEUCHOS_TEST_FOR_EXCEPTION(
-        soln_vs.is_null(), std::logic_error,
-        "numElasticityDim > 0 and (isMueLuUsed() or isFROSchUsed()): soln_map must be provided.");
-        if (isMueLuUsed()) {
-          plist->set("Nullspace", tpetraTraitsArray);
-        } else { // This means that FROSch is used
-          plist->set("Null Space", tpetraTraitsArray);
-        }
+      if (isMueLuUsed()) {
```
mperego, line 357: Why did you remove the Teuchos test for exception?

mperego (approved):
Thanks @jewatkins  for figuring this out and for fixing/testing this! It looks good.

src/Albany_NullSpaceUtils.cpp
jewatkins: I just moved it further up. If soln_vs is null, we shouldn't try to compute the nullspace.

PR #713, opened 2021-07-12
PyAlbany: Gather the global IDs when creating a map on rank 0 with all IDs

kliegeois:
This PR fixes an issue for IO of PyAlbany when the global IDs of a map are not contiguous.

This fix uses the approach of https://github.com/SNLComputation/Albany/blob/master/src/disc/stk/Albany_GenericSTKMeshStruct.cpp#L862-L868 .


mperego (approved):
@kliegeois Thanks! It looks good.
We should add a test (in a subsequent PR) using maps with noncontiguous ids.

kliegeois:
@mperego yes, I agree. This would be better!

PR #716, opened 2021-07-15
Add temporary throw when using Belos + Ifpack2/Amesos2

jewatkins:
@mperego based on this discussion: https://github.com/SNLComputation/Albany/pull/705


mperego (approved):
@jewatkins looks good!

PR #720, opened 2021-07-22
Add FieldUtils for unmanaged MDFields

jewatkins:
Motivation: The memory requirement for basis functions becomes large when using higher quadrature and the previous implementation was allocating each MDField once for every evaluation type. By allocating this data during problem setup using unmanaged MDFields, we can allocate this data once for all evaluation types.

FieldUtils was created to allocate and set unmanaged MDFields for general evaluators (e.g. ComputeBasisFunctions). It is currently only being used for LandIce StokesFO problems and Enthalpy.

Performance: this reduced the memory footprint by 20% for the enthalpy problem with wedge elements and 4th order quadrature. I'll test StokesFO next - I think it will be 20% too.

Note: When the mesh depends on the solution, we need to allocate a unique MDField per evaluation type so we will not see a significant memory reduction there.

@mcarlson801 You may want to take a look in case this affects dynamic sizing.

@mperego @bartgol @ikalash @rppawlo


mperego (commented):
@jewatkins Thanks for working on this! It looks good, but I don't understand what happens when the mesh is a Fad Type. In that case the Jacobian and basis gradients are fad types as well. Did you try compiling with `MESH_DEPENDS_ON_PARAMETERS=TRUE?`

jewatkins:
> @jewatkins Thanks for working on this! It looks good, but I don't understand what happens when the mesh is a Fad Type. In that case the Jacobian and basis gradients are fad types as well. Did you try compiling with `MESH_DEPENDS_ON_PARAMETERS=TRUE?`

In that case, it will still create the `Real` unmanaged fields but they will not be set to the Jacobian evaluation type. I'm not sure what would happen even if they were set. I imagine nothing since they're Real instead of FadType or maybe phalanx would throw since the unmanaged field is not needed.

I have not tested `MESH_DEPENDS_ON_PARAMETERS=TRUE`. If I turn that on, are there tests for it (i.e. I can just run ctest)?

mperego:
> I have not tested MESH_DEPENDS_ON_PARAMETERS=TRUE. If I turn that on, are there tests for it (i.e. I can just run ctest)?

Yes, in fact there are a couple of tests more.

mperego:
This is the right option
`-D ENABLE_MESH_DEPENDS_ON_PARAMETERS:BOOL=ON`

jewatkins:
Okay, tested and everything passes.

mperego (approved):

mperego:
@jewatkins thanks for checking!

PR #723, opened 2021-07-26
fixed race condition in PyAlbany writeMVector function, that was corr…

thartland:
The writeMVector function had a race condition when a parallel job tried to write a non-distributedFile, as each MPI process attempted to write to the same file. Not all MPI processes need to write to `filename+'.npy'` or `filename+'.txt'` as after gathering the multivector the only MPI process which has the global vector data is the root process.


kliegeois:
Looks good to me.
Thanks!

PR #725, opened 2021-07-29
Sideset Dynamic Sizing

mcarlson801:
This PR changes the way that sideset layouts are allocated. The sizes for sideset layouts are determined during `Albany_Application`'s `finalSetUp` instead of  during `buildProblem` so that we can have access to sideset information that is produced during `createDiscretization`.

The purpose of this change is to enable slim sideset allocation for all builds instead of just GPU builds.


src/Albany_Application.cpp
```diff
@@ -735,6 +735,71 @@ Application::finalSetUp(
...
+
+  // compute largest workset size over all worksets for each sideset name
+  std::map<std::string, unsigned int> maxSideSetSizes;
+  for (unsigned int i = 0; i < numWorksets; ++i) { 
+    const LocalSideSetInfoList& sideSetView = disc->getSideSetViews(i);
+    for (auto it = sideSetView.begin(); it != sideSetView.end(); ++it) {
+      if (maxSideSetSizes.find(it->first) == maxSideSetSizes.end()) {
+        maxSideSetSizes[it->first] = 0;
+      }
+      unsigned int sideSetSize = it->second.size;
```
jewatkins, line 754: Is it possible for this to be uninitialized? I worry this might be zero and a side field could be set with a zero size. Can we safely add an error here for that case?

src/Albany_Application.cpp
```diff
@@ -735,6 +735,71 @@ Application::finalSetUp(
...
+          t_dl.setExtents(t_dims[0], t_dims[1], t_dims[2], t_dims[3], t_dims[4]);
+          break;
+        case 6:
+          t_dl.setExtents(t_dims[0], t_dims[1], t_dims[2], t_dims[3], t_dims[4], t_dims[5]);
+          break;
+        default:
+          TEUCHOS_TEST_FOR_EXCEPTION (true, std::logic_error,
+            "Error! Sideset dynamic sizing as encountered a layout with more field tags than expected.\n");
+      }
+    }
```
jewatkins, line 799: Can we also throw if t_dims[0] is zero here? This should ensure that all layouts used by the field manager have a size that's initialized.

src/Albany_Application.cpp
```diff
@@ -2922,6 +2996,7 @@ Application::evaluateStateFieldManager(
         sfm[ps]
             ->setKokkosExtendedDataTypeDimensions<PHAL::AlbanyTraits::Jacobian>(
                 derivative_dimensions);
+
```
jewatkins, line 2999: I think you need `setSideSetExtents<EvalT>(sfm[ps]);` - i.e. let's go ahead and call this before all fm postReg, even if sides are not currently used. It should skip everything except the final check to make sure all dynamic layouts are set.

src/Albany_Application.cpp
```diff
@@ -735,6 +735,71 @@ Application::finalSetUp(
   for (int i = 0; i < responses.size(); ++i) { responses[i]->postRegSetup(); }
 }
 
+template<typename Traits>
+void
+Application::setSideSetExtents(Teuchos::RCP<PHX::FieldManager<PHAL::AlbanyTraits>>& in_fm) const
```
jewatkins, line 740: Let's call this `setDynamicLayoutSizes()`

src/problems/Albany_Layouts.cpp
```diff
@@ -128,42 +128,42 @@ Albany::Layouts::Layouts (int worksetSize, int numVertices, int numNodes, int nu
   isSideLayouts = true;
 
   // Collapsed sideset layouts are now used to ensure contiguous memory access for efficient GPU evaluation
-  sidesetWorksetSize = singleWorksetSizeAllocation ? sidesetWorksetSize : worksetSize*numSides;
+  // sidesetWorksetSize = singleWorksetSizeAllocation ? sidesetWorksetSize : worksetSize*numSides;
```
jewatkins, line 131: Remove unused inputs (e.g. worksetSize?) and comments.

src/problems/Albany_Layouts.cpp
```diff
@@ -128,42 +128,42 @@ Albany::Layouts::Layouts (int worksetSize, int numVertices, int numNodes, int nu
...
+  qp_scalar    = rcp(new MDALayout<Side,QuadPoint>(sideSetName,0,numQPts));
+  cell_scalar  = rcp(new MDALayout<Side,QuadPoint>(sideSetName,0,1));
+  cell_scalar2 = rcp(new MDALayout<Side>(sideSetName,0));
 
-  node_vector = rcp(new MDALayout<Side,Node,Dim>(sidesetWorksetSize,numNodes,vecDim));
-  qp_vector   = rcp(new MDALayout<Side,QuadPoint,Dim>(sidesetWorksetSize,numQPts,vecDim));
-  cell_vector = rcp(new MDALayout<Side,Dim>(sidesetWorksetSize,vecDim));
+  node_vector = rcp(new MDALayout<Side,Node,Dim>(sideSetName,0,numNodes,vecDim));
+  qp_vector   = rcp(new MDALayout<Side,QuadPoint,Dim>(sideSetName,0,numQPts,vecDim));
+  cell_vector = rcp(new MDALayout<Side,Dim>(sideSetName,0,vecDim));
```
jewatkins, line 141: Let's try to make these `PHX::DataLayout` like we discussed. It looks like the string name will have to be unique for each layout in that case:
https://github.com/trilinos/Trilinos/blob/master/packages/phalanx/example/FiniteElementAssembly_MixedFieldTypes/FiniteElementAssembly_MixedFieldTypes.cpp
I'm curious to see if we still get layout sizes if we dump debug info in this case.

jewatkins (changes requested):

bartgol (approved):
Jerry already gave all the important feedback, and I agree on everything he said. I'm gonna approve, modulo the changes to address his comments.

mcarlson801:
Thanks for the feedback, I think these changes make sense. I'll go make the changes and then remove the slim allocation stuff that is now no longer necessary after this PR.

src/Albany_Application.cpp
```diff
@@ -735,6 +735,71 @@ Application::finalSetUp(
...
+    t_dl.dimensions(t_dims);
+
+    // Check if dimension[0] is Side
+    std::string first_dim_name = t_dl.name(0);
+    std::string t_identifier = t_dl.identifier();
+    std::string sideSetName = t_identifier.substr(0, t_identifier.find("<"));
+
+    if (first_dim_name == "Side") {
+
+      t_dims[0] = maxSideSetSizes[sideSetName];
```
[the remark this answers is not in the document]
mcarlson801: It looks like adding this check has caused a few tests to fail. These tests are `SideSetLaplacian_3D`, `FO_ISMIP_TestC_Epetra`, and `FO_ISMIP_TestC_Tpetra`. 

It seems layouts are being constructed for side sets that don't have a corresponding `sideSetView`. Any ideas how this can happen or how to handle this case? If there is no `sideSetView`, there would have been an error in the first evaluator that operated on the side set so I assume nothing is being done with these side sets.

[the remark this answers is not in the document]
bartgol: I'm puzzled actually: why should the 1st chunk of `t_identifier` be a side set name? Shouldn't those be just field names?

src/problems/Albany_Layouts.cpp
mcarlson801: There are a few layouts that don't follow the <Side,...> pattern that use worksetSize as their first extent. (`node_node_scalar`, `node_node_vector`, and `node_node_tensor`)

I'm not using dynamic sizing for these at the moment since I don't know how they are used or if it is possible to reduce these allocations. Just a heads up as to why worksetSize is sticking around for the moment. The other unused inputs will be removed in my next commit.

src/Albany_Application.cpp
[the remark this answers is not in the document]
mcarlson801: Field identifiers have the field names but these are identifiers for the underlying layouts. MDALayout allows for a prefix to be added to these identifiers so I attached the side set names so we could pair the layouts to their corresponding side set.

The format for MDALayout identifiers is as follows:

Prefix<Rank1, Rank2,...>(Extent1,Extent2,...)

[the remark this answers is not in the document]
jewatkins: I'd have to look at the output of all side fields and side set views and the error output. It would seem to me that if there's a side field in the fm, there should be a max size for that side.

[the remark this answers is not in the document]
mcarlson801: > I'd have to look at the output of all side fields and side set views and the error output. It would seem to me that if there's a side field in the fm, there should be a max size for that side.

Here is an example output for one of the failing tests. I've also pushed my current commit so you can see the changes I made.

[log-ctest-dynamic-sizing.1012351.txt](https://github.com/SNLComputation/Albany/files/6918720/log-ctest-dynamic-sizing.1012351.txt)

src/problems/Albany_Layouts.cpp
jewatkins: Can we try to remove them? What are they used for? We should try to use the <Side,...> everywhere.

jewatkins: I queried the source and I didn't see any use for these. Let's go ahead and remove them and remove worksetsize unless you think there's another use for it.

src/Albany_Application.cpp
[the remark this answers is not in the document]
jewatkins: The log has `Adding sideset SideSet4 to maxSideSetSizes` but then throws with `SideSet4 was not found` so there's definitely something buggy in the logic. Can you run with one proc, print the full tag name, layout name and maybe the max size for SideSet4?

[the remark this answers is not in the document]
mcarlson801: I couldn't replicate with a single proc but I can with 2. I printed the requested information for each MPI rank and it looks like sideSet4 only gets added to maxSideSetSizes on rank 0 but rank 1 still sees a layout for sideSet4 in the field manager. I'm going to double check that the value for "numWorksets" is correct.

Here is the log:
[updated-log-dynamic-sizing.txt](https://github.com/SNLComputation/Albany/files/6919054/updated-log-dynamic-sizing.txt)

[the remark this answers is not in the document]
jewatkins: can you print the tag name of the field that uses the layout too?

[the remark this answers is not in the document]
bartgol: Might be that the sideset is empty on a rank, and maybe the disc setup does not create side views for them?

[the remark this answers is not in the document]
mcarlson801: Here is the log with field tag names included:
[updated-log-dynamic-sizing.txt](https://github.com/SNLComputation/Albany/files/6919536/updated-log-dynamic-sizing.txt)

There were issues with the Coord Vec field previously which we fixed by checking the response field manager for field tags in addition to the other field managers. I'm still not sure what situation a field tag would show up in a field manager for a side set that never gets set up.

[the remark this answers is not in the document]
jewatkins: It looks like in this particular case, SideSet4 is set as a BC which, I think, constructs the appropriate evaluator for all ranks (it looks like it only needs a mesh spec). But, maybe there's no side discretization on a rank because the sideset is empty (as Luca pointed out). In this case, the evaluator still runs but it's skipped:
```
if (workset.sideSetViews->find(sideSetName)==workset.sideSetViews->end()) return;
```
So, we need to apply the same check. For a given field side, loop over all sideSetViews. If the side doesn't exist, then skip setting the size. Then, you can keep the same error checks you currently have after this check.

Alternatively, we could eliminate the evaluators associated with this side in the fm but I think that might be a larger design choice.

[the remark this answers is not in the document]
jewatkins: Actually, it seems like maxSideSetSizes provides the same functionality and the errors would be redundant. So maybe something like:
```
if (maxSideSetSizes.find(sideSetName) == maxSideSetSizes.end()) continue;
```
would be equivalent.

src/problems/Albany_Layouts.cpp
mcarlson801: I switched the side set layout types to `PHX::DataLayout` as we discussed and now fields aren't showing up at all when I call `getFieldTagsForSizing`. I'm not sure why this would be the case. Do you want me to push the current commit so you can take a look at the code? I still think the change I mentioned to MDALayout on Phalanx's end would be a much more elegant solution and is likely the intended functionality.

Here is an example log. Note that this build has the print statements that print all (side set) field tags found in the field manager.
[log-ctest-dynamic-sizing.1012570.txt](https://github.com/SNLComputation/Albany/files/6934415/log-ctest-dynamic-sizing.1012570.txt)

jewatkins: can you dump all field tag identifiers? yes, go ahead and push the code. I wonder if the fm is having issues with a mix of MDA and Data layouts.

mcarlson801: It looks like the sideset tags are there but they don't have sizes in the identifier. Not sure why they weren't getting printed before.

Log: 
[log-ctest-dynamic-sizing.1012589.txt](https://github.com/SNLComputation/Albany/files/6934775/log-ctest-dynamic-sizing.1012589.txt)

jewatkins: Okay yeah I think that's correct. Is it possible that `first_dim_name` is not being set?

Could you also dump the phalanx graph to see if the sizes show up there?
`Phalanx Graph Visualization Detail: 1` in the input file

jewatkins (approved):
Looks good, thanks!

PR #726, opened 2021-07-30
Fix PyAlbany stackedTimers when more than 1 problem is used

kliegeois:
It has been observed that when more than one PyAlbany problem is created, the stackedTimer can be printed correctly only for the last created problem.

This PR fixes this issue; the stackedTimer can be correctly printed for all the created PyAlbany problems.

The tests passed correctly on Blake.

### DIFF
--- a/PyAlbany/src/Albany_Interface.cpp
+++ b/PyAlbany/src/Albany_Interface.cpp
@@ -78,10 +78,10 @@ PyProblem::PyProblem(std::string filename, Teuchos::RCP<PyParallelEnv> _pyParall
     PrintPyHeader(*out);
 
     stackedTimer = Teuchos::rcp(
-        new Teuchos::StackedTimer("Albany Total Time"));
+        new Teuchos::StackedTimer("PyAlbany Total Time"));
     Teuchos::TimeMonitor::setStackedTimer(stackedTimer);
 
-    stackedTimer->start("Albany: Setup Time");
+    stackedTimer->start("PyAlbany: Setup Time");
 
     comm = this->pyParallelEnv->comm;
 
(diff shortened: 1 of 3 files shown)


PyAlbany/src/Albany_Interface.cpp
```diff
@@ -434,6 +434,8 @@ Teuchos::RCP<PyTrilinosMultiVector> PyProblem::getReducedHessian(const int g_ind
 
 void PyProblem::performSolve()
 {
+    if (Teuchos::TimeMonitor::getStackedTimer() != stackedTimer)
+        Teuchos::TimeMonitor::setStackedTimer(stackedTimer);
```
mperego, line 438: Naive question, why do you need to check whether the stacked timers are the same, why don't you simply set it?

mperego (approved):
Looks good. Thanks!

PyAlbany/src/Albany_Interface.cpp
kliegeois: I thought that it would be best not to set the stackedTimer if it was already set correctly. But I have no strong feelings about that to be honest.

kliegeois:
@jewatkins @mcarlson801 this PR includes an example on how to access timers on levels deeper than the level 0 of the stackedTimer:
```
        setup_time_2 = stackedTimer.findBaseTimer("PyAlbany Total Time@PyAlbany: Setup Time").accumulatedTime()
        setup_time_fix_node_sharing = stackedTimer.findBaseTimer("PyAlbany Total Time@PyAlbany: Setup Time@Albany Setup: fix_node_sharing").accumulatedTime()
```
This should resolve the issue discussed during our last meeting.

kliegeois:
@mperego I updated all the timer names in PyAlbany to be "PyAlbany..." instead of "Albany...".

mperego (approved):
looks good to me. Thanks!!

jewatkins:
Thanks!

PR #729, opened 2021-08-10
Several QOL changes

jewatkins:
This adds some QOL changes based on a discussion with @mperego 
1. Force a single workset when using memoization (throw if trying to use multiple)
   - This is needed because memoization only works for a single workset.

2. Suppress MPI info, add option to print "Report MPI Info"
   - I was getting tired of seeing each rank dump the processor name in all simulations (particularly for the large ones I just did) so I added it as a debug option (off by default). I could turn it on by default if others prefer seeing it. Basically this:
```
Teuchos::GlobalMPISession::GlobalMPISession(): started processor with name ecw00108 and rank 0!
```
now looks like this (but now we can turn it on/off):
```
Rank 0 of 30 exists on processor ecw00108
```

3. Dump fad types and size at start of simulation
    - Example (it can be modified):
```
...
** Albany cxx compiler ---- Clang 10.0.0
** Albany FadType --------- SFad12
** Albany TanFadType ------ SLFad30
** Albany HessianVecFad  -- SFad12
** Simulation start time -- 2021-08-09 at 20:09:23
...
```

4. Change timer so that sims exit more cleanly
    - I noticed that if something fails during setup, the timer segfaults so this should fix it

5. Set default workset size in Extruded and Ioss to -1 (single workset)
    - this should always be more efficient anyways. The main difference will be: if you run out-of-memory, it's probably because of this so you have to manually specify a workset size in the input file to reduce the memory allocation.

@mperego @ikalash @bartgol @kliegeois @mcarlson801

### DIFF
--- a/src/Albany_Application.cpp
+++ b/src/Albany_Application.cpp
@@ -18,6 +18,7 @@
 
 #include "Teuchos_TimeMonitor.hpp"
 
+#include <stdexcept>
 #include <string>
 #include "Albany_DataTypes.hpp"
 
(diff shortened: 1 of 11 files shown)


mperego (approved):
Thanks @jewatkins! Fantastic that you suppressed MPI info!

jewatkins:
I'll give this a day to see if others have comments before pushing.

PR #730, opened 2021-08-12
Change to PyAlbany performSolve method

thartland:
performSolve is changed so that it returns a boolean, true if the solve converged and otherwise false. The purpose is that from a PyAlbany application code one can assess if the solve converged, which will impact how one would interpret the associated solution data.


mperego:
Looks good to me. @kliegeois are you OK with these change?

kliegeois:
Looks good to me!

PR #734, opened 2021-08-18
Match derivative dimensions with sfad size

jewatkins:
@mperego This will throw if derivative dimensions doesn't exactly match sfad size. SLFad size doesn't have a "static_size" and already throws if derivative dimensions is larger so I think this is the only case we need to check.

### DIFF
--- a/src/PHAL_Utilities.cpp
+++ b/src/PHAL_Utilities.cpp
@@ -4,9 +4,10 @@
 //    in the file "license.txt" in the top-level Albany directory  //
 //*****************************************************************//
 
+#include <stdexcept>
+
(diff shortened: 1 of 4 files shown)


mperego (approved):
@jewatkins, thanks! It looks good to me!

jewatkins:
So I just tested this with the performance tests and it fails because I use a larger SFAD size when running `PopulateMesh`. It's probably not great to be constrained like that. I'm going to see if I can find a work around before pushing this. I don't think PopulateMesh should be calling getDerivativeDimensions() since Jacobians are never evaluated.

PR #739, opened 2021-09-06
Consistently use PHX::print<Entity>()

mperego:
This is related to #655

@bartgol For some reason I had to add declarations of the specializations, otherwise PHX::print<Entity>() would not  use the the specializations.


bartgol (approved):
It makes sense. When the compiler parses the file after all the includes, it finds a definition of print<T>, so it simply uses that one. Before this PR, the existence of the specializations was not known until link time, but by then the compiler had already chosen to use the default base impl. Now we inform the compiler that a specialization exists, so it does not use the generic template version.

Good catch!

PR #740, opened 2021-09-08
Add warning if SFad size is larger than derivative dimension

jewatkins:
@mperego @ikalash @kliegeois This adds a warning if SFad size is set to something larger than what is required to store the derivative components. This capability is not tested and there's been issues with it in the past (see `LinComprNS_1D_standingWave` as an example). There may be some issue with how xdot is gathered and then used in local assembly for this specific example but there might be other issues.

I also removed some unnecessary calls to getDerivativeDimensions() because they're not used and would pollute runs with warnings.


mperego (approved):
Looks good to me. Thanks!

PR #743, opened 2021-09-20
Remove default status tests

jewatkins:
NOX requires well formed status tests and the current default status tests causes issues. see #742

@mperego @ikalash This fixes all tests. There were only a few tests which require the default. One notable case is populate mesh. If you have populate mesh input files which don't specify an exit tolerance, NOX will attempt to use some default epetra preconditioner and crash if Tpetra is used as the build type. We may want to look into that issue further. I'm not sure why there's a default preconditioner.


mperego (approved):
It looks good to me. Thanks a lot Jerry!

mperego:
> There were only a few tests which require the default. One notable case is populate mesh. If you have populate mesh input files which don't specify an exit tolerance, NOX will attempt to use some default epetra preconditioner and crash if Tpetra is used as the build type. We may want to look into that issue further.

Yes, I think we need to rethink the populate mesh capability.

PR #749, opened 2021-10-05
Enabling simultaneous inversion of bed topography and surface height

mperego:
This PR enables simultaneous inversion of bed topography and surface height

It also improves the handling of planar fileds such as thickness, surface height and bed topography, and of responses on planar sidesets

Other minor fixes.

### DIFF
--- a/src/LandIce/evaluators/LandIce_UpdateZCoordinate_Def.hpp
+++ b/src/LandIce/evaluators/LandIce_UpdateZCoordinate_Def.hpp
@@ -62,6 +62,13 @@ UpdateZCoordinateMovingTop (const Teuchos::ParameterList& p,
   rho_w = p_list->get<double>("Water Density");
 }
 
+template<typename EvalT, typename Traits>
+void UpdateZCoordinateMovingTop<EvalT, Traits>::
+postRegistrationSetup(typename Traits::SetupData d, PHX::FieldManager<Traits>& fm)
+{
+  d.fill_field_dependencies(this->dependentFields(),this->evaluatedFields());
+}
+
 //**********************************************************************
 template<typename EvalT, typename Traits>
 void UpdateZCoordinateMovingTop<EvalT, Traits>::
@@ -146,6 +153,13 @@ UpdateZCoordinateMovingBed (const Teuchos::ParameterList& p,
   this->setName("Update Z Coordinate Moving Bed");
 }
 
+template<typename EvalT, typename Traits>
+void UpdateZCoordinateMovingBed<EvalT, Traits>::
+postRegistrationSetup(typename Traits::SetupData d, PHX::FieldManager<Traits>& fm)
+{
+  d.fill_field_dependencies(this->dependentFields(),this->evaluatedFields());
+}
+
 //**********************************************************************
 //Kokkos functors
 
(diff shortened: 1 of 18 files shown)


src/evaluators/response/PHAL_ResponseSquaredL2DifferenceSide_Def.hpp
```diff
@@ -17,9 +17,24 @@ template<typename EvalT, typename Traits, typename SourceScalarT, typename Targe
 PHAL::ResponseSquaredL2DifferenceSideBase<EvalT, Traits, SourceScalarT, TargetScalarT>::
 ResponseSquaredL2DifferenceSideBase(Teuchos::ParameterList& p, const Teuchos::RCP<Albany::Layouts>& dl)
 {
-  // get response parameter list
+  // Get response parameter list
   Teuchos::ParameterList* plist = p.get<Teuchos::ParameterList*>("Parameter List");
 
+  // Validate response parameter list
```
bartgol, line 23: This is a good idea. We've all been bitten by params not correctly forwarded to the evaluator. Validation is a good idea, and  we should try to propagate it to other evals (if we have time), especially those that are auto-constructed, like responses.

src/evaluators/response/PHAL_ResponseSquaredL2DifferenceSide_Def.hpp
```diff
@@ -40,17 +55,32 @@ ResponseSquaredL2DifferenceSideBase(Teuchos::ParameterList& p, const Teuchos::RC
 
   layout->dimensions(dims);
 
+  std::string sideSetNameForMetric =
+      (plist->isParameter("Is Side Set Planar") && plist->get<bool>("Is Side Set Planar")) ?
```
bartgol, line 59: Minor: since plist is not const, you can simply do
```
std::string sideSetNameForMetric = siteSetName +
       (plist->get<bool>("Is Side Set Planar", false) ? "_planar" : "");
```
Might make the line easier to visually parse. (I wish Teuchos::ParameterList had an easy way to provide a default (if key is not found) even for const plists...)

src/evaluators/response/PHAL_ResponseSquaredL2DifferenceSide_Def.hpp
```diff
@@ -40,17 +55,32 @@ ResponseSquaredL2DifferenceSideBase(Teuchos::ParameterList& p, const Teuchos::RC
...
+  w_measure   = decltype(w_measure)(Albany::weighted_measure_name + "_" + sideSetNameForMetric, dl_side->qp_scalar);
   scaling     = plist->get("Scaling",1.0);
 
   this->addDependentField(sourceField);
+
+  rmsScaling = plist->isParameter("Root Mean Square Error Field Name");
+  if(rmsScaling) {
+    rootMeanSquareField = decltype(rootMeanSquareField)(plist->get<std::string>("Root Mean Square Error Field Name"),layout);
+    this->addDependentField(rootMeanSquareField);
+    rmsScaling = true;
```
bartgol, line 79: This line is redundant. You already set `rmsScaling=true` if you're in the if block.

src/evaluators/response/PHAL_ResponseSquaredL2DifferenceSide_Def.hpp
```diff
@@ -70,6 +100,9 @@ ResponseSquaredL2DifferenceSideBase(Teuchos::ParameterList& p, const Teuchos::RC
 
   this->setName("Response Squared L2 Error Side" + PHX::print<EvalT>());
 
+  if(plist->isParameter("Response Depends On Solution Column") && plist->get<bool>("Response Depends On Solution Column"))
```
bartgol, line 103: Here too: plist is not const, so you can use the get method providing a default, and make this line shorter.

src/evaluators/response/PHAL_ResponseSquaredL2DifferenceSide_Def.hpp
```diff
@@ -70,6 +100,9 @@ ResponseSquaredL2DifferenceSideBase(Teuchos::ParameterList& p, const Teuchos::RC
...
+    cell_topo = p.get<Teuchos::RCP<Teuchos::ParameterList> >("Parameters From Problem")->get<Teuchos::RCP<const CellTopologyData> >("Cell Topology");
```
bartgol, line 104: Is the template arg wrong? Shouldn't it be `Teuchos::ParameterList*` (notice the pointer missing in your version)?

src/LandIce/evaluators/LandIce_UpdateZCoordinate_Def.hpp
```diff
@@ -201,4 +215,112 @@ evaluateFields(typename Traits::EvalData workset)
...
+{
+  isTopSurfParam = p.isParameter("Top Surface Parameter Name");
+  isBedTopoParam = p.isParameter("Bed Topography Parameter Name");
+
+  if(isBedTopoParam == true) {
+    bedTopoIn = decltype(bedTopoIn)(p.get<std::string> ("Bed Topography Parameter Name"), dl->node_scalar);
+    this->addDependentField(bedTopoIn);
+    this->addEvaluatedField(bedTopo);
+  } else {
+    this->addDependentField(bedTopo);
```
bartgol, line 238: I'm curious: does this give a compiler error, since the dependent field is nor marked as const?

src/LandIce/problems/LandIce_StokesFOThermoCoupled.hpp
```diff
@@ -159,34 +153,59 @@ constructEvaluators (PHX::FieldManager<PHAL::AlbanyTraits>& fm0,
       p->set<std::string>("Updated Bed Topography Name", bed_topography_name);
       p->set<Teuchos::ParameterList*>("Physical Parameter List", &params->sublist("LandIce Physical Parameters"));
 
-      ev = Teuchos::rcp(new UpdateZCoordinateMovingBed<EvalT,PHAL::AlbanyTraits>(*p, dl));
+      ev = Teuchos::rcp(new LandIce::UpdateZCoordinateMovingBed<EvalT,PHAL::AlbanyTraits>(*p, dl));
```
bartgol, line 156: Curious: were these namespace specs necessary? Should compile even without, right?

bartgol (approved):
Looks good! I have a couple of comments, but nothing too serious. :)

src/evaluators/response/PHAL_ResponseSquaredL2DifferenceSide_Def.hpp
mperego: agreed, and it doesn't take that long to create the validation list.

src/evaluators/response/PHAL_ResponseSquaredL2DifferenceSide_Def.hpp
mperego: I know it's verbose, but I don't like that `get()` creates the parameter in the list. I spent quite some time in the past trying to find where the parameterlists are modified and default options added, when things go wrong.
I'm leaning towards making plist const.

src/evaluators/response/PHAL_ResponseSquaredL2DifferenceSide_Def.hpp
mperego: good catch, I'll remove the line.

src/LandIce/problems/LandIce_StokesFOThermoCoupled.hpp
mperego: no, they are not necessary. I don't recall why I added them.

src/LandIce/evaluators/LandIce_UpdateZCoordinate_Def.hpp
mperego: Good question. No, I don't get a compiler error... we'll see if clang complains.

bartgol: No need. I just recalled PHX giving a warning (I think) if the data type of a dependent field was not marked as const. Maybe it was a deprecated method, I don't recall. But even if you did get it, fixing it would make the code a tad complicated (you'd have to use switches later on, to get a ref to bedTopo vs bedTopoIn...).

This can be left as is.

src/evaluators/response/PHAL_ResponseSquaredL2DifferenceSide_Def.hpp
mperego: I think it's correct, it's already and RCP. Why are you expecting it to be an RCP of a pointer?

src/evaluators/response/PHAL_ResponseSquaredL2DifferenceSide_Def.hpp
mperego: same as above

src/evaluators/response/PHAL_ResponseSquaredL2DifferenceSide_Def.hpp
bartgol: Ah, I missed the RCP part. Sorry.

PR #753, opened 2021-10-07
Add Tpetra path for SDBC to run on device

jewatkins:
This adds a path to run SDBC on device. It improves CPU/GPU performance significantly (although the runtime on CPU was small to begin with) by avoiding a local matrix fill on each row and only filling when needed.

I tested this on Humboldt1km-10km and obtained the same results (both CPU/GPU compared to old code). I'm not sure if this evaluator is tested in any regression tests. I also think I found a bug (will comment below) but I'm not sure if it's ever used (still looking for a test that runs that part of the code).

### DIFF
--- a/src/evaluators/bc/PHAL_SDirichletField_Def.hpp
+++ b/src/evaluators/bc/PHAL_SDirichletField_Def.hpp
@@ -171,50 +171,22 @@ template<typename Traits>
 void SDirichletField<PHAL::AlbanyTraits::Jacobian, Traits>::
 evaluateFields(typename Traits::EvalData dirichlet_workset)
 {
-
-  Teuchos::RCP<Thyra_Vector>       f = dirichlet_workset.f;
-  if(Teuchos::nonnull(f)) {
-    Teuchos::RCP<const Thyra_Vector> x = dirichlet_workset.x;
-    Teuchos::ArrayRCP<ST> x_view = Teuchos::arcp_const_cast<ST>(Albany::getLocalData(x));
-
-    const Albany::NodalDOFManager& fieldDofManager = dirichlet_workset.disc->getDOFManager(this->field_name);
-    //MP: If the parameter is scalar, then the parameter offset is set to zero. Otherwise the parameter offset is the same of the solution's one.
-    auto fieldNodeVs = dirichlet_workset.disc->getNodeVectorSpace(this->field_name);
-    auto fieldVs = dirichlet_workset.disc->getVectorSpace(this->field_name);
-    bool isFieldScalar = (fieldNodeVs->dim() == fieldVs->dim());
-    int fieldOffset = isFieldScalar ? 0 : this->offset;
-    const std::vector<GO>& nsNodesGIDs = dirichlet_workset.disc->getNodeSetGIDs().find(this->nodeSetID)->second;
-
-    Teuchos::RCP<const Thyra_Vector> pvec = dirichlet_workset.distParamLib->get(this->field_name)->vector();
-    Teuchos::ArrayRCP<const ST> p_constView = Albany::getLocalData(pvec);
-
-    const std::vector<std::vector<int> >& nsNodes = dirichlet_workset.nodeSets->find(this->nodeSetID)->second;
-    auto field_node_indexer = Albany::createGlobalLocalIndexer(fieldNodeVs);
-    for (unsigned int inode = 0; inode < nsNodes.size(); inode++) {
-        int lunk = nsNodes[inode][this->offset];
-        GO node_gid = nsNodesGIDs[inode];
-        int lfield = fieldDofManager.getLocalDOF(field_node_indexer->getLocalElement(node_gid),fieldOffset);
-        x_view[lunk] = p_constView[lfield];
-    }
-  }
-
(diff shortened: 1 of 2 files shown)


src/evaluators/bc/PHAL_SDirichletField_Def.hpp
```diff
@@ -197,45 +198,80 @@ evaluateFields(typename Traits::EvalData dirichlet_workset)
...
-    if (row_is_dbc && fill_residual == true) {
-      int lunk = nsNodes[local_row][this->offset];
-      f_view[lunk] = 0.0;
-    }
+    for (LO local_row = 0; local_row < num_local_rows; ++local_row) {
+      Albany::getLocalRowValues(J, local_row, indices, entries);
+      auto row_is_dbc = col_is_dbc_data[local_row] > 0;
+      if (row_is_dbc && fill_residual == true) {
+        int lunk = nsNodes[local_row][this->offset];
+        f_view[lunk] = 0.0;
```
jewatkins, line 260: Is this correct? My understanding is that nsNodes is dimensioned by the list of nodes on the boundary, not the number of local matrix rows. I think there's an attempt to grab the local row, `lunk` but isn't `lunk == local_row`? If I can't find a test which actually uses this, I'll try to print everything to make sure my understanding is correct.

jewatkins: It seems to me like the number of local rows in the Jacobian matrix would include all components. Also I think if nsNodes is any subset of all nodes in the system, nsNodes[local_row] would go out-of-bounds. I'll try dumping values to confirm.

jewatkins: This is only needed in the residual vector which I don't think is changed in this evaluator for ALI (at least I haven't seen it changed - still looking for a case that uses this code). The residual evaluator does the right thing I think.

mperego: I realize I don't really understand this code. What do you mean with residual vector?

jewatkins: The `f` vector.

bartgol: How can it not happen for ALI? In our ME, every call to computeGlobalJacobian/Tangent also passes `f_out`, to try to compute the res alongside the jac/tangent.

jewatkins: f and J are passed but when computeGlobalJacobian is called, it seems to only do the J portions when evaluateFields<Jacobian> is called. I think f is filled when computeGlobalResidual is called. It does seem like at some point these were both being computed in computeGlobalJacobian.

src/evaluators/bc/PHAL_SDirichletField_Def.hpp
```diff
@@ -197,45 +198,80 @@ evaluateFields(typename Traits::EvalData dirichlet_workset)
...
+  this->set_row_and_col_is_dbc(dirichlet_workset);
 
+  const bool is_tpetra = Albany::build_type() == Albany::BuildType::Tpetra;
   bool const fill_residual = f != Teuchos::null;
+  if (is_tpetra)
+  {
+    Tpetra_Vector::dual_view_type::t_dev fView;
+    if (fill_residual)
+    {
+      printf("Fill residual...\n");
```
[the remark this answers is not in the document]
jewatkins: I verified that no test actually uses this evaluator. I'll enable it for the case you mentioned.

src/evaluators/bc/PHAL_SDirichletField_Def.hpp
bartgol: > f and J are passed but when computeGlobalJacobian is called, it seems to only do the J portions when evaluateFields<Jacobian> is called. I think f is filled when computeGlobalResidual is called. It does seem like at some point these were both being computed in computeGlobalJacobian.

That was my understanding: if f wasn't evaluated along with Jacobian/Tangent/DistParamDeriv, _then_ call computeGlobalResidual (as last resort to get f). I wonder if something changed in NOX/Piro, to cause f to be evaluated before (or at least not together with) other stuff...

mperego: @jewatkins I came to the conclusion that we should remove that piece of the code:
https://github.com/sandialabs/Albany/blob/dc81ea96bc6946668b72cacf5ccba41a48360afb/src/evaluators/bc/PHAL_SDirichletField_Def.hpp#L176-L199

in fact this is done, and correctly, in the preEvaluate:
https://github.com/sandialabs/Albany/blob/dc81ea96bc6946668b72cacf5ccba41a48360afb/src/evaluators/bc/PHAL_SDirichletField_Def.hpp#L297-L320

I still haven't figured out why f is always null there.

jewatkins: Okay will do. I verified the number of local rows is much larger than the number of node set nodes in the test I modified so this would be a bug if it were executed.

PR #755, opened 2021-10-19
Add the Hessian reconstruction for scalar parameters

kliegeois:
This PR adds the reconstruction of the Hessian of response function for scalar parameters.

All the tests passed and the branch builds warning free.

The weighted misfit function has been updated to be able to use vector of scalar parameters.

An option has been added to be able to replace the Hessian of a response w.r.t a parameter by the identity matrix.

@mperego @bartgol

### DIFF
--- a/src/Albany_Application.cpp
+++ b/src/Albany_Application.cpp
@@ -2487,7 +2487,7 @@ Application::evaluateResponse_HessVecProd_pp(
 }
 
 void
-Application::evaluateResponseDistParamHessian_pp(
+Application::evaluateResponseHessian_pp(
     int                                     response_index,
     int                                     parameter_index,
     const double                            current_time,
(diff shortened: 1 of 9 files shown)


src/Albany_Application.cpp
```diff
@@ -2498,51 +2498,110 @@ Application::evaluateResponseDistParamHessian_pp(
...
+  Albany::getParameterVectorID(l1, l1_is_distributed, param_name);
 
   // Get the crs Hessian:
   RCP<Tpetra_CrsMatrix> Ht = Albany::getTpetraMatrix(H);
   Ht->resumeFill();
 
-  // Create a colorer
-  Zoltan2::TpetraCrsColorer<Tpetra_CrsMatrix> colorer(Ht);
+  bool replace_by_I = 
+    problemParams->sublist("Hessian").sublist(Albany::strint("Response", response_index)).isParameter(Albany::strint("Replace H_pp by I parameter", parameter_index)) ?
```
mperego, line 2510: I would make this more verbose:
"Replace H_pp with Identity for Parameter"

src/Albany_Application.cpp
```diff
@@ -2498,51 +2498,110 @@ Application::evaluateResponseDistParamHessian_pp(
...
+    problemParams->sublist("Hessian").sublist(Albany::strint("Response", response_index)).get<bool>(Albany::strint("Replace H_pp by I parameter", parameter_index)) :
+    false;
 
-  colorer.computeColoring(coloring_params);
+  if (replace_by_I) {
```
mperego, line 2514: does this work also for dstributed parameters?

src/Albany_Application.cpp
mperego: It is better to validate the "Hessian" and "Response X" sublists to avoid typos.

bartgol: > I would make this more verbose: "Replace H_pp with Identity for Parameter"

Or even the shorter "Replace H_pp with Identity" (the part "for Parameter" is kind of implied by the "_pp" part).

Sorry, I'm a fan of short parameter list keys. Feel free to ignore this comment. :)

bartgol: Ah, I misread the code. I didn't notice it was "strint-ed" with the parameter index. Then I agree with you.

PR #757, opened 2021-10-28
PyAlbany: Add second order estimator for extreme event probability

kliegeois:
This PR adds the second order estimator for extreme event probability in PyAlbany.

The implementation is tested in the ctest.
The example has been updated to use and compare all the approaches.

@mperego


mperego (approved):
Looks good to me!

PR #760, opened 2021-11-04
Fix STK_DEPRECATED warning from stk_io package

bartgol:
Tunrs out I had this branch ready 4 weeks ago. I checked, and it builds against current trilinos develop.

### DIFF
--- a/src/disc/stk/Albany_IossSTKMeshStruct.cpp
+++ b/src/disc/stk/Albany_IossSTKMeshStruct.cpp
@@ -481,18 +481,20 @@ Albany::IossSTKMeshStruct::setFieldData (
     std::vector<double> ltr;
     int ordering;
     GO stride;
-    boost::any temp_any;
+    stk::util::Parameter temp_any;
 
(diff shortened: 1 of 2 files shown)


mperego (approved):
Looks good to me! thanks!
b.t.w. you should push your branches more often! ;-)

bartgol:
Fortunately, one test failed (FO_GIS_SensSMBwrtBeta_Restart), so I was able to spot that I was still missing one fix. I'm a bit surprised all other tests passed, since this last bug was preventing all restart tests that use Ioss from having a layered mesh numbering. I guess we usually load the 2d mesh and extrude it, even for restarts, which allows the layered mesh numbering to be rebuilt by ExtrudedSTKMeshStruct. In that test, we use Ioss for the 3d mesh, and the inability of loading an important layering parameter caused layered_mesh_numbering to not be created.

Anyhow, all tests pass now on my machine. @mperego if you are ok, I/you can merge.

mperego:
Merged. Thanks @bartgol !

Issue #761, opened 2021-11-13
Fix up FO_AIS tests and get them to run with IOPX


ikalash:
Request by @rstumin .  Was hoping to get help from @mperego and @bartgol to fix the input options so these tests run again.  Code complains about ```Field Origin: Output``` field.

mperego:
I think I fixed the tests. 
However, I'm wondering if requiring Iopx library for using serial mesh is the right thing to do. On my workstation I don't have Iopx but "Use Serial Mesh: true" works fine. @bartgol @jewatkins 

@ikalash courios about the ProSPect$ label... we haven't used it in years...

ikalash:
Thanks @mperego I'll close.  No idea where the ProSPect$ label came from.  I was curious about the $ symbol, but maybe it means "prospect money" (things funded by prospect)?

ikalash:
@mperego : do you know why tests that don't use FRoSch but have it in the input file are failing?  

https://testing.sandia.gov/cdash/viewBuildError.php?buildid=10240241

mperego:
@jewatkins do you understand what's the issue here"
https://sems-cdash-son.sandia.gov/cdash/test/1844066

mperego:
OK, so I "should" have fixed all the issues now, except the Cuda one.

bartgol:
> I'm wondering if requiring Iopx library for using serial mesh is the right thing to do. On my workstation I don't have Iopx but "Use Serial Mesh: true" works fine.

Isn't Iopx used to run decomp and/or to load a partitioned mesh? I wonder if it's not necessary for serial mesh which are partitioned at runtime...

mperego:
> > I'm wondering if requiring Iopx library for using serial mesh is the right thing to do. On my workstation I don't have Iopx but "Use Serial Mesh: true" works fine.
> 
> Isn't Iopx used to run decomp and/or to load a partitioned mesh? I wonder if it's not necessary for serial mesh which are partitioned at runtime...

right, so what's the requirement for "Use Serial Mesh" to work? Based on the dashboard, it seems that the build on Camobap
https://sems-cdash-son.sandia.gov/cdash/viewTest.php?onlyfailed&buildid=25282
cannot handle "Use Serial Mesh". (In this test "Use Serial Mesh" was wrongly set to true independently of the check on Seacas and Iopx).

jewatkins:
Doesn't seacas do the decomp at runtime?

mperego:
[...]

PR #762, opened 2021-11-15
Fix enabling of Use Serial Mesh

mperego:
Check whether parallel ExodusII is available and use it to conditionally enable Use Serial Mesh. Thanks @bartgol for helping.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -188,27 +188,6 @@ if (SET_COMPILERS_AUTOMATICALLY)
   PROJECT(Albany)
 endif()
 
-include(CheckCSourceCompiles)
-set(CMAKE_REQUIRED_LIBRARIES ${${PROJECT_NAME}_EXTRA_LINK_FLAGS})
-set(CMAKE_REQUIRED_INCLUDES  ${Trilinos_INCLUDE_DIRS})
-check_c_source_compiles(
-  "
-  #include <exodusII_par.h>
-  int main()
-  {
-    return 0;
-  }
-  "
-  PARALLEL_EXODUS_SUPPORTED
-  )
-  
-IF (PARALLEL_EXODUS_SUPPORTED)
-  SET(ALBANY_PARALELL_EXODUS TRUE)
-ELSE()
-  SET(ALBANY_PARALELL_EXODUS FALSE)
-ENDIF()
-
-
 # AGS: Adding these lines so Ctest can be run to submit to cdash dashboard
 #   Uses CTestConfig.cmake file for Cdash info.
 ENABLE_TESTING()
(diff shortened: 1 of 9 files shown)


mperego:
This is related to #761.

ikalash:
Thanks @mperego !  I'll test this out on camobap later this afternoon.

mperego:
@ikalash Thanks!
B.t.w. It appears that a couple of tests

> CahnHillElast2D_Epetra
> CahnHillElast2D_Tempus

were if-guarded by Albany_IOPX and were never enabled. This PR enables them and they are now failing:

>  Error, the parameter {name="Initial Order",type="int",value="0"}
> in the parameter (sub)list "Albany Parameters->Piro->Tempus->Tempus Integrator->Time Step Control"
> was not found in the list of valid parameters!

ikalash (changes requested):
Could you please add a print of whether ALBANY_PARALLEL_EXODUS is true/false in the CMakeLists.txt file?  Would be useful to see this in the configure step, for debugging, etc.

ikalash:
> @ikalash Thanks! B.t.w. It appears that a couple of tests
> 
> > CahnHillElast2D_Epetra
> > CahnHillElast2D_Tempus
> 
> were if-guarded by Albany_IOPX and were never enabled. This PR enables them and they are now failing:
> 
> > Error, the parameter {name="Initial Order",type="int",value="0"}
> > in the parameter (sub)list "Albany Parameters->Piro->Tempus->Tempus Integrator->Time Step Control"
> > was not found in the list of valid parameters!

Thanks for the heads up - I'll fix this.

CMakeLists.txt
```diff
@@ -188,6 +188,27 @@ if (SET_COMPILERS_AUTOMATICALLY)
...
+set(CMAKE_REQUIRED_INCLUDES  ${Trilinos_INCLUDE_DIRS})
+check_c_source_compiles(
+  "
+  #include <exodusII_par.h>
+  int main()
+  {
+    return 0;
+  }
+  "
+  PARALLEL_EXODUS_SUPPORTED
```
bartgol, line 202: Is there a reason not to just use `ALBANY_PARALLEL_EXODUS` as output variable?

bartgol (approved):
Looks good. I have one comment, but not very important.

CMakeLists.txt
mperego: Not really. I thought that  
`Performing Test PARALLEL_EXODUS_SUPPORTED - Success`
was more expressive but at the same time wanted to keep Albany way of naming variables. I can change that.

mperego:
> Could you please add a print of whether ALBANY_PARALLEL_EXODUS is true/false in the CMakeLists.txt file? Would be useful to see this in the configure step, for debugging, etc.

Yes, what message would you like to print? I don't see it done in a consistent way. Something like
`ALBANY_PARALLEL_EXODUS set to True` ?

CMakeLists.txt
bartgol: No, it's fine. I was just curious if there was some weird techincality in the CheckFunctionExists macro maybe. This is fine.

ikalash:
> > Could you please add a print of whether ALBANY_PARALLEL_EXODUS is true/false in the CMakeLists.txt file? Would be useful to see this in the configure step, for debugging, etc.
> 
> Yes, what message would you like to print? I don't see it done in a consistent way. Something like `ALBANY_PARALLEL_EXODUS set to True` ?

Yes exactly - to be consistent with the other messages printed when you configure Albany.

bartgol (approved):

ikalash:
This looks good.  I just verified that everything is OK on camobap with your change.  I'll go ahead and merge.

mperego:
@ikalash thanks for checking camobap. I usually prefer to rebase and merge to avoid useless merge commits.

B.t.w. the failing CahnHill tests don't appear on Camobap because ALBANY_PARALLEL_EXODUS is false there.

ikalash:
@mperego , yes, thank you for the reminder.  I'll fix it in master.

PR #764, opened 2021-11-30
Put git info in a separate config file

bartgol:
Premise: this bugged me for a whole afternoon, since I wasn't able to figure out why Albany sometimes was fully rebuilt.

When making a commit during development, the sha of the HEAD commit changes. If, by chance, cmake was triggered (e.g., change in an input file), this changes the Albany_config.h file int he bld dir, which triggers a rebuild of the whole library.

Step to reproduce:

- do minor mod to input file (e.g. test/small/SteadyHeat1D/input.yaml)
- commit micro mod
- run make in build dir

Git info is only used in `Albany_Utils.cpp`. By putting all git version info in a separate `Albany_GitVersion.h.in` file, and including `Albany_GitVersion.h` in `Albany_Utils.cpp`, we allow the rest of Albany to not be rebuilt when re-running CMake after making a commit.

@jewatkins I'm pinging you since I think you added the logic for the GIT info in the albany banner.

### DIFF
--- a/src/Albany_GitVersion.h.in
+++ b/src/Albany_GitVersion.h.in
@@ -1,0 +1,9 @@
+#ifndef ALBANY_GIT_VERSION_H
+#define ALBANY_GIT_VERSION_H
+
+// Trilinos/Albany version details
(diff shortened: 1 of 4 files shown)


jewatkins (approved):
This makes sense to me and is a nice addition, thanks!

PR #766, opened 2021-12-08
Add non Gaussian random variables

kliegeois:
This PR adds the `RandomPhysicalParameter`, a need type of parameter deduced from a Gaussian random variable and a mapping between the two.

This is used in the extreme event computation to have non-Gaussian random variables.

The tests passed on my workstation.

@mperego

### DIFF
--- a/src/LandIce/problems/LandIce_ParamEnum.hpp
+++ b/src/LandIce/problems/LandIce_ParamEnum.hpp
@@ -19,7 +19,9 @@ enum class ParamEnum
   Mu,
   Power,
   Homotopy,
-  GLHomotopy
+  GLHomotopy,
+  Theta_0,
+  Theta_1
 };
 
 namespace ParamEnumName
@@ -30,6 +32,8 @@ namespace ParamEnumName
   static const std::string Power           = "Power Exponent";
   static const std::string HomotopyParam   = "Homotopy Parameter";
   static const std::string GLHomotopyParam = "Glen's Law Homotopy Parameter";
+  static const std::string theta_0         = "Theta 0"; 
+  static const std::string theta_1         = "Theta 1";
 } // ParamEnum
 
 } // Namespace LandIce
--- a/src/LandIce/problems/LandIce_StokesFOBase.hpp
+++ b/src/LandIce/problems/LandIce_StokesFOBase.hpp
@@ -52,6 +52,9 @@
 #include "LandIce_Dissipation.hpp"
 #include "LandIce_UpdateZCoordinate.hpp"
 
+#include "PHAL_RandomPhysicalParameter.hpp"
+#include "PHAL_IsAvailable.hpp"
+
 #include <string.hpp> // For util::upper_case (do not confuse this with <string>! string.hpp is an Albany file)
 
 //uncomment the following line if you want debug output to be printed to screen
@@ -967,6 +970,48 @@ constructVelocityEvaluators (PHX::FieldManager<PHAL::AlbanyTraits>& fm0,
   ev = Teuchos::rcp(new StokesFOResid<EvalT,PHAL::AlbanyTraits>(*p,dl));
   fm0.template registerEvaluator<EvalT>(ev);
 
+  //--- Shared Parameter for Extreme Event ---//
+
+  p = rcp(new Teuchos::ParameterList("Theta 0"));
+  p->set< Teuchos::RCP<ParamLib> >("Parameter Library", paramLib);
+  param_name = "Theta 0";
+  p->set<std::string>("Parameter Name", param_name);
+  p->set<const Teuchos::ParameterList*>("Parameters List", &params->sublist("Parameters"));
+  p->set<double>("Default Nominal Value", 0.);
+  Teuchos::RCP<PHAL::SharedParameter<EvalT,PHAL::AlbanyTraits,ParamEnum,ParamEnum::Theta_0>> ptr_theta_0;
+  ptr_theta_0 = Teuchos::rcp(new PHAL::SharedParameter<EvalT,PHAL::AlbanyTraits,ParamEnum,ParamEnum::Theta_0>(*p,dl));
+  fm0.template registerEvaluator<EvalT>(ptr_theta_0);
+
+  p = rcp(new Teuchos::ParameterList("Theta 1"));
+  p->set< Teuchos::RCP<ParamLib> >("Parameter Library", paramLib);
+  param_name = "Theta 1";
+  p->set<std::string>("Parameter Name", param_name);
+  p->set<const Teuchos::ParameterList*>("Parameters List", &params->sublist("Parameters"));
+  p->set<double>("Default Nominal Value", 0.);
+  Teuchos::RCP<PHAL::SharedParameter<EvalT,PHAL::AlbanyTraits,ParamEnum,ParamEnum::Theta_1>> ptr_theta_1;
+  ptr_theta_1 = Teuchos::rcp(new PHAL::SharedParameter<EvalT,PHAL::AlbanyTraits,ParamEnum,ParamEnum::Theta_1>(*p,dl));
+  fm0.template registerEvaluator<EvalT>(ptr_theta_1);
+
+  auto rparams = params->sublist("Random Parameters");
+  if (rparams.isParameter("Number Of Parameters")) {
+    int nrparams = rparams.get<int>("Number Of Parameters");
+    for (int i_rparams=0; i_rparams<nrparams; ++i_rparams) {
+      auto rparams_i = rparams.sublist(Albany::strint("Parameter",i_rparams));
+
+      p = rcp(new Teuchos::ParameterList("Theta 1"));
+      p->set< Teuchos::RCP<ParamLib> >("Parameter Library", paramLib);
+      const std::string param_name = rparams_i.get<std::string>("Name");
+      p->set<std::string>("Parameter Name", param_name); //output name
+      const std::string rparam_name = rparams_i.get<std::string>("Standard Normal Parameter");
+      p->set<std::string>("Random Parameter Name", rparam_name); //input name
+      p->set<const Teuchos::ParameterList*>("Parameters List", &params->sublist("Parameters"));
+      p->set<const Teuchos::ParameterList*>("Distribution", &rparams_i.sublist("Distribution"));
+      Teuchos::RCP<PHAL::RandomPhysicalParameter<EvalT,PHAL::AlbanyTraits>> ptr_rparam;
+      ptr_rparam = Teuchos::rcp(new PHAL::RandomPhysicalParameter<EvalT,PHAL::AlbanyTraits>(*p,dl));
+      fm0.template registerEvaluator<EvalT>(ptr_rparam);
+    }
+  }
+
   //--- Shared Parameter for Continuation: Glen's Law Homotopy Parameter ---//
   p = Teuchos::rcp(new Teuchos::ParameterList("Homotopy Parameter"));
 
@@ -976,9 +1021,11 @@ constructVelocityEvaluators (PHX::FieldManager<PHAL::AlbanyTraits>& fm0,
   p->set<const Teuchos::ParameterList*>("Parameters List", &params->sublist("Parameters"));
   p->set<double>("Default Nominal Value", params->sublist("LandIce Viscosity").get<double>(param_name,-1.0));
 
-  Teuchos::RCP<PHAL::SharedParameter<EvalT,PHAL::AlbanyTraits,ParamEnum,ParamEnum::GLHomotopy>> ptr_gl_homotopy;
-  ptr_gl_homotopy = Teuchos::rcp(new PHAL::SharedParameter<EvalT,PHAL::AlbanyTraits,ParamEnum,ParamEnum::GLHomotopy>(*p,dl));
-  fm0.template registerEvaluator<EvalT>(ptr_gl_homotopy);
+  if(!PHAL::is_param_available<EvalT>(fm0, param_name, dl->shared_param)) {
+    Teuchos::RCP<PHAL::SharedParameter<EvalT,PHAL::AlbanyTraits,ParamEnum,ParamEnum::GLHomotopy>> ptr_gl_homotopy;
+    ptr_gl_homotopy = Teuchos::rcp(new PHAL::SharedParameter<EvalT,PHAL::AlbanyTraits,ParamEnum,ParamEnum::GLHomotopy>(*p,dl));
+    fm0.template registerEvaluator<EvalT>(ptr_gl_homotopy);
+  }
 
   //--- LandIce Flow Rate ---//
   auto& visc_pl = params->sublist("LandIce Viscosity");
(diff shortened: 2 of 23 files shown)


src/LandIce/problems/LandIce_StokesFOBase.hpp
```diff
@@ -967,6 +970,48 @@ constructVelocityEvaluators (PHX::FieldManager<PHAL::AlbanyTraits>& fm0,
...
+  p->set<std::string>("Parameter Name", param_name);
+  p->set<const Teuchos::ParameterList*>("Parameters List", &params->sublist("Parameters"));
+  p->set<double>("Default Nominal Value", 0.);
+  Teuchos::RCP<PHAL::SharedParameter<EvalT,PHAL::AlbanyTraits,ParamEnum,ParamEnum::Theta_1>> ptr_theta_1;
+  ptr_theta_1 = Teuchos::rcp(new PHAL::SharedParameter<EvalT,PHAL::AlbanyTraits,ParamEnum,ParamEnum::Theta_1>(*p,dl));
+  fm0.template registerEvaluator<EvalT>(ptr_theta_1);
+
+  auto rparams = params->sublist("Random Parameters");
+  if (rparams.isParameter("Number Of Parameters")) {
+    int nrparams = rparams.get<int>("Number Of Parameters");
```
bartgol, line 997: Minor but I would just do
```
int nrparams = rparams.get("Number Of Parameters",0);
```
and avoid the if statement, saving one level of indentation.

src/evaluators/utility/PHAL_IsAvailable.hpp
```diff
@@ -0,0 +1,46 @@
...
+//    in the file "license.txt" in the top-level Albany directory  //
+//*****************************************************************//
+
+#ifndef PHAL_IS_PARAM_AVAILABLE_HPP
+#define PHAL_IS_PARAM_AVAILABLE_HPP
+
+namespace PHAL
+{
+
+enum class FieldScalarType : int {
```
bartgol, line 13: It doesn't look like you use this enum in this file. Maybe leftover from copy+paste of LandIce_ProblemUtils.hpp?

src/evaluators/utility/PHAL_IsAvailable.hpp
```diff
@@ -0,0 +1,46 @@
...
+
+template<typename EvalT>
+Teuchos::RCP<PHX::FieldTag>
+createTag(const std::string& name, const Teuchos::RCP<PHX::DataLayout>& dl)
+{
+  return Teuchos::rcp(new PHX::Tag<typename EvalT::ScalarT>(name,dl));
+}
+
+template<typename EvalT>
+bool is_param_available (const PHX::FieldManager<PHAL::AlbanyTraits>& fm,
```
bartgol, line 28: It looks like this fcn can look for any kind of field, not just parameters. Wouldn't it be more apt to call it `is_field_evaluated`? Same change for the ifdef.

src/evaluators/utility/PHAL_IsAvailable.hpp
```diff
@@ -0,0 +1,46 @@
...
+  MeshScalar  = 1,
+  ParamScalar = 2,
+  Scalar      = 3
+};
+
+template<typename EvalT>
+Teuchos::RCP<PHX::FieldTag>
+createTag(const std::string& name, const Teuchos::RCP<PHX::DataLayout>& dl)
+{
+  return Teuchos::rcp(new PHX::Tag<typename EvalT::ScalarT>(name,dl));
```
bartgol, line 24: Wouldn't it be simpler to template on the scalar type itself, and let the user do `EvalT::ScalarT` at the call place, if need be? That would allow to create a tag with, say, Real scalar type during EvalT=Jacobian...

src/LandIce/problems/LandIce_ParamEnum.hpp
```diff
@@ -19,7 +19,9 @@ enum class ParamEnum
   Mu,
   Power,
   Homotopy,
-  GLHomotopy
+  GLHomotopy,
+  Theta_0,
```
bartgol, line 23: You already have Theta_0 and Theta_1 in the common Albany::ParamEnum... We could just use those, no?

src/LandIce/problems/LandIce_StokesFOBase.hpp
```diff
@@ -976,9 +1021,11 @@ constructVelocityEvaluators (PHX::FieldManager<PHAL::AlbanyTraits>& fm0,
   p->set<const Teuchos::ParameterList*>("Parameters List", &params->sublist("Parameters"));
   p->set<double>("Default Nominal Value", params->sublist("LandIce Viscosity").get<double>(param_name,-1.0));
 
-  Teuchos::RCP<PHAL::SharedParameter<EvalT,PHAL::AlbanyTraits,ParamEnum,ParamEnum::GLHomotopy>> ptr_gl_homotopy;
-  ptr_gl_homotopy = Teuchos::rcp(new PHAL::SharedParameter<EvalT,PHAL::AlbanyTraits,ParamEnum,ParamEnum::GLHomotopy>(*p,dl));
-  fm0.template registerEvaluator<EvalT>(ptr_gl_homotopy);
+  if(!PHAL::is_param_available<EvalT>(fm0, param_name, dl->shared_param)) {
```
bartgol, line 1024: Can you also move the param list creation/setup lines inside the if block?

src/LandIce/problems/LandIce_StokesFOBase.hpp
```diff
@@ -1299,9 +1346,11 @@ void StokesFOBase::constructBasalBCEvaluators (PHX::FieldManager<PHAL::AlbanyTra
                                pl->sublist("Basal Friction Coefficient").get<double>(param_name) : -1.0;
       p->set<double>("Default Nominal Value", nominalValue);
 
-      Teuchos::RCP<PHAL::SharedParameter<EvalT,PHAL::AlbanyTraits,ParamEnum,ParamEnum::Alpha>> ptr_alpha;
-      ptr_alpha = Teuchos::rcp(new PHAL::SharedParameter<EvalT,PHAL::AlbanyTraits,ParamEnum,ParamEnum::Alpha>(*p,dl));
-      fm0.template registerEvaluator<EvalT>(ptr_alpha);
+      if(!PHAL::is_param_available<EvalT>(fm0, param_name, dl->shared_param)) {
```
bartgol, line 1349: Same here, move all plist setup in the if block.

src/LandIce/problems/LandIce_StokesFOBase.hpp
```diff
@@ -1313,9 +1362,11 @@ void StokesFOBase::constructBasalBCEvaluators (PHX::FieldManager<PHAL::AlbanyTra
     p->set<const Teuchos::ParameterList*>("Parameters List", &params->sublist("Parameters"));
     p->set<double>("Default Nominal Value", pl->sublist("Basal Friction Coefficient").get<double>(param_name,-1.0));
 
-    Teuchos::RCP<PHAL::SharedParameter<EvalT,PHAL::AlbanyTraits,ParamEnum,ParamEnum::Lambda>> ptr_lambda;
-    ptr_lambda = Teuchos::rcp(new PHAL::SharedParameter<EvalT,PHAL::AlbanyTraits,ParamEnum,ParamEnum::Lambda>(*p,dl));
-    fm0.template registerEvaluator<EvalT>(ptr_lambda);
+    if(!PHAL::is_param_available<EvalT>(fm0, param_name, dl->shared_param)) {
```
bartgol, line 1365: ...and here.

src/LandIce/problems/LandIce_StokesFOBase.hpp
```diff
@@ -1326,9 +1377,11 @@ void StokesFOBase::constructBasalBCEvaluators (PHX::FieldManager<PHAL::AlbanyTra
     p->set<const Teuchos::ParameterList*>("Parameters List", &params->sublist("Parameters"));
     p->set<double>("Default Nominal Value", pl->sublist("Basal Friction Coefficient").get<double>(param_name,-1.0));
 
-    Teuchos::RCP<PHAL::SharedParameter<EvalT,PHAL::AlbanyTraits,ParamEnum,ParamEnum::Mu>> ptr_mu;
-    ptr_mu = Teuchos::rcp(new PHAL::SharedParameter<EvalT,PHAL::AlbanyTraits,ParamEnum,ParamEnum::Mu>(*p,dl));
-    fm0.template registerEvaluator<EvalT>(ptr_mu);
+    if(!PHAL::is_param_available<EvalT>(fm0, param_name, dl->shared_param)) {
```
bartgol, line 1380: ...and here.

src/LandIce/problems/LandIce_StokesFOBase.hpp
```diff
@@ -1342,9 +1395,11 @@ void StokesFOBase::constructBasalBCEvaluators (PHX::FieldManager<PHAL::AlbanyTra
     double default_val = (type == "FIELD") ? 1.0 : beta_list.get<double>(param_name, -1.0);
     p->set<double>("Default Nominal Value", default_val);
 
-    Teuchos::RCP<PHAL::SharedParameter<EvalT,PHAL::AlbanyTraits,ParamEnum,ParamEnum::Power>> ptr_power;
-    ptr_power = Teuchos::rcp(new PHAL::SharedParameter<EvalT,PHAL::AlbanyTraits,ParamEnum,ParamEnum::Power>(*p,dl));
-    fm0.template registerEvaluator<EvalT>(ptr_power);
+    if(!PHAL::is_param_available<EvalT>(fm0, param_name, dl->shared_param)) {
```
bartgol, line 1398: ...and here

src/evaluators/utility/PHAL_UnivariateDistribution.hpp
```diff
@@ -0,0 +1,300 @@
+//*****************************************************************//
+//    Albany 3.0:  Copyright 2016 Sandia Corporation               //
+//    This Software is released under the BSD license detailed     //
+//    in the file "license.txt" in the top-level Albany directory  //
+//*****************************************************************//
+
+#ifndef ALBANY_UNIVARIATEDISTRIBUTION_HPP
```
bartgol, line 7: Would you mind separating the words with an underscore? It makes the ifdef more readable. Also, throughout albany,

src/evaluators/utility/PHAL_UnivariateDistribution.hpp
```diff
@@ -0,0 +1,300 @@
...
+#define ALBANY_UNIVARIATEDISTRIBUTION_HPP
+
+#include <boost/math/special_functions/erf.hpp>
+
+namespace Albany
```
bartgol, line 12: Since this is not an evaluator (and not in PHAL namespace), maybe the file should be called Albany_UnivariateDistribution.hpp, and perhaps put in `src/utility`, rather than the evaluators folder?

src/problems/Albany_ThermalProblem.hpp
```diff
@@ -226,9 +276,11 @@ Albany::ThermalProblem::constructEvaluators(
     p->set<std::string>("Parameter Name", param_name);
     p->set<const Teuchos::ParameterList*>("Parameters List", &params->sublist("Parameters"));
     p->set<double>("Default Nominal Value", kappa[0]);
-    RCP<PHAL::SharedParameter<EvalT,PHAL::AlbanyTraits,Albany::ParamEnum,Albany::ParamEnum::Kappa_x>> ptr_kappa_x;
-    ptr_kappa_x = rcp(new PHAL::SharedParameter<EvalT,PHAL::AlbanyTraits,Albany::ParamEnum,Albany::ParamEnum::Kappa_x>(*p,dl));
-    fm0.template registerEvaluator<EvalT>(ptr_kappa_x);
+    if(!PHAL::is_param_available<EvalT>(fm0, param_name, dl->shared_param)) {
```
bartgol, line 279: Like in LandIce, all the plist construction/setup for all params can be moved inside their if blocks.

Note: this has nothing to do with perf, obviously. It just makes intent clearer, I think.

bartgol (commented):
Looks mostly good. I got some thoughts, suggestions. Feel free to disregard stuff.

kliegeois:
Thanks @bartgol for the super fast review! I am going to update the PR asap based on your comments.

src/LandIce/problems/LandIce_StokesFOBase.hpp
mperego: Do we want to default to 0, or do we want to fail if "Number Of Parameters" is not defined? I prefer the latter, so I would remove the if check and just do
`int nrparams = rparams.get<int>("Number Of Parameters");`

src/evaluators/bc/PHAL_Dirichlet_Def.hpp
```diff
@@ -26,9 +27,26 @@ DirichletBase(Teuchos::ParameterList& p) :
   value = p.get<RealType>("Dirichlet Value");
 
   std::string name = p.get< std::string >("Dirichlet Name");
+  isRandom = false;
+  if (p.isParameter("Random"))
+    isRandom = p.get< bool >("Random");
   const Teuchos::RCP<PHX::DataLayout> dummy = p.get< Teuchos::RCP<PHX::DataLayout> >("Data Layout");
   const PHX::Tag<ScalarT> fieldTag(name, dummy);
 
+  if(isRandom) {
```
mperego, line 36: This is a bit of a hack. Is there a way we can move this code out of this evaluator?

bartgol: I agree. But I think the only alternative is to create a new Dirichlet ev type, something like `DirichletField`, but that already exist (though we could rename that one to `DirichletDistributedField`). Then, modify the Dirichlet factory, so that it can build the SharedParam and RandomPhysicalParam evaluators. That's however hard to do with current yaml syntax. We'd need something like
```
DBC on NodeSet blah for DOF XYZ prescribe Field: [random_phys_name, shared_param_name]
```
so that the factory can auto build the SharedParam ev for $shared_param_name, and the RandomPhysicalParameter for $random_phys_name...

P.s: I've been willing to re-do how our DBC/NBC are prescribed, by having sublists, rather than encoding bc type, ns name, and dof name all in the yaml parameter entry... With something like
```
   DBC:
      Number DBCs: 2
      DBC 0:
         Type: prescribe field
         DOF: xyz
         Node Set: blah
         Field Name: beta
      DOF 1:
         Type: Value
          DOF: xyz
          Node Set: blah2
          Value: 0.0
```
It would be less hacky to implement this new DBC as a new evaluator.

src/LandIce/problems/LandIce_ParamEnum.hpp
kliegeois: It seems that it is still needed; I changed all the rest based on your comments though.

src/evaluators/bc/PHAL_Dirichlet_Def.hpp
kliegeois: I like this idea. I can look into that but I will not be able to do so before my holiday.
If preferred, we can leave the PR open and I can do that inside this PR.

src/LandIce/problems/LandIce_StokesFOBase.hpp
kliegeois: I updated the PR to be as suggested by @mperego .

bartgol: I'm not sure. What you suggest forces to either have a full valid parameter list, or to not have it at all. I don't see a problem having, say, an empty sublist in the yaml file, which can be parsed as "no request for random phys params"... But I'm ok either way.

src/evaluators/bc/PHAL_Dirichlet_Def.hpp
bartgol: I don't like PRs open for too long. If you plan to do the change, you can leave things as they are, and do a follow up PR soon.

mperego: Agreed

src/LandIce/problems/LandIce_StokesFOBase.hpp
mperego: The issue is when you just forget to put the "Number Of Parameters", and you think Albany is including the parameters but in fact it does not.

src/evaluators/bc/PHAL_Dirichlet_Def.hpp
kliegeois: OK, I will work on that in another PR then asap but probably after Christmas. Let me know if you want any other modifications in this PR.

PR #769, opened 2021-12-10
Remove atomic from Kokkos::Serial execution

### LINKED ISSUE #767: Performance regression on blake
I noticed a performance regression on blake that seems to be pretty significant (somewhere around 20% slowdown depending on the case). Unfortunately, I don't have a lot of data because the performance tests were down during that period but it happened between Oct. 26th and Nov. 6th and it seems to be related to the boundary conditions and the assembly:
Oct. 26th Run:
```
** Trilinos git commit id - 3bae707
** Albany git commit id --- 323a90b
|   |   |   |   Albany: Total Fill Time: 12.6747 - 99.895% [201] {min=12.5964, max=12.7485, std dev=0.0354548}
|   |   |   |   |   |   |   Phalanx: Evaluator 86: [<Jacobian>] StokesFOLateralResid<Jacobian>: 0.0241899 - 0.249467% [100] {min=0.000457268, max=0.177912, std dev=0.0314022}
|   |   |   |   |   |   |   Phalanx: Evaluator 87: [<Jacobian>] StokesFOBasalResid<Jacobian>: 0.0732972 - 0.755904% [100] {min=0.0612907, max=0.104363, std dev=0.00537095}
|   |   |   |   |   |   |   Phalanx: Evaluator 68: [<Jacobian>] Scatter Velocity Residual<Jacobian>: 2.58533 - 26.6621% [100] {min=2.46749, max=2.75813, std dev=0.0644871}
|   |   |   |   |   |   Albany Jacobian Fill: Export: 0.81807 - 7.53388% [100] {min=0.486474, max=1.10587, std dev=0.144271}
```
Nov. 6th Run:
```
** Trilinos git commit id - e15cf42
** Albany git commit id --- e14f44a
|   |   |   |   Albany: Total Fill Time: 15.1982 - 99.909% [201] {min=15.082, max=15.3161, std dev=0.0476297}
|   |   |   |   |   |   |   Phalanx: Evaluator 86: [<Jacobian>] StokesFOLateralResid<Jacobian>: 0.147669 - 1.31135% [100] {min=0.000441569, max=1.17069, std dev=0.203318}
|   |   |   |   |   |   |   Phalanx: Evaluator 87: [<Jacobian>] StokesFOBasalResid<Jacobian>: 0.116388 - 1.03357% [100] {min=0.103936, max=0.128278, std dev=0.00484907}
|   |   |   |   |   |   |   Phalanx: Evaluator 68: [<Jacobian>] Scatter Velocity Residual<Jacobian>: 3.95909 - 35.1579% [100] {min=3.7384, max=4.11777, std dev=0.0938339}
|   |   |   |   |   |   Albany Jacobian Fill: Export: 1.47505 - 11.272% [100] {min=0.553215, max=2.04266, std dev=0.258961}
```
There was also a regression in the residual. This doesn't seem to show up in the GPU tests, it seems to be CPU-only. I don't see anything significant in the history other than the STK deprecation changes.

@mperego @bartgol Any ideas? Was there something to the STK changes that would cause this? We may just have to let this one slide because I don't see anything obvious...

jewatkins:
This PR fixes #767. I created some helper structs that decide whether to use an atomic based on execution space (I only avoid the atomic for Kokkos::Serial). I tested it on one of the performance tests on blake and it fixes the performance regression. I'll try testing clang and weaver/cuda.

Albany_KokkosUtils.hpp was a rewrite because it wasn't a unix file...

### DIFF
--- a/src/utility/Albany_KokkosUtils.hpp
+++ b/src/utility/Albany_KokkosUtils.hpp
@@ -10,16 +10,8 @@
 namespace KU
 {
 
+// Check device memory
 #ifdef KOKKOS_ENABLE_CUDA
-
-template <typename T>
-KOKKOS_INLINE_FUNCTION
-const T& max (const T& a, const T& b) { return a > b ? a : b; }
-
-template <typename T>
-KOKKOS_INLINE_FUNCTION
-const T& min (const T& a, const T& b) { return a < b ? a : b; }
-
 inline bool
 IsNearDeviceMemoryLimit()
 {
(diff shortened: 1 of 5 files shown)


src/utility/Albany_KokkosUtils.hpp
```diff
@@ -1,53 +1,103 @@
...
+using std::min;
+#endif
+using Sacado::Fad::max;
+using Sacado::Fad::min;
+using Sacado::Fad::Exp::max;
+using Sacado::Fad::Exp::min;
+
+// Choose atomic based on execution space
+template <typename ExeSpace>
+struct IsAtomic
```
bartgol, line 51: Very subjective (i.e., feel free to leave like this): I think `NeedsAtomic` would be more suited than `IsAtomic`, since it makes no sense to say that the exe space is atomic.

bartgol (approved):
Looks good. My very subjective comment is not that important.

P.s: it might be worth pointing out the slowdown to the Tpetra and/or KokkosKernels team(s). They might be interested in hard-coding a bypass of atomic for Serial, much like we did.

src/utility/Albany_KokkosUtils.hpp
jewatkins: This makes sense, thanks! I'll make the change in push. I tested this on clang/weaver and it works fine.

PR #770, opened 2021-12-16
Add a test for FO+Hydrology

bartgol:
Most relevant changes:

- fixes a bug in GatherSolutionSide
- enable FO_Hydrology test (always)
- improvements to AddNoise ev
- adds Kappa as parameter in LandIce_ParamEnum.hpp


mperego (approved):
Sorry it took me a while. Looks good to me. Thanks!

PR #779, opened 2022-01-14
Mods to parallel exodus check at config time

bartgol:
* Add trilinos' tpl include dirs, to ensure we pick up the tpls headers that trilinos used, and not some system folder ones.
* Check that NC was configured with both HDF5 *and* PnetCDF, to ensure that parallel IO works, regardless of exo file type.


mperego (approved):
Looks good. Thanks for fixing this!!

Issue #780, opened 2022-01-17
Fix evaluateResponse_HessVecProd_pp for non-contiguous map


kliegeois:
Related to https://github.com/trilinos/Trilinos/issues/9500 .

There is one known occurrence of a reconstructed Hessian which is not symmetrical for non-contiguous map of the distributed parameter.
It can be deduced from previous investigation that the issue is related to `evaluateResponse_HessVecProd_pp`.

PR #782, opened 2022-01-27
Solves #780

kliegeois:
This PR solves #780 by changing the stride of the extruded meshes to take into account a non-zero first global index.

The PR adds new tests.

Thanks @bartgol for the discussion!


src/disc/stk/Albany_ExtrudedSTKMeshStruct.cpp
```diff
@@ -327,7 +327,7 @@ void Albany::ExtrudedSTKMeshStruct::setBulkData(
   Teuchos::RCP<const Tpetra_Map> sides_map = Tpetra::createNonContigMapWithNode<LO, Tpetra_GO, KokkosNode>(indices(),comm);
 
   GO globalElemStride = cells_map->getMaxAllGlobalIndex() - cells_map->getMinAllGlobalIndex() + 1;
-  GO globalVerticesStride = nodes_map->getMaxAllGlobalIndex() - nodes_map->getMinAllGlobalIndex() + 1;
+  GO globalVerticesStride = nodes_map->getMaxAllGlobalIndex() + 1;
```
mperego, line 330: I think that the definition of the globalVerticesStride is correct. Maybe it's not used as it should.

mperego (changes requested):
Thanks Kim. I think that the stride definition is correct. My guess is that it is not used as it should.
Also if you change the definition of the GlobalVerticesStride you should change by consistensy all the other strides (e.g. the elem stride and the side strids). But I don't think it is needed.

bartgol:
@mperego I think the changes _are_ needed in order for the layered numbering to be correct. As an example, consider a 2d mesh with 2 nodes, with gids 9 and 10. The current master would set stride=2, and therefore, the gids of the 2nd level would be 11, 12. Doing mod arithmetic to retrieve base_id from 11, 12 and stride 2 would return 1,2, which is wrong. The correct stride is 10, which gives gids 19,20 on the second level.

mperego:
> @mperego I think the changes _are_ needed in order for the layered numbering to be correct. As an example, consider a 2d mesh with 2 nodes, with gids 9 and 10. The current master would set stride=2, and therefore, the gids of the 2nd level would be 11, 12. Doing mod arithmetic to retrieve base_id from 11, 12 and stride 2 would return 1,2, which is wrong. The correct stride is 10, which gives gids 19,20 on the second level.

OK, I see. Then we should fix all the strides in the Extruded mesh struct for consistency.

bartgol:
Yes, definitely. @kliegeois can you fix all the strides as part of this PR?

kliegeois:
@bartgol  @mperego  I am rebuilding with those 2 changes and run test.

mperego (approved):
Looks good. Thanks for fixing it.

mperego:
> > @mperego I think the changes _are_ needed in order for the layered numbering to be correct. As an example, consider a 2d mesh with 2 nodes, with gids 9 and 10. The current master would set stride=2, and therefore, the gids of the 2nd level would be 11, 12. Doing mod arithmetic to retrieve base_id from 11, 12 and stride 2 would return 1,2, which is wrong. The correct stride is 10, which gives gids 19,20 on the second level.
> 
> OK, I see. Then we should fix all the strides in the Extruded mesh struct for consistency.

I'm actually surprised that we didn't have other issues because of this.

bartgol:
@mperego  We probably never had a basal mesh with min gid > 0.

mperego:
> @mperego We probably never had a basal mesh with min gid > 0.

I think most mesh coming form MPAS are like that. And the Humbold mesh worked fine, other than the Hessian coloring thing.

bartgol:
I don't know then. :/

PR #784, opened 2022-01-31
Regression testing and other fixes

mperego:
This branch fixes the sensitivity regression testing.

Apparently a lot of tests had sensitivity tests that were not executed because the sensitivity were not computed or because of errors in the input files. We now detect more carefully errors in the input files.
Several tests had sensitivity regression results different from the ones that are currently computed so I had to re-bless many of them (it's practically impossible to track at what point the tests stopped being executed and the tests started giving different results) and I removed some of the sensitivity checks (e.g. in the manufacutured solution tests, I removed the check for sensitivities since they were not based on analytic results, in my understanding).  

This PR also adds a couple of fixes to the computation of L2 error side response and flux divergence evaluators  (sorry for adding them to the same PR).


mperego:
@ikalash For Tempus, I've seen that you had some logic to check whether the sensitivity was computed for a given response/parameter. I removed that logic. We always check whether the sensitivity is nonnull, so it should be OK even if Tempus doesn't compute all the sensitivities. B.t.w. I noticed that all the Tempus tests have only one response, it might be good to add a tests with more responses to make sure that the logic works.

tests/small/LandIce/FO_GIS/input_fo_humboldt_frosch_fluxdiv.yaml
```diff
@@ -45,26 +46,21 @@ ANONYMOUS:
     LandIce Viscosity:
       Extract Strain Rate Sq: true
       Type: Glen's Law
-      Glen's Law Homotopy Parameter: 0.3e+00
+      Glen's Law Homotopy Parameter: 0.3
       Continuous Homotopy With Constant Initial Viscosity: true
-      Coefficient For Continuous Homotopy: 8.00000000000000000e+00
+      Coefficient For Continuous Homotopy: 8.0
```
bartgol, line 51: Thanks for trimming all of these (overly) long decimals! =)

bartgol (approved):
Looks much cleaner.

tests/small/LandIce/FO_GIS/input_fo_humboldt_frosch_fluxdiv.yaml
mperego: Sure. I did that only for the tests that I had to modify. If each of us trim those when working on a particular test at some point we'll get rid of those long decimals. We could make a script to fix them.. but it's not always trivial, and we do not necessarily want all the numbers in the same format.

bartgol: I don't think it's worth the time to write a script. As you said, we can just trim decimals as we work on tests.

ikalash:
I tested this just now.  CISM-Albany is good.  There is one failing test - FO_Test_Dome_interpSurf:

```
Finished eval of first model: Params, Responses 
 
 Parameter vector 0:
 
                     Glen's Law Homotopy Parameter	3.000000000000e-01
 
 Response 0: Solution Max Value
 
                     9.940412288601e+01
 
 Sensitivities (0,0):
 
                     -6.592799261249e-04
 Sensitivity (0,0), Test 0,0: -6.592799261249e-04 != -1.318560000000e-03 (rel 1.000000000000e-04 abs 1.000000000000e-04)
 
 Response 1: Solution Max Value
 
                     9.940412288601e+01
 
 Sensitivities (1,0):
 
                     -6.592800805675e-04
 Sensitivity (1,0), Test 0,0: -6.592800805675e-04 != -1.318560006692e-03 (rel 1.000000000000e-04 abs 1.000000000000e-04)
 
 Response 2: Solution Average
 
                     -7.011281728917e-15
 
 Sensitivities (2,0):
 
                     -8.530690802193e-13
 
 Main_Solve: MeanValue of final solution -7.011281728917e-15
 
 Number of Comparisons Attempted: 6
 Number of Failed Comparisons: 2
```

mperego:
Thanks for checking @ikalash .
As for the failing test, I think we have seen this before. The Max response is not differentiable and it's sensitivity strongly depends on the solver tolerance. I just verified that changing the tolerance has a big impact on the sensitivity. I'm going to remove the sensitivity tests for the Max response. Then I'll merge.

mperego:
@ikalash I just committed a fix for compiler warnings and an issue with fad of sqrt(0) that would give NaN. This should take care of the failings on the Albany dashboard.

mperego:
@jewatkins I should have fixed the failures in the perf tests as well.

jewatkins:
> @ikalash I just committed a fix for compiler warnings and an issue with fad of sqrt(0) that would give NaN. This should take care of the failings on the Albany dashboard.

There's also a Sacado `safe_sqrt` that gets rid of the NaN. We could put it under the KU namespace.

mperego:
> 



> > @ikalash I just committed a fix for compiler warnings and an issue with fad of sqrt(0) that would give NaN. This should take care of the failings on the Albany dashboard.
> 
> There's also a Sacado `safe_sqrt` that gets rid of the NaN. We could put it under the KU namespace.

We could but then we would not compute the right derivative at 0, whereas using sqt(eps) the derivative will be almost correct.

bartgol:
> > 
> 
> > > @ikalash I just committed a fix for compiler warnings and an issue with fad of sqrt(0) that would give NaN. This should take care of the failings on the Albany dashboard.
> > 
> > 
> > There's also a Sacado `safe_sqrt` that gets rid of the NaN. We could put it under the KU namespace.
> 
> We could but then we would not compute the right derivative at 0, whereas using sqt(eps) the derivative will be almost correct.

Do we need the right derivative at 0? If we ever find ourselves needing the derivative of sqrt(0), then our simulation is probably already garbage, no?

mperego:
You are right, I got confused. We should not need the right derivative at 0.

The issue is that I used an algorithm to compute the flux divergence based on a FV approach that relies on mesh triangles being acute, which is typically true for mpas meshes, but not for the mesh in question. I'll modify the test when I have time and throw an error when the acute triangle condition does not hold.

PR #786, opened 2022-02-03
Update cmake minimum requirement

jewatkins:
Update cmake minimum requirement to be compatible with Trilinos. I don't think this should affect anyone since most of us use up-to-date Trilinos builds.

This caused an issue with kokkos/develop so it will eventually be needed: https://github.com/kokkos/kokkos/issues/4611


CMakeLists.txt
```diff
@@ -7,7 +7,7 @@
 
 # CMAKE File for Albany building against an installed Trilinos
 
-cmake_minimum_required(VERSION 2.8.12)
```
bartgol, line 10: I can't believe we were still accepting CMake 2.X...

dalg24: Me neither :)

bartgol (approved):

mperego:
@jewatkins can you hold this a bit? We had a linking issue on cori with gcc where we had to build Trilinos with cmake 3.18 and switch back to 3.14 for Albany

jewatkins:
> @jewatkins can you hold this a bit? We had a linking issue on cori with gcc where we had to build Trilinos with cmake 3.18 and switch back to 3.14 for Albany

Sure. If you want to hold off on working it, we could set the min to 3.14.

mperego:
Right. But let me first duble check if we still have that issues.

mperego:
@jewatkins I'm building on cori with intel compiler and for some reason, if I use cmake 3.18 it wants to link libgcc_s but it does not find it. However if I use cmake 3.14 it does not link that library. Any clue?

mperego:
I'm also using `export CRAYPE_LINK_TYPE=static`.

jewatkins:
I saw a lot of warnings about using glibc but it still built fine:
```
warning: Using 'dlopen' in statically linked applications requires at runtime the shared libraries from the glibc version used for linking
```
I think I always see those though and it still runs fine.

jewatkins:
Add this: `-D CMAKE_EXE_LINKER_FLAGS="-fPIC -ldl". I think that will fix it.

mperego:
> Add this: `-D CMAKE_EXE_LINKER_FLAGS="-fPIC -ldl"`. I think that will fix it.

To Trilinos or Albany configure? I did that in Albany but it didn't help (and it didn't recognize fPIC). 
I had this before `-D CMAKE_EXE_LINKER_FLAGS:STRING="-static -Wl,-zmuldefs -fPic -ldl"`.

jewatkins:
I have it in both trilinos and albany. Okay, that looks fine. I don't see anything else that would be related to libgc. I'm currently using `cmake/3.21.3` maybe try with that? I can send you my files to see what the differences are.

mperego:
Let's try that. And the modules as well. I'm also using `cmake/3.21.3`. Is albany linking `lgcc_s`?

jewatkins:
I pushed my files here:
https://github.com/sandialabs/Albany/tree/master/doc/LandIce/cori
I see -lgcc in the link list but not -lgcc_s. I do see -lgcc_s in the gcc build which I build using shared libs.

jewatkins:
Do you have this for trilinos?
```
  -D TPL_FIND_SHARED_LIBS:BOOL=OFF \
  -D Trilinos_LINK_SEARCH_START_STATIC:BOOL=ON \
```
Maybe that will avoid the shared lib.

bartgol:
> > @jewatkins I'm building on cori with intel compiler and for some reason, if I use cmake 3.18 it wants to link libgcc_s but it does not find it. However if I use cmake 3.14 it does not link that library. Any clue?
> 
> I think I've run into similar issues before due to dynamic linking. My current intel build on Cori uses `export CRAYPE_LINK_TYPE=static`. Let me try rebuilding with this change to see if my build works.

In the E3SM fork of Kokkos we ended up doing a couple of mods, and one was [this](https://github.com/E3SM-Project/kokkos/blob/2dbe8cecae5a91685eeae07fe6e88b542b4ff34e/cmake/kokkos_check_env.cmake#L8), which was precisely to avoid link warnings/errors on Cori.

jewatkins:
I think we can do the same with `-D Kokkos_ENABLE_LIBDL:BOOL=OFF \` in the trilinos config. I think it's still needed for the profiling tools though which is why I have it on.

I think it still works despite the warning. The shared libs exist. I don't think that was always the case though.

bartgol:
Yes, it's just a warning. E3SM needed the fix because of `-Werror` in the nightly testing build options... I was just pointing out we're not alone with that error on Cori. And yes, I think it can be ignored safely.

jewatkins:
Ah okay makes sense

mperego:
@jewatkins I recompiled everything using your modules (but my scripts) and it worked fine. Maybe was `cray-libsci` that was creating an issue? 
Not sure what happens with a gcc build.. but we probably not need to worry about that since intel works better on Cori. @ikalash would it be OK to update the cori nightly build using intel compiler so that we can merge this PR?

mperego:
Another cmake option that seems to be critical is
`-D Albany_BUILD_STATIC_EXE:BOOL=OFF \`
If it's ON, I get
`/usr/bin/ld: /opt/intel/compilers_and_libraries_2019.3.199/linux/mkl/lib/intel64/libmkl_intel_lp64.so: undefined reference to 'mkl_serv_load_vml'`

mperego:
Thanks!

PR #787, opened 2022-02-03
Remove the ParamName template argument from SharedParameter

kliegeois:
This PR removes the ParamName template argument from SharedParameter.
Instead of static data member, this branch relies on maps of accessors at the problem level to synchronize the value of shared parameters among field manager of the same data type.

The work of PR has been heavily influenced by @bartgol, thanks a lot!

The ctest tests passed on Blake and the branch is building warning free.


src/evaluators/utility/PHAL_SharedParameter.hpp
```diff
@@ -8,27 +8,38 @@
...
 #include "Sacado_ParameterAccessor.hpp"
+#include "Albany_Accessors.hpp"
 
 //IKT 6/3/2020 TODO: implement support for vector parameters, which is not available currently.
 
 namespace PHAL
 {
 
-template<typename EvalT, typename Traits, typename ParamNameEnum, ParamNameEnum ParamName>
+template<typename EvalT, typename Traits, typename ParamNameEnum>
```
bartgol, line 18: Can't we also remove the last template arg?

src/LandIce/problems/LandIce_Enthalpy.hpp
```diff
@@ -623,6 +623,7 @@ LandIce::Enthalpy::constructEvaluators (PHX::FieldManager<PHAL::AlbanyTraits>& f
       entity = Albany::StateStruct::NodalDistParameter;
       p = stateMgr.registerStateVariable(stateName, dl->node_scalar, elementBlockName, true, &entity, "");
       p->set<std::string>("Parameter Name", stateName);
+      p->set<Teuchos::RCP<Albany::AccessorsMap>>("Accessors", this->getAccessors());
```
bartgol, line 626: General comment: I would make the class/file/strings names more verbose: instead of `Accessors`, I would use `ParameterAccessors`, so it's clearer what kind of 'access' we are talking about.

src/utility/Albany_Accessors.hpp
```diff
@@ -0,0 +1,38 @@
...
+#include "Albany_Meta_Utils.hpp"
+#include "PHAL_AlbanyTraits.hpp"
+
+
+namespace Albany
+{
+
+template<typename ST>
+struct Accessor {
+  ST& getValue () { return value; }
```
bartgol, line 13: I'm surprised we don't need a const method as well...

Edit: of course, since you only store shared_ptr<Accessor<...>>, it's always a non-const object. See comments in SharedParameter though.

src/evaluators/utility/PHAL_SharedParameter.hpp
```diff
@@ -8,27 +8,38 @@
...
 
-  SharedParameter (const Teuchos::ParameterList& p, const Teuchos::RCP<Albany::Layouts>& dl)
+  SharedParameter (Teuchos::ParameterList& p, const Teuchos::RCP<Albany::Layouts>& dl)
   {
     param_name   = p.get<std::string>("Parameter Name");
     param_as_field = PHX::MDField<ScalarT,Dim>(param_name,dl->shared_param);
 
+    Teuchos::RCP<Albany::Accessors<EvalT>> accessors = p.get<Teuchos::RCP<Albany::AccessorsMap>>("Accessors")->at<EvalT>();
+    if (accessors->accessors.count(param_name)==0) {
+      Albany::Accessor<ScalarT> tmp0;
```
bartgol, line 38: I think you can remove this line, and just call make_shared below with no input argument, no?

src/evaluators/utility/PHAL_SharedParameter.hpp
```diff
@@ -8,27 +8,38 @@
...
   typedef ParamNameEnum             EnumType;
 
-  SharedParameter (const Teuchos::ParameterList& p, const Teuchos::RCP<Albany::Layouts>& dl)
+  SharedParameter (Teuchos::ParameterList& p, const Teuchos::RCP<Albany::Layouts>& dl)
   {
     param_name   = p.get<std::string>("Parameter Name");
     param_as_field = PHX::MDField<ScalarT,Dim>(param_name,dl->shared_param);
 
+    Teuchos::RCP<Albany::Accessors<EvalT>> accessors = p.get<Teuchos::RCP<Albany::AccessorsMap>>("Accessors")->at<EvalT>();
+    if (accessors->accessors.count(param_name)==0) {
```
bartgol, line 37: I think it shouldn't be SharedParameter creating the accessors...It should be the problem class. Then, you could actually already pass the right accessor (for the right ScalarT) to the ctor of SharedParam, rather than pass the whole map.

src/evaluators/utility/PHAL_SharedParameter.hpp
```diff
@@ -60,7 +71,7 @@ class SharedParameter : public PHX::EvaluatorWithBaseImpl<Traits>,
             this->registerSacadoParameter(param_name, paramLib);
             if (pvi.isParameter("Nominal Value")) {
               double nom_val = pvi.get<double>("Nominal Value");
-              value = nom_val;
+              accessor->getValue() = nom_val;
```
bartgol, line 74: If the problem created all the accessors, we would be able to remove this clunky initialization code from this evaluator, and stuff it in the problem class instead. I would add a method to AbstractProblem, like the following:

```
template<typename EvalT>
void addParameterAccessor (const std::string& name, const double default);
```
This method would do exactly what is done in SharedParameter (i.e., scan the "Paramater" section of the "Problem" list for a nominal value, and use the default if not found). Since the accessors are stored in the problem, it's conceptually more appropriate if the problem inits them.

src/LandIce/problems/LandIce_Enthalpy.hpp
bartgol: Even better `ScalarParameterAccessor`, to clarify that this has nothing to do with distributed parameters.

src/evaluators/utility/PHAL_Field_Source_Def.hpp
```diff
@@ -95,11 +95,12 @@ namespace PHAL
                     Teuchos::RCP<Teuchos::ParameterList> p = Teuchos::rcp(new Teuchos::ParameterList(Albany::strint("Gaussian: Amplitude", num)));
                     p->set<Teuchos::RCP<ParamLib>>("Parameter Library", paramLib);
                     p->set<std::string>("Parameter Name", param_name);
+                    p->set<Teuchos::RCP<Albany::AccessorsMap>>("Accessors", scalarParam_list.get<Teuchos::RCP<Albany::AccessorsMap>>("Accessors"));
```
bartgol, line 98: Another point in support of not creating accessors in SharedParam: it seems silly to pass around the whole map when we know already which accessor will be needed. If only they were already created outside (by the problem), we'd be able to pass around only the one that is needed.

src/evaluators/utility/PHAL_SharedParameter.hpp
```diff
@@ -8,27 +8,38 @@
...
 {
 
-template<typename EvalT, typename Traits, typename ParamNameEnum, ParamNameEnum ParamName>
+template<typename EvalT, typename Traits, typename ParamNameEnum>
 class SharedParameter : public PHX::EvaluatorWithBaseImpl<Traits>,
                         public PHX::EvaluatorDerived<EvalT, Traits>,
                         public Sacado::ParameterAccessor<EvalT, SPL_Traits>
 {
+private:
+  std::shared_ptr<Albany::Accessor<typename EvalT::ScalarT>> accessor;
```
bartgol, line 24: If the accessors are created in the problem, this could be a `shared_ptr<const ...>` instead, ensuring SharedParam cannot change the parameter values (you never know with copy+paste in future refactors/changes...).

src/utility/Albany_Meta_Utils.hpp
```diff
@@ -0,0 +1,226 @@
+#ifndef ALBANY_META_UTILS_HPP
```
bartgol, line 1: Since this file was taken verbatim from an open source library, we need to put a header stating it. Something like

```
// This file is copied from "ekat_meta_utils.hpp", a file in the EKAT library (https://github.com/E3SM-Project/EKAT).
```

src/utility/Albany_Accessors.hpp
```diff
@@ -0,0 +1,38 @@
...
+  ST value;
+};
+
+template<typename EvalT>
+struct Accessors {
+  std::map<std::string,std::shared_ptr<Accessor<typename EvalT::ScalarT>>> accessors;
+};
+
+using AccessorsKeyList =
```
bartgol, line 22: You could name this `AlbanyEvalTypes`, which probably conveys more meaning.

src/utility/Albany_Accessors.hpp
```diff
@@ -0,0 +1,38 @@
...
+  Albany::TypeList<PHAL::AlbanyTraits::Residual,
+                  PHAL::AlbanyTraits::Jacobian,
+                  PHAL::AlbanyTraits::Tangent,
+                  PHAL::AlbanyTraits::DistParamDeriv,
+                  PHAL::AlbanyTraits::HessianVec>;
+using AccessorsValueList =
```
bartgol, line 28: Since you are at it, you could use the `ApplyTemplate` meta util you took from EKAT, and do

```
template<typename EvalT>
using AccessorsRCP = Teuchos::RCP<Albany::Accessors<EvalT>>;

using AccessorsValueList = typename ApplyTemplate<AccessorsRCP,AccessorsKeyList>::type;
```

bartgol (changes requested):
Big undertaking, and big cleanup. The patch I did to fix the inconsistency between field managers was ugly, heavy, and not scalable. Thanks for cleaning it up.

I have some comments, suggestions, and requests. See inline comments.

kliegeois:
@bartgol as discussed, I have updated the PR following all your comments except the fact to not build `Accessor`s (not `Accessors`) into `SharedParameter` as not all the shared parameters are created at the problem level yet.
This could be updated in a following PR though.

bartgol (approved):
Looks good. What you said about SharedParameters being created outside the problem forces us to keep the Accessors creation inside the evaluator. Maybe we can revisit down the road.

PR #789, opened 2022-02-07
Minor string utils clean up and add-on

bartgol:
Some minor mods:

* Renamed `string.hpp` to `Albany_StringUtils.hpp` (the former was a way too common name, and way to close to std lib `string`).
* Moves some string utils out of `Albany_Utils.*pp` and into `Albany_StringUtils.*pp`.
* Add a unit test file for string utils.
* Add a function to parse a string representing nested lists into a `Teuchos::ParameterList`.

I'm planning to use the latter in a few places, when parsing strings representing block structures.


kliegeois (approved):
Looks good to me; that's a lot of change!

PR #791, opened 2022-02-10
Removed template parameter from STK field containers.

bartgol:
It was never used to deduce compile-time information, so we can simply store it as a class member. Each of the container classes was instantiated 3 times (due to 3 different values of DiscType), so this should cut a couple of instantiations per class.


kliegeois (approved):
Looks good to me.
Just out of curiosity; is the impact of the PR on the size of the lib and/or the compilation time significative?

bartgol:
I don't think you'll notice it, but it seemed a waste anyways. Besides, simplifying disc stuff is important for working on block discretizations: removes unnecessary stuff, lowering the amount of code to maintain/fix/adjust.

PR #794, opened 2022-02-14
Removed AuxData/EigenData related structures

bartgol:
These include:

- EigenData, Eigenvector, and AuxData evalautors
- EigendataInfoStruct
- Eigensolve value for SolutionMethod enum

They were never used, a vestige of capabilities that have long been discontinued in Albany. I meant to clean these files/classes up long time ago, during the Thyra refactor, but I think we were not really sure if we still needed them. After removing and testing, I can confirm they are not used anymore. So, for the sake of future development, is probably best to prune them, to keep the amount of code to maintain/adapt smaller.

While removing this, I grepped for `eigen` in the repo, and found some matches in `Albany_StatelessObserverImpl.hpp`, in the comments at the top. Reading the comments, I started wondering whether stateless observers are still needed? @mperego @ikalash do you know? Can we remove them, and have one single observer class?


mperego (approved):
Thanks @bartgol . I think it makes sense to remove EigenData and AuxData stuff.

mperego:
@bartgol I think the stateless observers are used to print transient sensitivities.

PR #798, opened 2022-03-18
Replace "-l-L" with "-L" in albany export file

jewatkins:
This fixes the issue where `-l-L` shows up in the albany export file. It was safer just to replace it with `-L` since the extra `-l` might be needed for some other configuration I'm not aware of.

I tested it by adding `-D Trilinos_EXTRA_LINK_FLAGS:STRING="-L/some/lib" ` in a trilinos build. It works fine.


mperego (approved):
Thanks!

PR #800, opened 2022-03-25
Use Cn instead of C2 basis that will be deprecated + cleaning

mperego:
Quadratic FE basis (C2) will be deprecated in Intrepid2, in favor of basis or variable order (Cn).
Here we use Cn, with n=2, instead of C2.


ikalash:
When is the deprecation happening, @mperego ?  I'll need to make similar changes in LCM.

mperego:
> When is the deprecation happening, @mperego ? I'll need to make similar changes in LCM.

Ideally soon, altough we'll ait a few months before removing them. I'm trying to check whether this could be harmful for applications. I sent you and Alejandro and email for LCM.

mperego:
Not sure any longer we are going to deprecate C2 functions as they are used for quadratic meshes (whose ordering of nodes mach that of C2 but not that of Cn). Anyway taking this opportunity for a little cleanup. Removed spectral elements and quadratic meshes, that wer not tested anyay.

mperego:
> I'm curious: why does Intrepid2 still keeps the specialization for first order then?

Backword compatibility and performance. Also Cn, with n=1 is the same as C1. That's not true for C2 (because of different ordering of DoFs), which creates troubles.

PR #801, opened 2022-04-19
PyAlbany based Randomized SVD and HODLR compression codes

thartland:
Here I am adding methods for compressing matrix-free operators in PyAlbany into Randomized SVD and hierarchical off-diagonal low-rank (HODLR) matrix formats, via methods under RandomizedCompression. Tests are included in which the errorBounds are guaranteed to be satisfied with probability 1 - 6 * 10^(-10).


thartland:
Sure thing @mperego . When writing up the notes (so that they are more consistent with the code), I saw a way to generate symmetric low-rank approximations via a doublePass algorithm so have included that as well. Let me know if there anything is unclear.
[writeup.pdf](https://github.com/sandialabs/Albany/files/8526630/writeup.pdf)

kliegeois:
Thanks @thartland , I am starting the review process.

PyAlbany/python/Utils.py
```diff
@@ -25,6 +25,31 @@ def inner(distributedVector1, distributedVector2, comm):
...
+def innerMVectorMat(distributedMVector, array):
+    """@brief computes C = A B, where A and C are distributed n x r and B is r x r and available to each process"""
+    r1, nloc = distributedMVector.shape
+    r2       = array.shape[1]
+    dtype    = distributedMVector.dtype 
+    C = np.zeros((r2, nloc), dtype=dtype)
+    for k in range(r1):
+        for i in range(r2):
+            C[i, :] += array[k, i] * distributedMVector[k, :]
+    return C
```
kliegeois, line 50: For this function, if I get it correctly, both `C` and `A` are multivectors. However, here `C` is not a distributed Tpetra multivector, it is a numpy array. I think that the comment should be updated or that the function should create a new Tpetra multivector using the map of the existing one and return that Tpetra multivector.

Regardless of the chosen solution, we should update the comment to specify that r1 and r2 can be different, for now the comment only uses r.

PyAlbany/python/RandomizedCompression.py
```diff
@@ -0,0 +1,342 @@
...
+  input: Op, operator which contains a Map member data and a 
+                  dot method which takes a Tpetra MultiVector
+                             and returns a Tpetra MultiVector 
+          k, the rank of the output operator U \Lambda U^T
+  output: \Lambda, U
+          \Lambda, eigenvalues,  a nondistributed numpy array with k entries
+          U,       eigenvectors, a distributed Tpetra MultiVector with k columns
+"""
+def singlePass(Op, k):
+    comm       = Teuchos.DefaultComm.getComm()
```
kliegeois, line 52: This comment applies to every functions of this file that uses `Teuchos.DefaultComm.getComm()`.

While it is true that currently (in the master branch) we always have the fact that `comm == Teuchos.DefaultComm.getComm()`, it will not be true anymore after a PR that I will submit soonish.
It is best practice here to do the following:
```
def singlePass(Op, k, comm = Teuchos.DefaultComm.getComm()):
    rank       = comm.getRank()
```
This does not change how you call the function from your test but it allows the user to use a non-default communicator.

PyAlbany/python/RandomizedCompression.py
```diff
@@ -0,0 +1,342 @@
...
+    N          = Op.Map.getGlobalNumElements()
+    nMaxOrthog = min(50, k)
+
+    omega = Tpetra.MultiVector(Op.Map, k, dtype="d")
+    q     = Tpetra.MultiVector(Op.Map, k, dtype="d")
+    u     = Tpetra.MultiVector(Op.Map, k, dtype="d")
+    
+    for i in range(k):
+        omega[i, :] = np.random.randn(nElems)
+    y = Op.dot(omega)
```
kliegeois, line 65: I like the notion of Operator, I did something similar in ExtremeEvent.py for `class HessianOperator(slinalg.LinearOperator)`.

I think that it would be best to gather the common operator class into a same python file to avoid having the same (or similar) thing in different files.
I am not sure if you have to do this in this PR though.
I can do it later.

PyAlbany/python/RandomizedCompression.py
```diff
@@ -0,0 +1,342 @@
...
+          \Lambda, eigenvalues,  a nondistributed numpy array with k entries
+          U,       eigenvectors, a distributed Tpetra MultiVector with k columns
+"""
+def singlePass(Op, k):
+    comm       = Teuchos.DefaultComm.getComm()
+    rank       = comm.getRank()
+    nprocs     = comm.getSize()
+    nElems     = Op.Map.getNodeNumElements()
+    N          = Op.Map.getGlobalNumElements()
+    nMaxOrthog = min(50, k)
```
kliegeois, line 57: Can you precise in your comment about `k` that it cannot be above 50?

PyAlbany/python/RandomizedCompression.py
```diff
@@ -0,0 +1,342 @@
...
+    v[:,:]     = utils.innerMVectorMat(qz, vhat)
+    return u, sig, v
+
+"""
+  doublePassSymmetric: randomized singular value decomposition of an operator
+                       here operator is assumed symmetric
+  input: Op, operator which contains a Map member data and a 
+                  dot method which takes a Tpetra MultiVector
+                             and returns a Tpetra MultiVector 
+          k, the rank of the output operator U \Lambda U^T
```
kliegeois, line 148: Same comment as above

kliegeois (changes requested):
This looks pretty good to me.
Thanks for your work!

However, I have couple of small comments that would require some small modifications.
All the tests passed on my workstation.

PyAlbany/python/RandomizedCompression.py
```diff
@@ -0,0 +1,342 @@
...
+  output: \Lambda, U
+          \Lambda, eigenvalues,  a nondistributed numpy array with k entries
+          U,       eigenvectors, a distributed Tpetra MultiVector with k columns
+"""
+def singlePass(Op, k):
+    comm       = Teuchos.DefaultComm.getComm()
+    rank       = comm.getRank()
+    nprocs     = comm.getSize()
+    nElems     = Op.Map.getNodeNumElements()
```
mperego, line 55: `getNodeNumElements() `is deprecated. Please use `getLocalNumElements()` instead, here and in the rest of the PR. It's just a renaming.

kliegeois: Oups, I forgot that even though my compiler complained, thanks for catching that!

PyAlbany/python/RandomizedCompression.py
thartland: I recall an issue when using `wpa.orthogTpMVecs`, when using more than 50 vectors at a time, that is orthogonalizing more than 50 vectors at once. Here it is reduced so that we attempt to orthogonalize the k columns using no more than 50 columns. I should look back into the orthogonalization at some point though.

PyAlbany/python/RandomizedCompression.py
thartland: Good idea to allow for this code to be more flexible and not be constrained to the Techos DefaultComm, the change has been made.

PyAlbany/python/RandomizedCompression.py
thartland: I see what you mean about having a common way to define this HessianOperator class. One thing is that I wanted the class to have a method to take in a `MultiVector` and output a `MultiVector`, but I suppose this could be accomodated by writing a `matmat` method. Let me know if you would like me to make these changes.

PyAlbany/python/RandomizedCompression.py
thartland: This change has now been made.

PyAlbany/python/RandomizedCompression.py
```diff
@@ -0,0 +1,337 @@
...
+   
+    args = np.argsort(lam)
+
+    lam    = lam[args[::-1]]
+    utilde = utilde[:, args[::-1]]
+
+    u      = utils.innerMVectorMat(q, utilde)
+    return lam, u
+
+def doublePass(Op, k, comm = Teuchos.DefaultComm.getComm(), symmetric=True):
```
mperego, line 89: Sorry, I just noticed this. I think it's better not to provide a default value for "symmetric" (which might require changing the order of the parameters). However, if you want to give a default value, it should be `false`, so that it will work for all matrices even if not optimally.

thartland: This default has been changed.

kliegeois:
Thanks @thartland !
It looks good to me.

mperego:
Thanks a lot @thartland!

PR #802, opened 2022-04-27
Update to the extreme event probability computation

kliegeois:
This PR includes several modifications:
- Of Albany:
1. Add `PowerScalarResponseFunction`: a response that takes the value of another response, substract a targeted value, and raises the difference to a user defined power.
2. Add `LinearCombinationParameterBase`: an evaluator that computes a field define as a linear combination of depend fields.
2. Add `LogGaussianDistributedParameterBase`: and evaluator that computes a log normal distributed field using a mean value, and a Gaussian field (defined as a `LinearCombinationParameterBase`).
3. Modify `LandIce_StokesFOBase.hpp` to have the option to represent the basal friction as log normal random field (using both `LinearCombinationParameterBase` and `LogGaussianDistributedParameterBase`).
- Of PyAlbany:
1. Add strategies to compute the KL expansion terms and eigenvalues: using Stokhos with `PyAlbany_Stokhos`, using a Nystrom, Collocation, or Galerkin method (in Python in `ExtremeEvent.py`)
2. Couples of small updates.

The tests passed on my workstation, on blake, and build warning free.

Sorry for the long PR.


src/evaluators/utility/PHAL_LogGaussianDistributedParameter.hpp
```diff
@@ -0,0 +1,324 @@
...
+// * Specializations
+// **************************************************************
+// **************************************************************
+
+
+// **************************************************************
+// Residual
+// **************************************************************
+template<typename Traits>
+class LogGaussianDistributedParameter<PHAL::AlbanyTraits::Residual,Traits>
```
bartgol, line 106: Like for lin comb: all specializations seem to do the same thing. Can we do everything in the base class?

src/responses/Albany_PowerScalarResponseFunction.hpp
```diff
@@ -0,0 +1,168 @@
...
+#ifndef ALBANY_POWER_SCALAR_RESPONSE_FUNCTION_HPP
+#define ALBANY_POWER_SCALAR_RESPONSE_FUNCTION_HPP
+
+#include "Albany_SamplingBasedScalarResponseFunction.hpp"
+#include "Teuchos_Array.hpp"
+
+namespace Albany {
+
+  /*!
+   * \brief A response function that ...
```
bartgol, line 16: can you add some comment here?

bartgol (commented):
I only reviewed the stuff in `src`. I did not look at the `PyAlbany` folder. I feel like I miss knowledge to review that.

kliegeois:
Thanks @bartgol ! I will work on this!

src/LandIce/problems/LandIce_StokesFOBase.hpp
```diff
@@ -1065,65 +1065,71 @@ constructVelocityEvaluators (PHX::FieldManager<PHAL::AlbanyTraits>& fm0,
...
       p->set<std::string>("Parameter Name", param_name);
       p->set<std::size_t>("Number of modes", numModes);
 
       for (std::size_t i = 0; i < numModes; ++i) {
         Teuchos::RCP<Teuchos::ParameterList> pi = Teuchos::rcp(new Teuchos::ParameterList("Mode"));
-        pi->set<std::string>("Coefficient Name", lcparams_i.sublist(util::strint("Mode",i)).get<std::string>("Coefficient Name"));
-        pi->set<std::string>("Mode Name", lcparams_i.sublist(util::strint("Mode",i)).get<std::string>("Mode Name"));
+        pi->set<std::string>("Coefficient Name", coeff_names[i]);
+        pi->set<std::string>("Mode Name", mode_names[i]);
         p->sublist(util::strint("Mode",i)) = *pi;
```
bartgol, line 1125: I still think it would be easier to set the array of modes and the array of coeffs names in the parameter list `p`, rather than set N sublists with two strings each...

bartgol (approved):
I still have one commend on the LCParam list structure, but might be personal taste.

kliegeois:
Thanks @bartgol ! I can change it too; I thought that it was less important as it is not at the user level but I can change that.

bartgol (approved):

kliegeois:
Thanks! to both of you, I am merging it.

Issue #768, opened 2021-12-10
Nightly tests failing due to "Random Parameters"


jewatkins:
@kliegeois @mperego @bartgol 
It looks like the tests are failing due to a recent PR https://github.com/sandialabs/Albany/pull/766.

Was the intention to make "Random Parameters" a requirement in the LandIce StokesFO problem? If not, then the sublist needs to be checked for existence otherwise it will get created.

jewatkins:
Failure example:
```
p=1: *** Caught standard std::exception of type 'Teuchos::Exceptions::InvalidParameterName' :

 Error!  The parameter "Number Of Parameters" does not exist
 in the parameter (sub)list "Albany Parameters->Problem->Random Parameters".
 
 The current parameters set in (sub)list "Albany Parameters->Problem->Random Parameters" are:
 
   {
   }
 
 
 Throw number = 1
```

kliegeois:
Thanks @mperego !

PR #803, opened 2022-04-27
set cmake new policy CMP0057 to avoid Kokkos error

mperego:
See https://github.com/trilinos/Trilinos/issues/10467.


ikalash (approved):
I tested this on my workstation camobap, and the problem is fixed.  Thanks for fixing it so fast!  @lxmota , I'll propagate these changes into LCM so we don't have build errors there in the nightlies.

jewatkins:
@mperego Do you want me to close #786? I'd prefer increasing the minimum requirement to fix the issue rather than turning on policies but maybe there's still issues with that PR.

mperego:
I'm using cmake 3.21.1, so it seems that cmake version alone is not going to fix this. I think we should merge #768 anyway.

jewatkins:
Did you set the minimum to 3.21.1? That's what sets the new cmake policies. What you have here is an ability to toggle the old policy.

mperego:
3.21 gives the warning if the policy is not set. It does not set it to new by default. See e.g.
https://cmake.org/cmake/help/latest/policy/CMP0057.html
I think if you merge #768 in addition to this PR we are good.

jewatkins:
Setting `cmake_minimum_required(VERSION 3.17.0)` will update the policy (see https://cmake.org/cmake/help/latest/command/cmake_minimum_required.html). This is similar to what Trilinos does (see https://github.com/trilinos/Trilinos/blob/develop/CMakeLists.txt). I tested this back in Feb. and it worked. Did you change `cmake_minimum_required(VERSION 3.21)` and it gave a warning?

PR #807, opened 2022-05-05
Fix #806

### LINKED ISSUE #806: PyAlbany compilation error
It looks like PyAlbany failed to compile last night due to a missing include:

```
/nightlyCDash/repos/Albany/PyAlbany/src/PyAlbany_Stokhos.hpp:10:10: fatal error: Stokhos_KL_ExponentialRandomField.hpp: No such file or directory
```
https://sems-cdash-son.sandia.gov/cdash/viewBuildError.php?buildid=33107

My guess is the file included either got removed or renamed in Trilinos.  Do we really need the include?  Would be great if you could have a look @kliegeois .

kliegeois:
This should fix #806

### DIFF
--- a/PyAlbany/src/PyAlbany_Stokhos.cpp
+++ b/PyAlbany/src/PyAlbany_Stokhos.cpp
@@ -3,6 +3,7 @@
 using namespace PyAlbany;
 
 KLExpansion::KLExpansion(int _ndim) : ndim(_ndim) {
+#ifdef ALBANY_STOKHOS    
     domain_upper = Teuchos::Array<double>(ndim, 1.0);
     domain_lower = Teuchos::Array<double>(ndim, 0.0);
     correlation_lengths = Teuchos::Array<double>(ndim, 1.0);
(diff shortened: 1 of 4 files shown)


PyAlbany/src/PyAlbany_Stokhos.cpp
```diff
@@ -3,6 +3,7 @@
 using namespace PyAlbany;
 
 KLExpansion::KLExpansion(int _ndim) : ndim(_ndim) {
+#ifdef ALBANY_STOKHOS
```
bartgol, line 6: All these ifdefs are pointless, no?

bartgol (approved):
Looks good. I have minor comment, but nothing serious.

mperego (commented):
Thanks @kliegeois. 
If  Stokhos is not enabled, then the class KLExpansion would not work as expected. Isn't it better to avoid compiling the files Albany_Stokhos.cpp and Albany_Sthokos.hpp?

Is this code used at all? If it's used in a test we should disable the test as well when ALBANY_STOKHOS is false.

kliegeois:
There is no tests that use those. The main issue and the raison why I did it with so many ifdef is swig. I am 90% sure that only one per file would not compile PyAlbany correctly; I am not 100% sure though.

kliegeois:
I can try to reduce the number of ifdef though.

mperego:
Thanks. We really should have a test for it. No need to do add the test in this PR, though.

ikalash:
I merged this in.  Will check tomorrow if the failed nightly is fixed.

kliegeois:
It should fix the issue but it is not yet working perfectly. It will build and run the tests correctly.

mperego:
@ikalash we were still discussing about this PR. That's why Kim added the WIP label. Not a big deal, but I think it's in general better not to merge PR's by others without checking first.

ikalash:
Oh shoot...  I missed the WIP label.  Sorry about this - somehow the commenting made me erroneously believe it was ready to be merged.  I can revert the merge if you'd like.

bartgol:
I think we can keep it in now. If Kim has additional commits, he can submit a separate PR, I guess.

mperego:
No worries. No need to revert, thanks.

PR #808, opened 2022-05-19
Laplacian regularization

mperego:
This PR contains (each point is a separate commit):

- a fix to the  `L2ProjectedBoundaryLaplacianResidual` which was buggy
- an update of the Humboldt mesh and relative tests. The new mesh is a bit smaller and from a more recent dataset
- a new response, `Quadratic Linear Operator Based`, that computes `p A' inv(D) A p`, where p is a field and `A` and `D` are matrices read from file, `D` being diagonal
- added/modified tests to use the response for computing the Laplacian squared regularization, where `A` is a stiffening matrix (including a mass and robin boundary condition term) and `D` is a lumped mass matrix. The response results are equivalent to those computed with the` Boundary Squared L2 Norm` response together with the `L2ProjectedBoundaryLaplacianResidual` evaluator


src/responses/Albany_QuadraticLinearOperatorBasedResponseFunction.cpp
```diff
@@ -0,0 +1,343 @@
...
+    loadLinearOperator();
+
+  Teuchos::RCP<const Thyra_Vector> field = app_->getDistributedParameterLibrary()->get(field_name_)->vector();
+
+  // A p
+  A_->apply(Thyra::EOpTransp::NOTRANS, *field, vec1_.ptr(), 1.0, 0.0);
+
+  // coeff inv(D) A p
+  vec2_->assign(0.0);
+  Thyra::ele_wise_divide( coeff_, *vec1_, *D_, vec2_.ptr() );
```
bartgol, line 93: Thyra::VectorBase has a method to compute a weighted 2 norm. You could just do (here and in the other evaluate methods)
```
g->assign (std::pow(vec1->norm_2(*D_),2));
```
and do everything in one loop.

bartgol: NVM, the Tpetra adapter impl does precisely what you are doing.

tests/small/LandIce/FO_GIS/input_fo_humboldt_frosch_fluxdiv.yaml
```diff
@@ -105,7 +115,7 @@ ANONYMOUS:
     Exodus Output File Name: output/humboldt.exo
     Element Shape: Wedge
     Use Serial Mesh: false
-    Columnwise Ordering: true
+    Columnwise Ordering: false
```
jewatkins, line 118: What's the reasoning behind this change?

bartgol (approved):
Looks good.

jewatkins (approved):
Looks good.

tests/small/LandIce/FO_GIS/input_fo_humboldt_frosch_fluxdiv.yaml
mperego: The new response works only for contiguous maps (it would error out otherwise). When the columnwise ordering is true, the basal map is not contiguous.

src/responses/Albany_QuadraticLinearOperatorBasedResponseFunction.cpp
mperego: here I need to divide by `D`, not multiply by `D`.

mperego:
Thanks @bartgol and @jewatkins for the review!

src/responses/Albany_QuadraticLinearOperatorBasedResponseFunction.cpp
bartgol: Ah, right.

PR #810, opened 2022-05-25
Fix evaluate response

kliegeois:
This PR makes sure the scalar parameter values are updated before calling evaluateResponse.

All the tests passed on my workstation except:

```
         44 - MPNIQuad2D_Epetra (Failed)
         45 - MPNIQuad2D_Tpetra (Failed)
```
but they failed on the master branch too.

The branch compiles warning free.

### DIFF
--- a/src/Albany_Application.cpp
+++ b/src/Albany_Application.cpp
@@ -1344,13 +1344,6 @@ Application::computeGlobalResidualImpl(
   // Scatter distributed parameters
   distParamLib->scatter();
 
-  // Set parameters
-  for (int i = 0; i < p.size(); i++) {
-    for (unsigned int j = 0; j < p[i].size(); j++) {
-      p[i][j].family->setRealValueForAllTypes(p[i][j].baseValue);
-    }
-  }
-
   // Zero out overlapped residual
   overlapped_f->assign(0.0);
   f->assign(0.0);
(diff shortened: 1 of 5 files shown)


src/Albany_Application.cpp
```diff
@@ -1567,13 +1560,6 @@ Application::computeGlobalJacobianImpl(
   // Scatter distributed parameters
   distParamLib->scatter();
 
-  // Set parameters
```
mperego, line 1570: Wouldn't it be safer to zero out the derivatives here as well?

src/Albany_Application.cpp
```diff
@@ -1344,13 +1344,6 @@ Application::computeGlobalResidualImpl(
...
-  for (int i = 0; i < p.size(); i++) {
-    for (unsigned int j = 0; j < p[i].size(); j++) {
-      p[i][j].family->setRealValueForAllTypes(p[i][j].baseValue);
-    }
-  }
-
   // Zero out overlapped residual
```
mperego, line 1347: Wouldn't it be safer to zero out the derivatives  here as well?

src/Albany_Application.cpp
```diff
@@ -2110,13 +2098,6 @@ Application::applyGlobalDistParamDerivImpl(
   // Scatter distributed parameters
   distParamLib->scatter();
 
-  // Set parameters
-  for (int i = 0; i < p.size(); i++) {
-    for (unsigned int j = 0; j < p[i].size(); j++) {
-      p[i][j].family->setRealValueForAllTypes(p[i][j].baseValue);
-    }
-  }
-
```
mperego, line 2119: Wouldn't it be safer to zero out the derivatives here as well?

mperego (commented):
Thanks Kim. I left a couple of comments.

mperego (approved):
After talking to @kliegeois I realized that the current implementation is safe. So I'm approving this.

Issue #814, opened 2022-06-25
Failing nightlies due to STK changes


ikalash:
Looks like we're getting failed Albany builds due to STK: https://my.cdash.org/viewBuildError.php?buildid=2181971 .  

```
[Albany/src/disc/stk/Albany_IossSTKMeshStruct.cpp:106](https://github.com/gahansen/Albany/blob/master/Albany/src/disc/stk/Albany_IossSTKMeshStruct.cpp#L106):25: error: 'class stk::io::StkMeshIoBroker' has no member named 'meta_data_rcp'; did you mean 'meta_data_ptr'?
```
Is this due to changes in STK / removal of deprecations, which were discussed by email a few weeks ago?  Tagging @alanw0 .

alanw0:
@ikalash Hi Irina! Yes this method was deprecated. There is now a similar method called 'meta_data_ptr' which you can use instead, however it returns std::shared_ptr instead of Teuchos::RCP.
In stk our procedure is to delete deprecated methods after they have been marked deprecated for approx 6 weeks.

ikalash:
It looks like there is still a problem in one of the builds:

https://sems-cdash-son.sandia.gov/cdash/viewBuildError.php?buildid=35604

``
/.../repos/Albany/src/disc/stk/Albany_STKDiscretization.cpp(2312): error: no instance of overloaded function "stk::io::StkMeshIoBroker::set_bulk_data" matches the argument list
``

mperego:
@ikalash, it seems it's building an older version of Albany (c08885d7cd76cdbbb0016d62bfd411df3e81c2a1).

PR #815, opened 2022-06-28
Comply with changes in STK

mperego:
Fixes issue #814. 
Obtain `BulkData` class using `MeshBuilder` class instead of using `BulkData` constructor which is deprecated.

### DIFF
--- a/src/disc/stk/Albany_SideSetSTKMeshStruct.cpp
+++ b/src/disc/stk/Albany_SideSetSTKMeshStruct.cpp
@@ -8,12 +8,15 @@
 
 #include "Albany_SideSetSTKMeshStruct.hpp"
 
+#include "Teuchos_RCPStdSharedPtrConversions.hpp"
+
 #include <Shards_BasicTopologies.hpp>
 
 #include <stk_mesh/base/Entity.hpp>
 #include <stk_mesh/base/GetEntities.hpp>
 #include <stk_mesh/base/GetBuckets.hpp>
 #include <stk_mesh/base/Selector.hpp>
+#include <stk_mesh/base/MeshBuilder.hpp>
 
 #ifdef ALBANY_SEACAS
 #include <stk_io/IossBridge.hpp>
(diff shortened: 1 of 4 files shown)


alanw0 (commented):
These changes look correct to me.
Minor note - the 'add_fmwk_data' property is false by default so you don't need to set that, although it doesn't hurt to be explicit.

mperego:
> These changes look correct to me. Minor note - the 'add_fmwk_data' property is false by default so you don't need to set that, although it doesn't hurt to be explicit.

Thanks @alanw0 !

src/disc/stk/Albany_SideSetSTKMeshStruct.cpp
```diff
@@ -102,9 +105,12 @@ SideSetSTKMeshStruct::SideSetSTKMeshStruct (const MeshSpecsStruct& inputMeshSpec
                                                                 ebn, ebNameToIndex, this->interleavedOrdering));
 
   const Teuchos::MpiComm<int>* mpiComm = dynamic_cast<const Teuchos::MpiComm<int>* > (commT.get());
-  bulkData = Teuchos::rcp(new stk::mesh::BulkData(*metaData, *mpiComm->getRawMpiComm(),
-                                                   stk::mesh::BulkData::NO_AUTO_AURA,
-                                                   false, NULL, worksetSize));
+  stk::mesh::MeshBuilder meshBuilder = stk::mesh::MeshBuilder(*mpiComm->getRawMpiComm());
```
bartgol, line 108: I forgot why we don't let GenericSTKMeshStruct create the bulk data in this class...

bartgol (approved):

mperego:
thanks @bartgol !

src/disc/stk/Albany_SideSetSTKMeshStruct.cpp
mperego: Don't ask me.. :D

PR #820, opened 2022-07-11
Remove the need to use try: from PyAlbany import Utils

kliegeois:
@ikalash this PR removes the need to use: 

```
try: 
    from PyAlbany import Distributions as dist
    from PyAlbany import Utils
```    
for the ctest tests.

Instead, during the compilation of PyAlbany, a folder named `${BuilDir}/PyAlbany/PyAlbany` is created which includes all the python files of `PyAlbany/python` and the PyAlbany library generated with pybind11. This folder is then added to the PYTHONPATH of the ctest tests.

### DIFF
--- a/PyAlbany/src/CMakeLists.txt
+++ b/PyAlbany/src/CMakeLists.txt
@@ -137,6 +137,12 @@ FetchContent_Declare(
 
 FetchContent_MakeAvailable(pybind11)
 
+IF(ALBANY_PYTHON_TESTS)
+  MESSAGE("-- ALBANY_PYTHON_TESTS is TRUE.")
+ELSE()
+  MESSAGE("-- ALBANY_PYTHON_TESTS is FALSE.")
+ENDIF()
+
 # Retrieve the Mpi4Py version
 EXECUTE_PROCESS(COMMAND
   ${PYTHON_EXECUTABLE} -c "import mpi4py; print(mpi4py.__version__)"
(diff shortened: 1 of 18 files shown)


ikalash:
I tried this PR and am getting the same errors as before on blake in the warning-free build.  Perhaps you could test it yourself?  The modules file to use is this one: https://github.com/sandialabs/Albany/blob/master/doc/dashboards/blake.sandia.gov/blake_gcc_modules_submit.sh .

PyAlbany/src/CMakeLists.txt
```diff
@@ -137,6 +137,12 @@ FetchContent_Declare(
 
 FetchContent_MakeAvailable(pybind11)
 
+IF(ALBANY_PYTHON_TESTS)
+  MESSAGE("-- ALBANY_PYTHON_TESTS is TRUE.")
```
mperego, line 141: I think for other tests we use enabled or not enabled instead of True or False, e.g.
`-- Python Tests           NOT Enabled`

mperego (approved):

PyAlbany/src/CMakeLists.txt
kliegeois: Do you mean that it would be better to have the `MESSAGE` to use enabled or not enabled to be consistent with the rest?

mperego: yes, thanks.

kliegeois:
@ikalash I updated the PR based on our discussion.

For the warning free build, you can disable exomerge for now so that the tests pass in the nightly builds (the extreme event tests are now optional and are enabled if exomerge3.py (or exomerge2.py if python 2 is used) is found). I can revisit those extreme event tests later but I think that it is good to already enable the warning free nightly build for PyAlbany as it was never warning free before due to PyTrilinos.

mperego (approved):
Looks good. Thanks!

PR #821, opened 2022-07-13
Make Albany build with a Epetra free Trilinos build

mperego:
Trilinos can now be built Epetra free (no Epetra, AztecOO, Amesos, Ifpack, ML, etc.).
After the minor changes in this PR, Albany can build against an Epetra-free Trilinos install. 
Unfortunately Teko still has a required dependency on Epetra, so we had to disable a few analysis test that use Teko (for computing Hessian based dot-product).

@ikalash,  when you get a chance it would be good to make one of the nightly tests Epetra free.


bartgol (approved):
Does this mean that in order to use Teko we still need to build the Epetra stack?

mperego:
> Does this mean that in order to use Teko we still need to build the Epetra stack?

Yes :-(. Hopefully some of the Trilinos developers will make Teko optionally dependent on Epetra... or we will.

PR #830, opened 2022-09-06
Fix creation of export_albany.in

bartgol:
This creates the `export_albany.in` file without the namespace cmake notation for linked libraries/targets, but using the actual link line needed.


bartgol:
Still not 100% working: albany libs are linked as `../src/libalbanyLib.so`. I need to modify that to point to the install dir.

bartgol:
@mperego I tried to compare with master, but master produces the `blah::all_libs` entries (of course), so I can't be sure this is fine. But the look of it seems fine. Can you try this out? Or maybe point me to an old working `export_albany.in` on some machine?

mperego:
thanks @bartgol . Can you post the `export_albany.in` you get?
In the meantime I'll rebuild Trilinos

bartgol:
> thanks @bartgol . Can you post the `export_albany.in` you get? In the meantime I'll rebuild Trilinos



> thanks @bartgol . Can you post the `export_albany.in` you get? In the meantime I'll rebuild Trilinos

I've made a few changes to the branch (push coming), but this is what I get (had to rename adding `.txt` or else github would not attach it).
[export_albany.in.txt](https://github.com/sandialabs/Albany/files/9508214/export_albany.in.txt)

mperego:
here is how it used to be 
[albany_export.in.txt](https://github.com/sandialabs/Albany/files/9508321/albany_export.in.txt)

bartgol:
I think they are equivalent. The old file did not have absolute paths, so it set linker search dir, and then linked libs by name only. We can of course do that, but unless we plan to support 2+ versions of trilinos and/or albany, absolute path linking is fine, imho.

mperego:
OK, let me check if it works

mperego (approved):

bartgol:
@mperego Should we merge?

mperego:
> @mperego Should we merge?

Yes, let's give this a try

ikalash:
Is it possible that this PR broke CISM-Albany?

https://my.cdash.org/build/2214176/configure

```
CMake Error at /project/projectdirs/piscees/nightlyCoriCDash/build/AlbanyInstall/lib64/Albany/cmake/AlbanyConfig.cmake:11 (message):
  File or directory referenced by variable Albany_INCLUDE_DIRS does not exist
  !
Call Stack (most recent call first):
  /project/projectdirs/piscees/nightlyCoriCDash/build/AlbanyInstall/lib64/Albany/cmake/AlbanyConfig.cmake:26 (set_and_check)
  CMakeLists.txt:211 (FIND_PACKAGE)


-- Configuring incomplete, errors occurred!
See also "/global/project/projectdirs/piscees/nightlyCoriCDash/build/CoriCismAlbany/CMakeFiles/CMakeOutput.log".
```

bartgol:
Yes, I made a small (but fatal) find+replace mistake. Fix coming.

PR #831, opened 2022-09-09
Fix installed cmake config file

bartgol:
There was a find+replace error, and I also forgot to update the AlbanyConfig.cmake.in file.


ikalash:
Thanks @bartgol this fixes the problem.

ikalash (approved):

PR #833, opened 2022-09-13
Pyalbany new API and paper examples

kliegeois:
This PR includes the new updated API of PyAlbany that hides the RCP from the users.

It adds example of use of PyAlbany used into the future PyAlbany paper.


mperego (approved):
It looks good to me!

PR #834, opened 2022-09-15
Fix annoying issue of ldl-2 showing up in export_albany.in

bartgol:


mperego (approved):
Thanks!

bartgol:
Note: the python script that generates `export_albany.in` will fail if the library name contains a dot `.` in the name. E.g., if we end up linking against `some.cool.lib.so`, the script will add `-lsome` to the link line, which would obviously fail. I don't know of any major library that would use the dot in the name, but wanted to put this warning here.

ikalash:
@bartgol : I'm getting errors when I try to build Albany that look like this. 

```
[329/333] Generating ../export_albany.in
FAILED: export_albany.in
cd /nightlyCDash/build/IKTAlbany/dummy && /nightlyCDash/repos/Albany/cmake/CreateExportAlbany --bin-dir /nightlyCDash/build/IKTAlbany --install-lib-dir /nightlyCDash/build/IKTAlbanyInstall/lib64
Traceback (most recent call last):
  File "/nightlyCDash/repos/Albany/cmake/CreateExportAlbany", line 91, in <module>
    _main_func(__doc__)
  File "/nightlyCDash/repos/Albany/cmake/CreateExportAlbany", line 29, in _main_func
    with infile.open('r') as fd:
  File "/usr/lib64/python3.6/pathlib.py", line 1183, in open
    opener=self._opener)
  File "/usr/lib64/python3.6/pathlib.py", line 1037, in _opener
    return self._accessor.open(self, flags, mode)
  File "/usr/lib64/python3.6/pathlib.py", line 387, in wrapped
    return strfunc(str(pathobj), *args)
FileNotFoundError: [Errno 2] No such file or directory: '/nightlyCDash/build/IKTAlbany/dummy/CMakeFiles/dummy.dir/link.txt'
[330/333] Linking CXX executable dummy/dummy
[331/333] Linking CXX executable src/Albany
[332/333] Linking CXX executable src/AlbanyAnalysis
ninja: build stopped: cannot make progress due to previous errors.
Command exited with the value: 1
MakeCommand:/usr/bin/cmake --build . --config "Release" --target "install" -- -k 999999
Error(s) when building project
```

I think it is related to this PR.  Could you please have a look?

bartgol:
Hard to debug without actually accessing the machine. The file that is missing is a CMake-generated file, which should be there.

bartgol:
Ah, wait, you are using Ninja. I've never checked the structure of the CMakeFiles folder with ninja generator. I can try to see what that is, and perhaps modify the script. I can't promise a quick fix though. If this build is important, you can fall back to gnu makefile generator for the time being...

ikalash:
I did some testing on cee-compute020 and it does appear to be due to ninja.  I am attaching a configure script you can use to reproduce the problem.  If you build on cee-compute020, you should be able to just point to the Trilinos that is already built there.  The modules to use are: https://github.com/sandialabs/Albany/blob/master/doc/dashboards/cee-compute011.sandia.gov/sems-intel-modules.sh .  I could switch back to plain cmake / make, but I like ninja b/c it is so much faster.  It would be great if you could have a look at this.  I suspect it will affect a lot of the builds since many of the nightlies use ninja now.

ikalash:
BTW did INSTALL_ALBANY to away as a configure option?

bartgol:
Yes, it was pointless, imho. CMake native `install` commands are already only ran at `make install` time, so that option was not adding anything special.

bartgol:
> I did some testing on cee-compute020 and it does appear to be due to ninja. I am attaching a configure script you can use to reproduce the problem. If you build on cee-compute020, you should be able to just point to the Trilinos that is already built there. The modules to use are: https://github.com/sandialabs/Albany/blob/master/doc/dashboards/cee-compute011.sandia.gov/sems-intel-modules.sh . I could switch back to plain cmake / make, but I like ninja b/c it is so much faster. It would be great if you could have a look at this. I suspect it will affect a lot of the builds since many of the nightlies use ninja now.

Yeah, ninja produces different build files, so I need to parse a different file in order to grab albany's dependencies.

ikalash:
It would be great if you could look into this @bartgol .  Good to know about the install option.  I agree it was redundant.

ikalash:
BTW I forgot to attach the configure script in my earlier post.  Here it is.
[do-cmake-ninja.txt](https://github.com/sandialabs/Albany/files/9610882/do-cmake-ninja.txt)

PR #844, opened 2022-09-28
Fix creation of export_albany.in when CMAKE_GENERATOR=Ninja

bartgol:
This should allow to use Ninja as cmake generator, and produce a valid `export_albany.in` file. I tested it on my machine, and the generated file is equivalent to that generated when using Unix Makefiles.

This will have to be tested on some other machines, to make sure a) it works for Ninja, and b) doesn't break Unix Makefiles. Unfortunately, the way cmake generates the build/link files is not always consistent (e.g., absolute paths vs relative paths, link with full filepath vs -l$libname, etc), so it's likely I did not cover certain cases.

Further note: I based my implementation on whatever file ninja generated for me on my laptop, which is `build.ninja` in the binary root dir. I _think_ this is indeed the general case, but will have to keep an eye out to make sure of it.

### DIFF
--- a/cmake/CreateExportAlbany
+++ b/cmake/CreateExportAlbany
@@ -6,72 +6,156 @@ import sys, os, argparse, pathlib
 ###############################################################################
 def parse_command_line(args, description):
 ###############################################################################
-  parser = argparse.ArgumentParser()
+    parser = argparse.ArgumentParser()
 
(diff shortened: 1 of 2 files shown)


ikalash:
I tested this on my machine camobap and all looks good.  I will merge.

PR #847, opened 2022-10-04
Replace private Kokkos headers with Kokkos_Core.hpp

mcarlson801:
Including private Kokkos headers is now deprecated as of 3.7, replacing existing private headers with Kokkos_Core.hpp


jewatkins (approved):
Looks good, thanks!

PR #849, opened 2022-10-11
Update intel build on blake

jewatkins:
This will update the intel build on blake to intel 19.3.199 (which is necessary for the next trilinos/kokkos update). All tests passed on a serial build. There were tpl issues with oneapi. I think we'd have to build our own tpls if we want to update further.

Next step after this push would be pardiso and c++17.

@mperego @ikalash


bartgol:
@jewatkins wait, are you saying this fixes also the test that used to fail with 1% diffs on the solution avg? Or was that only for OpenMP builds?

jewatkins:
No, this has nothing to do with the failing performance tests. This is just to update our intel build which will soon be too old for trilinos/kokkos. It might be interesting to see what the new value is with this compiler.

jewatkins:
^Pardiso works too so I'll just add that in here too.

mperego (approved):
lgtm, thanks!

jewatkins:
I confirmed c++17 works too.

PR #850, opened 2022-10-12
Simplified GatherSolution evaluator

bartgol:
I incorporated this as part of my branch that used the dof manager, but I figured I pushed this as a separate PR, so that we limit the mods to be reviewed in the next PR.

I ran on blake and weaver, to verify performance was unaffected, and it seems that's the case. That's expected, since branch predictor is making the if statement inside the loops virtually irrelevant (b/c the if always evaluates to the same branch).

This reduces code length and makes it easier to maintain, with less code duplication.


jewatkins (approved):
Looks good, thanks for cleaning this up.

PR #851, opened 2022-10-12
Perlmutter module script update

mcarlson801:
Updated perlmutter module scripts to match E3SM modules (https://github.com/E3SM-Project/E3SM/blob/master/cime_config/machines/config_machines.xml). Trilinos has configuration errors when using the `gpu_nvidia` and `cpu_amdclang` modules so that will require some debugging if we want to target those compilers. I've also updated the Trilinos config scripts to use ZEN3 now that we have a new enough version of gcc available.

There is currently an issue with the SuperLU module on Perlmutter and I've opened a ticket with NERSC to see if it can be resolved. If we need SuperLU in the meantime, I can try building it myself.


jewatkins (commented):
Looks good. The env variables seem reasonable. We may want to try building openmp instead of serial at some point (it should be slightly faster but less robust which is why I always build serial). I'd move the amdclang and nvidia modules into it's own directory, maybe `testing`, so as not to confuse others who look at this directory.

jewatkins:
@mperego Looks like superlu on perlmutter might take some time. Let us know if you want us to build a local version.

mcarlson801:
> We may want to try building openmp instead of serial at some point (it should be slightly faster but less robust which is why I always build serial).

Should I drop the OMP flags until then?

jewatkins:
No, keep them for the mpas builds which might have some openmp.

mperego:
> @mperego Looks like superlu on perlmutter might take some time. Let us know if you want us to build a local version.

let's hear first the feedback from NERSC, then we can decide what to do. It's not urgent.

jewatkins (approved):

PR #859, opened 2022-11-09
Correctly handle -L/path entries when generating export_albany.in

bartgol:
This should fix E3SM-Project/spack#6.


ikalash:
I have verified that this fixes the problem in my spack build!  Thanks for working on this, I know it was a pain and required learning new stuff about the spack build.  But at least now you are familiar with the spack build!

ikalash (approved):

PR #861, opened 2022-11-11
Switch from workset arrays to DOFManager

bartgol:
This huge PR introduces panzer's DOFManager class stack in Albany, and switches the gather/scatter evaluators (as well as other related utilities/structures) from using n-dim workset arrays to use DOFManager's utilities. Some comments:

- I added an Albany verison of DOFManager and ConnManager, mostly to offer some extra interfaces (e.g., avoid always specifying the block name, since we currently always have one elem block, or provide both device and host views).
- DOF/Conn managers are stored in the abstract disc class, and I ideally hope we can remarkably slim down STKDiscretization
- I reorganized a bit the unit test folder

This is still work in progress, but some interfaces are somewhat stable, and can be already reviewed and/or used in other work. In particular, I still have to update how STK disc computes the graph; but (perhaps not right away) I'm hoping to be able to do this outside the STK classes, handling it completely via general interfaces and DOFManager's.

You are welcome to ignore the PR for a while, but it might help to start looking at it, and perhaps put the check mark on some files, to reduce the review workload gradually.


bartgol:
Update: I added a couple of unit tests, to verify the internal state of STKDiscretization. I need to add more, but so far I added one that checks the Jac pattern and one that checks nodesets info, both for a 2D structured mesh. They finally both pass.

I now have to slowly remove all the remaining uses of workset arrays throughout the disc folder. The evaluator folder and LandIce should already be done, but all the STK field managers still use the NodalDOFManager struct (not to be confused with panzer's DOFManager), and they will have to move to DOFManager.

src/disc/Albany_DOFManager.hpp
```diff
@@ -0,0 +1,93 @@
...
+class DOFManager : public panzer::DOFManager {
+public:
+  // Initializes DOF manager
+  DOFManager (const Teuchos::RCP<ConnManager>& conn_mgr,
+              const Teuchos::RCP<const Teuchos_Comm>& comm);
+
+  void build ();
+
+  // Returns (elem_LID,idof)->dof_lid numbering
+  const DualView<const int**>& elem_dof_lids () const;
```
mcarlson801, line 29: Are these instances of the lightweight DualView that you added in Albany_KokkosTypes? Maybe it would be good to add something to differentiate this from the Kokkos DualView to avoid confusion down the road.

bartgol: Yeah, that's it. I guess I could change the name of the type, I just don't have any good idea (all combination I can think of are clunky), so suggestions are welcome. I also thought of using Kokkos's dual view, but in order to sync the two views, one needs to explicitly mark them as modified, and I didn't want to bother. But perhaps, especially considering we don't update these often (ever?), I could really use kokkos dual views...

bartgol: Also, I guess the same could be said for `Albany::ConnManager` and `Albany::DOFManager`, which may be confused with their panzer base classes. I can of course prepend Albany to both of them, but it makes them a bit longer. Perhaps ConnMgr and DOFMgr, though they might be too concise?

mcarlson801: I feel like at least with the Conn manager and DOF manager, the context is probably enough to know what they are our implementations. But if we start using Kokkos DualViews when removing UVM dependence, things could get messy without prepending `Albany::`. We could cut down on line bloat by just having a single `using dview = Albany::DualView..` like you've used elsewhere. Then at least it is made explicit somewhere in the file.

bartgol: I guess I don't use the type name `DualView` that often, so probably I can prepend Albany:: in those few places where I do. I am already doing it in some places, so might as well do it everywhere.

bartgol:
Update. After merging master, I'm getting `6 tests failed out of 38` in the corePDEs tests folder: 
```
The following tests FAILED:
	 10 - corePDEs_SteadyHeatConstrainedOpt2D_Dirichlet_Dist_Param_Analysis (Failed)
	 29 - corePDEs_SideSetLaplacian_3D (Failed)
	 34 - corePDEs_SteadyHeat2DSS_Epetra (Failed)
	 35 - corePDEs_SteadyHeat2DSS_Tpetra (Failed)
	 36 - corePDEs_SteadyHeat2DSS_dydx_Epetra (Failed)
	 37 - corePDEs_SteadyHeat2DSS_dydx_Tpetra (Failed)
```

Unfortunately, aside from the SideSetLaplacian test, which fails b/c of a `map::at` exception (easy to fix, and hopefully working afterwards), the other tests do not pass due to failed comparisons. These are the harder to debug. Test 10 in particular, as the other might point to some issues in Neumann evaluators (they are all robin bcs, if I understand correctly the input files).

LandIce was still full of failures before the merge, but I decided to fix all corePDEs tests first.

tests/unit/CMakeLists.txt
```diff
@@ -4,41 +4,98 @@
...
+#    PREFIX <p>         String to prepend to test name
+#    SOURCES <files>    List of files to build the exec
+#    LIBS <libs>        List of libs to link to the exec.
+#    SERIAL_ONLY        Whether we need to run the test only in serial mode (default: OFF)
+#
+# If prefix is present, test names start with ${PREFIX}_${NAME},
+# otherwise they simply start with ${NAME}.
+# Note: the main albanyLib library is already linked, via the albany_ut_main, so no need to specify it
+
+macro (create_unit_test)
```
mperego, line 19: this is much cleaner. thanks!

tests/unit/CMakeLists.txt
```diff
@@ -4,41 +4,98 @@
...
-target_link_libraries (string_utils PUBLIC albany_ut_main)
-add_test(NAME ${testNameRoot}_StringUtils_Serial
-         COMMAND ${SERIAL_CALL} string_utils)
+create_unit_test (NAME    StringUtils
+                  PREFIX  ${dirName}
+                  SOURCES StringUtils.cpp
+                  SERIAL_ONLY)
+
+# Check HessianVec fad
+create_unit_test (NAME    HessianVecFad
```
mperego, line 93: thanks for creating new tests!

src/disc/Albany_AbstractDiscretization.hpp
```diff
@@ -41,39 +44,100 @@ class AbstractDiscretization
   //! Destructor
   virtual ~AbstractDiscretization() = default;
 
+  //! Get the DOF manager
+  Teuchos::RCP<const DOFManager>
+  getNewDOFManager (const std::string& fieldName) const {
```
mperego, line 49: Are you going to rename this `getDOFManager` before merging the PR?

mperego (commented):
This is massive!
Overall looks good to me. I didn't go too much into the details but it looks good.
Thanks!!

src/disc/stk/Albany_ExtrudedSTKMeshStruct.cpp
```diff
@@ -342,18 +342,28 @@ void Albany::ExtrudedSTKMeshStruct::setBulkData(
...
 
   std::vector<double> ltr(layerThicknessRatio.size());
   for(size_t i=0; i< ltr.size(); ++i) {
     ltr[i]=layerThicknessRatio[i];
   }
   fieldContainer->getMeshVectorStates()["layer_thickness_ratio"] = ltr;
   fieldContainer->getMeshScalarIntegerStates()["ordering"] = static_cast<int>(Ordering);
-  fieldContainer->getMeshScalarIntegerStates()["stride"] = (Ordering==LAYER) ? lVertexColumnShift : vertexLayerShift;
-  fieldContainer->getMeshScalarInteger64States()["global_stride"] = (Ordering==LAYER) ? vertexColumnShift : vertexLayerShift;
+  fieldContainer->getMeshScalarIntegerStates()["local_stride"] = (Ordering==LAYER) ? lElemColumnShift : elemLayerShift;
```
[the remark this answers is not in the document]
bartgol: Oh, I see. For stk-related stuff, the nodal layer data still gives the correct ids. I'll add it back.

And yes, I can add "elem" to the var name, for clarity.

src/disc/stk/Albany_ShardsHack.hpp
```diff
@@ -0,0 +1,81 @@
+#ifndef ALBANY_SHARDS_HACK_HPP
```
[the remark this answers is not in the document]
bartgol: I removed it in a recent commit, but in order to remove it I need the change in shards, for Particle. I can submit a PR in trilinos for that.

src/disc/stk/Albany_STKDiscretization.cpp
```diff
@@ -1054,325 +992,246 @@ STKDiscretization::setSolutionFieldMV(
...
+  for (const auto& it : name_to_partAndDim) {
+    const auto& field_name = it.first;
+    const auto& part_name  = it.second.first;
+    const auto& dof_dim    = it.second.second;
 
-      dofs->dofManager.setup(numComponents,
-                             numNodes,
-                             maxGlobalNodeGID,
-                             interleavedOrdering);
+    // NOTE: for now we hard code P1. In the future, we must be able to
```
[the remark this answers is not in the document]
bartgol: Well, there's no way of choosing other elements. There's no way for the user to request them. In the future, we want to offer some way to specify the solution (blocks) disc type/order. Something like
```
Discretization:
   # 2d Navier stokes
   Solution Names: [u1, u2, p]
   Solution FE Types: [P1, P1, P0]
```
Until then, we just assume each solution component is P1.

src/disc/Albany_AbstractDiscretization.hpp
bartgol: Yes. I could do it now, since the old NodalDOFManager has already been removed. But I might just wait till the end for the cleanup stuff.

bartgol:
The SideSetLaplacian test was hard to fix, but revealed several bugs, so I'm glad we had it. Now, in the corePDEs folder, the tests still failing are

```
corePDEs_SteadyHeat2DSS_Epetra
corePDEs_SteadyHeat2DSS_Tpetra
corePDEs_SteadyHeat2DSS_dydx_Epetra
corePDEs_SteadyHeat2DSS_dydx_Tpetra
corePDEs_SteadyHeatConstrainedOpt2D_Dirichlet_Dist_Param_Analysis
```
The first 4 are all using Robin BCs, while no other test does, so hopfully it's a simple fix in the Neumann class. The last one is more difficult, since other tests share some similarities with it, and they pass.

Also, I have disabled the subfolders `SteadyHeat2DEB`, `HeatQuadTri`, and `SteadyHeat2DRobin`, since we cannot currently handle 2 element blocks. I can probably support that with some work, so long as the two blocks have the same topology, but the QuadTri test might be out of reach now (waiting for some input from Roger on interaction between blocks with different topology and DOFManager).

I have not run LandIce/DemoPDEs tests since the last fixes, so that I'm leaving for the winter break with the hope that they might be mostly working...

src/Albany_KokkosTypes.hpp
```diff
@@ -42,6 +51,122 @@ using DeviceLocalGraph  = Kokkos::StaticCrsGraph<LO, Kokkos::LayoutLeft, KokkosN
...
+  DualView (const DualView&) = default;
+
+  template<typename SrcDT>
+  typename std::enable_if<
+    std::is_same<SrcDT,nonconst_DT>::value,
+    DualView<DT>
+  >::type&
+  operator= (const DualView<SrcDT>& src) {
+    d_view = src.d_view;
+    h_view = src.d_view;
```
mcarlson801, line 104: I think this is a typo. This should be `h_view = src.h_view` right?

src/disc/Albany_ConnManager.hpp
```diff
@@ -46,6 +46,15 @@ class ConnManager : public panzer::ConnManager {
...
   virtual int part_dim (const std::string& part_name) const = 0;
+  
+  /** Get vector of bools associated to connectivity for a particular element, indicating whether the entity is owned by this rank
+    *
+    * \param[in] localElmtId Local element ID
+    *
+    * \returns Pointer to beginning of bools, with total size
+    *          equal to <code>getConnectivitySize(localElmtId)</code>
+    */
+  virtual const std::vector<bool>& getOwnership(LO localElmtId) const = 0;
```
bartgol, line 57: `std::vector<bool>` is a dangerous type. Iterators and access members are different from a normal `std::vector`. I've had my share of bad surprises with it, so I would use either a `std::vector<int>`, `std::vector<char>`, or `std::deque<bool>`.

src/disc/stk/STKConnManager.hpp
```diff
@@ -67,6 +67,20 @@ class STKConnManager : public ConnManager {
...
+    *
+    * \param[in] localElmtId Local element ID
+    *
+    * \returns vector of bools, with total size
+    *          equal to <code>getConnectivitySize(localElmtId)</code>
+    */
+  const std::vector<bool>& getOwnership(LO localElmtId) const override {
+     auto begin = m_ownership.begin()+m_elmtLidToConn[localElmtId];
+     auto end = begin+m_connSize[localElmtId];
+     m_owned.assign(begin,end);
```
bartgol, line 81: Why do you need `m_owned` on top of `m_ownership`?

mperego: it's because vector<bool> is weird beast, as you said. Instead of returning a pointer to &m_ownership[i], as done for getConnectivity,  I'm creating a new subvector and returning it. In this way we are safe. If you prefer to convert it to vector<int> go for it.

bartgol: Yeah, I think the risk of small mods that cause unexpected behavior downstream (and hours of debugging) is not negligible. I wish they never implemented that specialization of std::vector... I will change it to something safer.

src/disc/Albany_ConnManager.hpp
mperego: Yeah, in fact it gave some troubles because `&m_ownership[i]` is not ` bool*`. I think the way I'm using it now is safe, see the reply to your comment in `STKConnManager.hpp`. However, feel free to change it.

mperego (approved):
@bartgol I think this PR is almost ready to be merged. A couple of things it might be better to do before merging are:
1. change `getNewDOFManager()` to `getDOFManager()`
2. update the reference map file for test ` PyAlbany_SteadyHeat_StateMap` (I'll take care of this)

src/disc/Albany_ConnManager.hpp
```diff
@@ -16,6 +16,11 @@
 
 namespace Albany {
 
+enum Ownership {
+  Owned = 1,
+  Ghosted = 2
+};
+
```
mperego, line 23: Nice!

bartgol:
@mperego the last commit fixes all the warnings on CUDA. Also, testing with the proper FAD choices makes all tests pass. If you're ok with the last changes, I can merge.

PR #865, opened 2022-12-06
Refactor of the Hessian response for allowing for custom solvers

mperego:
Until now the Hessian pp matrix has always been a Crs Matrix that can be solved for using standard solvers. However, in some cases it is not easy or performant to explicitly build the matrix. This PR allows a custom linear operator and solves for each response. If such operator is not provided, the Hessian response matrix is built using the current coloring-seeding approach that relies on Hessian vectors products and on the assumption of a finite-element like pattern (for distributed parameters)

This PR also implements the solver for the quadratic response.

A few tests are modified to comply with the changes.

This PR should be merged after https://github.com/trilinos/Trilinos/pull/11353.

### DIFF
--- a/src/Albany_Application.cpp
+++ b/src/Albany_Application.cpp
@@ -395,6 +395,45 @@ Application::initialSetUp(const RCP<Teuchos::ParameterList>& params)
 #endif
   }
 
+  //validate Hessian parameters
+  if(problemParams->isSublist("Hessian")) {
+    Teuchos::RCP<Teuchos::ParameterList> validHessianParams = Teuchos::rcp(new Teuchos::ParameterList("validHessianParams"));
+    validHessianParams->set<bool>("Write Hessian MatrixMarket", false);
+    validHessianParams->set<bool>("Use AD for Hessian-vector products (default)", true);
+
+    validHessianParams->sublist("Residual").set<bool>("Use AD for Hessian-vector products (default)", false);
+    validHessianParams->sublist("Residual").set<std::string>("Enable AD for Hessian-vector product contributions of", "(x,x) (p0,p0) (p0,p1) (p1,p0) (p1,p1)");
+    validHessianParams->sublist("Residual").set<std::string>("Disable AD for Hessian-vector product contributions of", "(x,p0) (x,p1) (p0,x) (p1,x)");
+
+    // We want to allow the user to add hessian settings for responses (parameters) even if those responses (parameters) are not used.
+    int maxNumResonses(3),maxNumParameters(3);
+
+    //Make sure that the problem does not require more responses (parameters) than anticipated. 
+    if (problemParams->isSublist("Response Functions")) 
+      maxNumResonses = std::max(maxNumResonses, problemParams->sublist("Response Functions").get<int>("Number Of Responses"));
+    if (problemParams->isSublist("Parameters")) 
+      maxNumParameters = std::max(maxNumParameters, problemParams->sublist("Parameters").get<int>("Number Of Parameters"));
+
+    for(int response_index = 0;  response_index < maxNumResonses; response_index++) {
+      auto& validHessianResponseParams = validHessianParams->sublist(util::strint("Response", response_index));
+      validHessianResponseParams.set<bool>("Use AD for Hessian-vector products (default)", false);
+      validHessianResponseParams.set<bool>("Reconstruct H_pp", false);
+      validHessianResponseParams.set<std::string>("Enable AD for Hessian-vector product contributions of", "(x,x) (p0,p0) (p0,p1) (p1,p0) (p1,p1)");
+      validHessianResponseParams.set<std::string>("Disable AD for Hessian-vector product contributions of", "(x,p0) (x,p1) (p0,x) (p1,x)");
+      for (int i=0; i<maxNumParameters; ++i) {
+        auto& pl = validHessianResponseParams.sublist(util::strint("Parameter", i), false,"");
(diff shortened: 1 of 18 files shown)


src/Albany_Application.cpp
```diff
@@ -395,6 +395,45 @@ Application::initialSetUp(const RCP<Teuchos::ParameterList>& params)
...
+    Teuchos::RCP<Teuchos::ParameterList> validHessianParams = Teuchos::rcp(new Teuchos::ParameterList("validHessianParams"));
+    validHessianParams->set<bool>("Write Hessian MatrixMarket", false);
+    validHessianParams->set<bool>("Use AD for Hessian-vector products (default)", true);
+
+    validHessianParams->sublist("Residual").set<bool>("Use AD for Hessian-vector products (default)", false);
+    validHessianParams->sublist("Residual").set<std::string>("Enable AD for Hessian-vector product contributions of", "(x,x) (p0,p0) (p0,p1) (p1,p0) (p1,p1)");
+    validHessianParams->sublist("Residual").set<std::string>("Disable AD for Hessian-vector product contributions of", "(x,p0) (x,p1) (p0,x) (p1,x)");
+
+    // We want to allow the user to add hessian settings for responses (parameters) even if those responses (parameters) are not used.
+    int maxNumResonses(3),maxNumParameters(3);
```
kliegeois, line 409: Why 3?

src/Albany_ModelEvaluator.cpp
```diff
@@ -567,10 +567,19 @@ ModelEvaluator::create_hess_g_pp( int j, int l1, int l2 ) const
...
 
+  auto pl = app->getProblemPL()->sublist("Hessian").sublist(util::strint("Response",j)).sublist(util::strint("Parameter",l1));
+  bool HessVecProdBasedOp = pl.get("Reconstruct H_pp using Hessian-vector products",true);
+
   if (l1 < num_param_vecs) {
+    TEUCHOS_TEST_FOR_EXCEPTION(!HessVecProdBasedOp, std::logic_error, 
+            std::endl
+                << "Error!  Albany::ModelEvaluator::create_hess_g_pp():  "
+                << "Hessian pp operator for scalar response " << j << " and parameter " << l1 << " can only be reconstructed via Hessian-vector products"
+                << std::endl);
```
kliegeois, line 578: Just to be sure, the Hessian associated to a non-distributed parameter can only be created using the reconstruction via Hessian-vector products, is that right? If so, can we update the error message with  `<< " and (non-distributed) parameter " <<` or so.

kliegeois (commented):
Thanks @mperego , I just have two quick comments.

src/Albany_Application.cpp
mperego: This is somewhat arbitrary. I don't want to choose a number that is too large, but at the same time I don't want to set it to zero, otherwise one has to delete the unused Hessian responses settings. we can chat about this.

src/Albany_ModelEvaluator.cpp
mperego: good point. I'll do that.

src/Albany_Application.cpp
bartgol: You could validate without recursion if num_params or num_responses is 0. And validate the sublists from inside the pfor.

mperego: The problem is not the recursion. With the approach you suggest I cannot validate the list at the Hessian level.
Here is how a list could look like
```
Hessian:
  Residual: {}
  Response 0: 
    Parameter 0: {}
```
To validate the Hessian list I need to make a guess of how many responses are listed.

bartgol: I see, thanks.

src/Albany_Application.cpp
```diff
@@ -395,6 +395,45 @@ Application::initialSetUp(const RCP<Teuchos::ParameterList>& params)
...
+    if (problemParams->isSublist("Response Functions")) 
+      maxNumResonses = std::max(maxNumResonses, problemParams->sublist("Response Functions").get<int>("Number Of Responses"));
+    if (problemParams->isSublist("Parameters")) 
+      maxNumParameters = std::max(maxNumParameters, problemParams->sublist("Parameters").get<int>("Number Of Parameters"));
+
+    for(int response_index = 0;  response_index < maxNumResonses; response_index++) {
+      auto& validHessianResponseParams = validHessianParams->sublist(util::strint("Response", response_index));
+      validHessianResponseParams.set<bool>("Use AD for Hessian-vector products (default)", false);
+      validHessianResponseParams.set<bool>("Reconstruct H_pp", false);
+      validHessianResponseParams.set<std::string>("Enable AD for Hessian-vector product contributions of", "(x,x) (p0,p0) (p0,p1) (p1,p0) (p1,p1)");
```
bartgol, line 421: Can I ask what p0 and p1 refer to? Are you assuming there are only two parameters? Or is this to say that the two ps in H_pp could refer to different parameters?

kliegeois: @bartgol it is the later. It suggests that $\boldsymbol{p}^{(1)} != \boldsymbol{p}^{(2)}$.

mperego: Yes, in this case we are considering two parameters, but you could have terms like   (p0, p4), (x, p17), etc.

mperego:
I forgot to commit two new files. See last commit.

kliegeois (approved):
It looks good to me, thanks!

mperego:
Thanks @kliegeois!

PR #869, opened 2022-12-14
Update of the examples based on the reviews of the PyAlbany paper

kliegeois:
This PR updates the PyAlbany examples to follow reviewers's comments.

### DIFF
--- a/PyAlbany/examples/paper/.gitignore
+++ b/PyAlbany/examples/paper/.gitignore
@@ -2,3 +2,4 @@ steady2d*
 *.jpeg
 *.sh
 *.txt
+input_scalar_updated.yaml
--- a/PyAlbany/CMakeLists.txt
+++ b/PyAlbany/CMakeLists.txt
@@ -26,17 +26,19 @@ ENDIF()
 
 # Get the python version
 EXECUTE_PROCESS(COMMAND ${PYTHON_EXECUTABLE} -c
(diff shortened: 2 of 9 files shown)


PyAlbany/examples/paper/.gitignore
```diff
@@ -1,4 +1,5 @@
 steady2d*
 *.jpeg
 *.sh
-*.txt
\ No newline at end of file
+*.txt
+input_scalar_updated.yaml
```
mperego, line 5: why did you add `input_scalar_updated.yaml` to `.gitignore`?

mperego (approved):
I left a comment. Other than that it looks good to me

PyAlbany/examples/paper/.gitignore
kliegeois: Because I used an input file `PyAlbany/examples/paper/input_scalar_to_update.yaml` as a "template" to generate an updated one (updated in the script) and then write it to the disc to use as input argument of the executable. We don't want to commit the updated one as it changes all the time.

This is in the context of the MC simulation using the C++ executable.

mperego: I see. It makes sense.

kliegeois:
Can I merge it @mperego ?

mperego:
Sure!

PR #870, opened 2022-12-19
Organization and cleaning of tests and problems. Phase 2: rearrangement

mperego:
This is part of the work to reorganize Albany folders and tests, see[ here](https://acme-climate.atlassian.net/wiki/spaces/FAN/pages/3539469051/Albany+tests+and+problems+re-organization) and follows PR #854.

In particular the new tests structure is
```
tests
- unit
- corePDEs
- demoPDEs
- landIce
```

Tests always have in the name one of the above folder, so that it's easy to identify where they belong.
We got rid of the distinction between small and large tests, however, large tests can be built optionally as before.
Remove a couple of tests that are no longer relevant, in particular the Performance tests (We have another repo now for performance testing).

The source folder has been re-organized as well as follows
```
src
- corePDEs
  - evaluators
  - problems
- demoPDEs
  - evaluators
  - problems
- landIce
  - evaluators
  - problems
  - mpasInterface
  - 
- evaluators
- problems
- utility
- disc
```

Evaluators that are no longer used have been removed.

This is only a proposed change, we can discuss tomorrow at the Albany meeting what we prefer to do.
Also, we need to manage conflicts with the DofManager PR. (An option would be to split this PR in two, one addressing the test refactor and one the source refactor and start merging only the test refactor PR).


src/CMakeLists.txt
```diff
@@ -44,6 +43,7 @@ set (ALBANY_INCLUDE_DIRS
      ${Albany_SOURCE_DIR}/src/disc/stk
      ${Albany_SOURCE_DIR}/src/disc
      ${Albany_SOURCE_DIR}/src/utility
+     ${Albany_SOURCE_DIR}/src/corePDEs/evaluators
```
jewatkins, line 46: A small nitpick but I would probably call this directory `CorePDEs` and the other one `DemoPDEs` to stick with the very loose convention of using CamelCase (e.g. `LandIce`). Probably better than using snake_case which is used for all other directories so we avoid having to change LandIce.

jewatkins (commented):
Looks good.

bartgol:
My browser cannot handle the large number of files to diff, and hangs. I agree with Jerry that making case uniform looks cleaner. I'm ok with the reorg.

mperego:
Thanks @bartgol and @jewatkins. I'm ok using CamelCase, but then for consistency we should change all the other folders? (`evaluators`, `problems`, `utility`, `disc`.. etc)?

bartgol:
Ugh, yeah, that's annoying. And lower case is easier when typing and autocompleting. I am ok with whatever you pick then.

Btw, if you wanted to change `LandIce` into `landice`, I would be fine with that too (saves me from pressing the annoying Shift key...).

jewatkins:
As long as everyone's okay with changing LandIce to landice or land_ice then that's probably the easier option. I was thinking corePDEs was camelCase. Would it then be `corepdes` or `core_pdes`?

mperego:
We could do camelCase and have `landIce`, `corePDEs`, `demoPDEs`

jewatkins:
camelCase is okay. I think at least the interface dirs would have to be changed: landIce/interfaceWithMPAS

mperego:
b.t.w. also the LandIce tests change name, because we are pre-pending landIce to their current names (and fix a few inconsistencies)

jewatkins (approved):

PR #878, opened 2022-12-22
Prune unused CMake option

bartgol:
I found it while working on something else, and figured I would spend 30sec  to clean it up while I was at it.


mperego (approved):
lgtm. Thanks!!

PR #880, opened 2022-12-25
Use camelCase folder naming convention for PyAlbany

kliegeois:


mperego:
@kliegeois looks good. Thanks@! 
In Albany tests we got rid of the `small`/`large` folders, however we run the large tests only when ALBANY_LARGE_TESTS is defined. We also renamed `LandIce` into `landIce`. So, you could consider doing the same. Also, `inputOutput` might be better than `iO`.

kliegeois:
@mperego that sounds good with me. I will update the PR to be consistent with the rest of the repo.

kliegeois:
(I just checked and the ctest tests using spack passed)

mperego (approved):
Looks good to me. Thanks!

kliegeois:
Thanks @mperego !

PR #881, opened 2023-01-04
Remove deprecated Kokkos function calls

kliegeois:
Remove deprecated Kokkos function calls in PyAlbany.


mperego (approved):
It looks good to me. Thanks!

PR #882, opened 2023-01-05
Fix to Dirichlet conditions for DistParamDeriv evaluation.

mperego:
In order to allow different Dirichlet conditions to be applied to the same dof (the last one applied is the one that gets prescribed), we need to pre-process the Lagrangian multiplier in the pre-evaluation phase. 

Fixed regression values in a couple of tests accordingly and tighten regression tolerances of a few tests.

### DIFF
--- a/src/evaluators/bc/PHAL_DirichletField_Def.hpp
+++ b/src/evaluators/bc/PHAL_DirichletField_Def.hpp
@@ -228,6 +228,40 @@ DirichletField(Teuchos::ParameterList& p) :
 // **********************************************************************
 template<typename Traits>
 void DirichletField<PHAL::AlbanyTraits::DistParamDeriv, Traits>::
+preEvaluate(typename Traits::EvalData dirichletWorkset) {
+  bool trans = dirichletWorkset.transpose_dist_param_deriv;
+
+  if (trans) {
+    // For (df/dp)^T*V we zero out corresponding entries in V
+    Teuchos::ArrayRCP<Teuchos::ArrayRCP<ST>> Vp_nonconst2dView = Albany::getNonconstLocalData(dirichletWorkset.Vp_bc);
+    bool isFieldParameter =  dirichletWorkset.dist_param_deriv_name == this->field_name;
+    int num_cols = dirichletWorkset.Vp->domain()->dim();
+    const std::vector<std::vector<int> >& nsNodes = dirichletWorkset.nodeSets->find(this->nodeSetID)->second;
+
+    if(isFieldParameter) {
+      // Note: it is important to set Vp_bc to Vp at Dirichlet dof nodes even if Vp_bc was already initialized with Vp
+      // because other boundary conditions applied before this one could have zeroed it out (if the bcs are applied to the same dofs).
+      Teuchos::ArrayRCP<Teuchos::ArrayRCP<const ST>> Vp_const2dView = Albany::getLocalData(dirichletWorkset.Vp);
+      for (unsigned int inode = 0; inode < nsNodes.size(); inode++) {
+        int lunk = nsNodes[inode][this->offset];
+        for (int col=0; col<num_cols; ++col) {
+          Vp_nonconst2dView[col][lunk] = Vp_const2dView[col][lunk];
+        }
+      }
+    } else {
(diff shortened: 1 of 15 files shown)


src/evaluators/bc/PHAL_DirichletField_Def.hpp
```diff
@@ -257,18 +289,10 @@ evaluateFields(typename Traits::EvalData dirichletWorkset) {
         int lunk = nsNodes[inode][this->offset];
         GO node_gid = nsNodesGIDs[inode];
         int lfield = fieldDofManager.getLocalDOF(field_node_indexer->getLocalElement(node_gid),fieldOffset);
-        for (int col=0; col<num_cols; ++col) {
+	      for (int col=0; col<num_cols; ++col) {
```
kliegeois, line 292: Alignment looks off here.

src/evaluators/bc/PHAL_SDirichlet_Def.hpp
```diff
@@ -280,45 +280,52 @@ SDirichlet<PHAL::AlbanyTraits::DistParamDeriv, Traits>::SDirichlet(
...
 
     for (unsigned int inode = 0; inode < nsNodes.size(); inode++) {
       int lunk = nsNodes[inode][this->offset];
 
-      for (int col = 0; col < num_cols; ++col) {
-        //(*Vp)[col][lunk] = 0.0;
+      for (int col=0; col<num_cols; ++col) {
         Vp_nonconst2dView[col][lunk] = 0.0;
-      }
+       }
```
kliegeois, line 304: The alignment looks off here.

src/evaluators/bc/PHAL_SDirichletField_Def.hpp
```diff
@@ -356,37 +356,49 @@ SDirichletField(Teuchos::ParameterList& p) :
 // **********************************************************************
 template<typename Traits>
 void SDirichletField<PHAL::AlbanyTraits::DistParamDeriv, Traits>::
-evaluateFields(typename Traits::EvalData dirichlet_workset) {
+preEvaluate(typename Traits::EvalData dirichletWorkset) {
 
-  bool isFieldParameter =  dirichlet_workset.dist_param_deriv_name == this->field_name;
-  TEUCHOS_TEST_FOR_EXCEPTION(
+  bool isFieldParameter =  dirichletWorkset.dist_param_deriv_name == this->field_name;
+    TEUCHOS_TEST_FOR_EXCEPTION(
```
kliegeois, line 362: The alignment looks off here.

kliegeois (commented):
This globally looks good to me. I just have 3 minor comments.

src/evaluators/bc/PHAL_Dirichlet_Def.hpp
```diff
@@ -260,39 +260,50 @@ Dirichlet(Teuchos::ParameterList& p) :
...
     for (unsigned int inode = 0; inode < nsNodes.size(); inode++) {
       int lunk = nsNodes[inode][this->offset];
 
       for (int col=0; col<num_cols; ++col) {
-        //(*Vp)[col][lunk] = 0.0;
         Vp_nonconst2dView[col][lunk] = 0.0;
        }
     }
-  } else {
+  }
```
bartgol, line 285: Why is there no `else` branch anymore? I don't really understand what `trans` really means, to be fair...

mperego: I removed it because this step is already done in the pre-evaluate phase for the transpose case.
In the transpose case, the DistParamDeriv evaluation computes  $(\partial_p f)^t \lambda$ whereas $(\partial_p f) q$ is computed in the non-transpose case. Here $f$ is the residual, $p$ is the parameter, $\lambda$ is a vector of the same size as the solution (typically the Lagrange multiplier) and $q$ is a vector having the same size of the distributed parameter $p$.

src/evaluators/bc/PHAL_DirichletField_Def.hpp
mperego: thanks, fixed this and the other indentation errors.

PR #885, opened 2023-01-11
Minimize stuff to rebuild when config options change

bartgol:
Most importantly, I removed the DEMO_PDES and LANDICE macros from Albany_config.h, and added them as CPP defs to the executables. They are only needed when creating the problem factory, which is done only in the two main.cpp files (the mpas interface does use pb factories, but sets it up manually, without checking any macro). This allows one to disable, say, LandIce and/or DemoPdes while developing code, without causing the whole Albany to be rebuilt.

I also reorganized some CMake code, moving it out of the main CMakeLists.txt file, which was imho getting too big. There's no need to put LandIce options in there, we can declare/handle them from inside the landice folder.

I still have to fully test this, and it might take a couple of days till I can. I did it while I waited for the code on another branch to compile. :)


mperego (approved):
Looks good, thanks!

bartgol:
I did a minor change, so no need for re-review. It builds fine on my workstation. Merging.

PR #888, opened 2023-01-14
Remove Tetra as an option for Extruded meshes

bartgol:
This PR removes Tetra as an option for Extruded meshes. Hence, if the basal element shape is Triangle, the volume element will always be a Wedge.

This is somewhat needed in the DOF manager PR, where it has already been assumed. Having 2+ elements in each layer makes retrieval of dofs a bit more complicated.

I ran all the tests (except pyAlbany), and got some failures in FO_GIS tests, with responses/sensitivities being a bit off. I tried to increase quadrature degree and make NOX tolerance stricted, but it only improves a bit. I suspect that our current tolerances were not tight enough to make the results "converged", so there's no hope to get a match. I'm going to verify this next week, by taking a test in master, and bumping up quadrature degree while making NOX tolerances stricter. This way, I should get an "exact" response/sensitivity from master, which should match the ones I get in this branch.

@mperego I am assuming that, provided that the quadrature/tolerances are strict enough, all the responses at the top/bottom with Tets (master) should match the values with Wedges (branch). I feel like the FE approx on the top/bottom should be the same for Tets/Wedges, but I am not expert enough in what Intrepid2 does with Wedges. Is my assumption correct? If not, then it's possible we cannot get an exact match with current master, no matter how strict the convergence is.


mperego:
@bartgol I don't think we can expect that using wedges will produce the same results as using tets. The discretizations are different (piecewise linear for tets, piecewise bilinear for wedges), so they would converge to the same solution only as the mesh size goes to zero. If you do not see large differences between the tets and wedges solution, I would just re-bless the tests.

bartgol:
> @bartgol I don't think we can expect that using wedges will produce the same results as using tets. The discretizations are different (piecewise linear for tets, piecewise bilinear for wedges), so they would converge to the same solution only as the mesh size goes to zero. If you do not see large differences between the tets and wedges solution, I would just re-bless the tests.

The differences are all like the following:
```
Response Test 0: 1.098837995820e+08 != 1.076679182380e+08 (rel 1.000000000000e-04 abs 1.000000000000e-04)
Sensitivity (0,0), Test 0,0: 1.826738903461e+07 != 1.867976080160e+07 (rel 1.000000000000e-04 abs 1.000000000000e-04)
---
Response Test 0: 1.089173687130e+08 != 1.066976120250e+08 (rel 1.000000000000e-06 abs 1.000000000000e-06)
Sensitivity (0,0), Test 0,0: 1.720834249957e+07 != 1.753719284188e+07 (rel 1.000000000000e-06 abs 1.000000000000e-06)
----
Response Test 0: 1.111601023583e+08 != 1.091294526860e+08 (rel 1.000000000000e-04 abs 1.000000000000e-04)
Sensitivity (0,0), Test 0,0: 1.849596738867e+07 != 1.882621076480e+07 (rel 1.000000000000e-04 abs 1.000000000000e-04)
---
Piro Analysis Test Two Norm: 9.69190977257 != 9.66706 (rel 1e-05 abs 1e-05)
---
Piro Analysis Test Two Norm: 90.6074319631 != 85.2913 (rel 1e-05 abs 1e-05)
```

They are all of the order of 2-3%, except for the analysis tests, which I assume are more sensitive to convergence, since there are two nonlinear solvers (NOX+ROL). If you think this kind of diffs could be attributed simply to the difference of discretizations, I can bless them.

mperego:
Yes, they look reasonable. I would bless them.

tests/landIce/FO_GIS/input_fo_gis_analysis_stiffening.yaml
```diff
@@ -381,10 +380,10 @@ ANONYMOUS:
         Status Test Check Type: Minimal
 
   Regression For Response 0:
-    Test Value: 6.14131581607e+01
+    Test Value: 6.068069886135e+01
     Relative Tolerance: 1.0e-05
     Absolute Tolerance: 1.0e-05
     Sensitivity For Parameter 0:
-      Test Value: 1.06024176845e+00
+      Test Value: 1.231269869500e+00
```
bartgol, line 387: @mperego This change is more than a few %. It's about 15%. The response itself is <2% diff though. Do you think it's fine?

tests/landIce/FO_GIS/input_fo_gis_beta_smb.yaml
```diff
@@ -256,11 +255,11 @@ ANONYMOUS:
...
     Sensitivity For Parameter 0:
-      Test Value: 1.107433043522e+01
-    Test Value: 6.436521674120e+01
+      Test Value: 1.088006785671e+01
+    Test Value: 6.555972610432e+01
     Relative Tolerance: 1.00000000000000004e-04
   Regression For Response 1:
     Absolute Tolerance: 1.00000000000000004e-04
-    Test Value: 3.907406428441e+03
+    Test Value: 6.422658118363e+03
```
bartgol, line 263: @mperego This is very different though. It's almost a 2x. Is it worrisome?

tests/landIce/FO_GIS/input_fo_gis_analysis_stiffening.yaml
mperego: I think it's fine. These GIS meshes are super coarse.

tests/landIce/FO_GIS/input_fo_gis_analysis_stiffening_memT.yaml
```diff
@@ -73,9 +73,8 @@ ANONYMOUS:
     Workset Size: -1
     Method: Extruded
     Number Of Time Derivatives: 0
-    Cubature Degree: 1
+    Cubature Degree: 3
```
mperego, line 76: We typically use Cubature Degree: 4 for Wedges. I found that sometime with degree 3 it takes more iterations to converge.

tests/landIce/FO_GIS/input_fo_gis_beta_smb.yaml
mperego: This is a lot, at the same time I think that the flux div evaluator, used for the SMB response has been tested more for  wedges than for tets. Maybe you can check if increasing the cubature rule makes any difference.

src/disc/Albany_DiscretizationFactory.cpp
```diff
@@ -118,10 +118,7 @@ DiscretizationFactory::createMeshStruct(Teuchos::RCP<Teuchos::ParameterList> dis
           basal_params->set("Workset Size", -1);
         } else if (!basal_params->isParameter("Workset Size")) {
           // Compute basal workset size based on extruded workset size
-          const int num_cells_per_column_per_layer = disc_params->get<std::string>("Element Shape") ==
-              "Tetrahedron" ? 3 : 1; // number of 3D cells in one extruded 2D cell
-          const int num_cells_per_column = num_cells_per_column_per_layer * disc_params->get<int>("NumLayers");
-          int basal_ws_size = extruded_ws_size / num_cells_per_column;
+          int basal_ws_size = extruded_ws_size / disc_params->get<int>("NumLayers");
```
mperego, line 121: This might create issues with the MPAS mesh struct. Need to check this branch with MPAS before you merge.

tests/landIce/FO_GIS/input_fo_gis_analysis_stiffening_memT.yaml
bartgol: Ok, I can sed/replace 4 everywhere. I found 3 somewhere, so I used that, but I can just use 4 everywhere. What about the lateralside quadrature order? With tets, we used 3. Is 4 enought?

mperego: I think 4 is enough. The reason why we need much higher quadrature for prism than tets for the volume integration is that the viscosity depends  (nonlinearly) only on the grad of the velocity, which is piece-wise constant for Tets so 1 quad point is enough.

bartgol: Well, the lateralside BC also has the immersedRatio quantity, which is something like `z_s/H`, which is also nonlinear.

mperego: agreed, but is more or less equally bad both for triangles and quads

bartgol: Fixed!

mperego:
I created a [PR](https://github.com/MPAS-Dev/compass/pull/510) in Compass to switch the tests using Tets to use Wedges. 
@jewatkins when are you planning to do the same with the performance tests?  
Note that when you switch to wedges, you need to set the `Cubature Degree `in the `Discretization` list to 4 (or more) and you'll have to fix the regression values.

jewatkins:
> I created a [PR](https://github.com/MPAS-Dev/compass/pull/510) in Compass to switch the tests using Tets to use Wedges. @jewatkins when are you planning to do the same with the performance tests? Note that when you switch to wedges, you need to set the `Cubature Degree `in the `Discretization` list to 4 (or more) and you'll have to fix the regression values.

I think I need to wait until blake is back up. blake has more tests than weaver so I can't get the new regression values without blake. If this is a blocker, I could try setting up performance tests on attaway and get the new values there? We should probably do that anyways if we want to run large simulations there.

bartgol:
Can't we take care of blake performance tests once we get it back? We're pretty convinced to switch to Wedge anyways, so I think we can adjust the performance regression tests later.

jewatkins:
okay maybe what I'll go ahead and do is update the weaver performance tests to use wedge once nightly albany is building again. Then this should be safe to push. Once blake is back up, I'll do the same for the rest of the tests.

mperego (approved):
I think we can merge this. Both Compass and Ali_perf_tests moved to Wedges.

bartgol:
Running one last round of tests on mappy, to ensure the conflicts resolution during master merge didn't change things.

jewatkins:
FYI there's been a couple of pretty large regressions in performance on blake. One around 10/21 and another while blake was down (12/12 was the last point). I changed the test names to no longer have "tet" or "wdg" so it won't be captured in future plots (plots are constructed based on test name). But I saved the old plot. I'll try to investigate a bit.

bartgol:
All tests pass on mappy. Merging.

Issue #889, opened 2023-01-18
Add config option to disable all output to file


bartgol:
On some machines, filesystem and/or TPL issues might make writing to file very slow. E.g., on my desktop, right now, I'm getting timers summary containing stuff like
```
 |   |   Albany: Output to File: 21.6071 - 83.6868% [161] {min=21.0195, max=22.0719, std dev=0.444008}
```
which is insane. It makes my tests run ~5x slower, just to write the solution to the mesh. Avoiding to write to file would dramatically increase my turnaround time on some tests.

During development, we often only care about the actual numbers of the solution field(s). Of course, if working on disc infrastructure, we should prob make sure we can write solutions to file correctly, but in general we don't care about output files. Besides, we don't even _check_ the content of output files in our testing.

I am not sure if this is already possible, but I suspect not.

ikalash:
I think if you remove "Exodus Output File" from the .yaml file, it will not do output to Exodus file. 

I kind of like having the Exodus output files dumped by the nightlies b/c I sometimes look at them when I'm trying to do something in Albany, but that does not happen that often these days.

ikalash:
Oh ok, I see.  I don't think Albany has that option currently.

PR #891, opened 2023-01-18
Allow disabling of output mesh files

bartgol:
Two things.

1. Add `ALBANY_DISABLE_OUTPUT_MESH` to avoid any output mesh write. It's OFF by default. Addresses #889. If ON, all `writeMatrixMarket` calls will still run (so one can use them during debugging cycles). Also, if ON, all restart tests are disabled.
2. Disable a few relatively large tests (>100s) if `ENABLE_LARGE_TESTS=OFF`.


tests/CMakeLists.txt
```diff
@@ -4,6 +4,62 @@
 ##    in the file "license.txt" in the top-level Albany directory  //
 ##*****************************************************************//
 
+MESSAGE ("\n ***********************")
+MESSAGE (" *** Testing Options ***")
+MESSAGE (" ***********************\n")
+
+# Add option to enable performance tests. Results are machine-specific
+OPTION(ENABLE_PERFORMANCE_TESTS "Flag to turn on Performance Tests" OFF)
```
mperego, line 12: I don't think this is used any longer. We moved the performance tests in another repo.

CMakeLists.txt
```diff
@@ -926,39 +860,29 @@ IF(ALBANY_MPI)
...
-  ENDIF()
-  IF (ALBANY_PYTHON)
-    SET(ALBANY_PYTHON_TESTS FALSE)
-  ENDIF()
-ELSE()
+#Add option to disable all tests 
+OPTION(DISABLE_ALBANY_TESTS "Flag to disable tests" OFF)
+IF (DISABLE_ALBANY_TESTS)
+  MESSAGE("-- DISABLE_ALBANY_TESTS is ON.  Tests will not be built.")
+elseif (ALBANY_LIBRARIES_ONLY)
```
mperego, line 867: Are `DISABLE_ALBANY_TESTS` and `ALBANY_LIBRARY_ONLY` still used? The purpos of the latter was to avoid building the `Albany` and `AlbanyAnalysis` executables.

mperego (approved):
Looks good to me. I've only a couple of comments.

CMakeLists.txt
bartgol: Yeah, they seem very correlated... If you disable tests, why would you build the executable? I can remove one

bartgol:
All comments addressed. Merging.

PR #894, opened 2023-01-23
Get rid of AbstractFieldContainer and the FieldRequirements concept

### LINKED ISSUE #892: Can we get rid of FieldRequirements?
These vectors are set in the problem, and passed to the discretization...but they are never used! They are ultimately passed to the field container structures, which simply ignore them (the MultiSTKFieldContainer doesn't even take it as a parameter).

They were originally supposed (I think) to serve as a way to check that all required fields were in the mesh, but it seems like a lot of infrastructure just for checking that. Besides, if a field is missing, we will get an error from the DAG check, since the LoadStateField (or similar) evaluator was never created.

This is not urgent, but there are hundreds of lines of code that can be removed. Our disc code is already complex enough to maintain, no need to add extra (and pointless) headaches.

Edit: they are still used in the GenericSTKMeshStruct, as a way of checking that all required fields are loaded from the mesh. But, again, this check is probably not needed (DAG covers it), and extremely heavy in terms of LOC to maintain.

bartgol:
The requirements were meant to signal fields that must be present in the mesh. However, the where not specifying anything about the layout, so in itself it was an incomplete piece of information. A while ago, we added the `Required Fields Info` sublist in the discretization, which contains all information for input/output fields from/to mesh. The problems are supposed to use this info, to decide what to load/save from/to the mesh, making the FieldRequirements concept obsolete.

This PR removes the AbstractFieldContainer class,whose only job was to provide the `FieldRequirements` typedef.

Fixes #892


mperego (commented):
Love this cleanup. Let me check with MALI first. It's probably good anyway to wait until the nightly tests are up again before merging this.

bartgol:
Sure, no rush. It doesn't impact my work anyways.

mperego:
@bartgol I tried building MALI against this branch and it works fine. Feel free to merge.

Issue #893, opened 2023-01-20
Build issues due to ALBANY_MPI flag being undefined


mcarlson801:
(This is unrelated to [Trilinos Issue #11487](https://github.com/trilinos/Trilinos/issues/11487) which is the culprit of current failures on cdash. This issue arises after implementing a workaround for the Trilinos issue.)

I'm currently seeing the following error when trying to build Albany:

```/usr/bin/ld: ../../../src/libalbanyLib.so: undefined reference to `Albany::getMpiCommFromTeuchosComm(Teuchos::RCP<Teuchos::Comm<int> const>&)'
collect2: error: ld returned 1 exit status
make[2]: *** [tests/unit/evaluators/CMakeFiles/gatherDistributedParameters_unit_tester.dir/build.make:227: tests/unit/evaluators/gatherDistributedParameters_unit_tester] Error 1
make[1]: *** [CMakeFiles/Makefile2:2073: tests/unit/evaluators/CMakeFiles/gatherDistributedParameters_unit_tester.dir/all] Error 2```

After digging into `Albany_CommUtils.cpp`, it looks like this is undefined because somehow the `ALBANY_MPI` flag is undefined. I took out the check for `ALBANY_MPI` and this built without issue. Anyone have any ideas why this might not be getting defined? I tried defining it as a cmake flag with no success.

Also, @bartgol, @jewatkins, is there ever a case where Albany gets built without MPI enabled? I wonder if we even need some of the logic based on `ALBANY_MPI`.

**Comment by bartgol** [comment]
I think the non MPI case was introduced long time ago. IMHO, we could remove it. Grepping in trilinos, it's already a required TPL for some of the packages we use:
```
$ git grep -w MPI | grep REQUIRED
cmake/tribits/examples/MockTrilinos/packages/panzer/cmake/Dependencies.cmake:  LIB_REQUIRED_TPLS MPI Boost
packages/panzer/disc-fe/cmake/Dependencies.cmake:SET(LIB_REQUIRED_DEP_TPLS MPI)
packages/panzer/dof-mgr/cmake/Dependencies.cmake:SET(LIB_REQUIRED_DEP_TPLS MPI)
packages/panzer/mini-em/cmake/Dependencies.cmake:SET(LIB_REQUIRED_DEP_TPLS MPI)
packages/pike/blackbox/cmake/Dependencies.cmake:  LIB_REQUIRED_TPLS MPI
[...]

PR #895, opened 2023-01-23
Remove ALBANY_MPI flag and corresponding logic

mcarlson801:
As pointed out by Luca in #893, there is no longer any situation where we would build Albany without MPI. This PR removes the ALBANY_MPI flag and any logic surrounding it. This PR will also fix issue #893.


src/Albany_CommTypes.hpp
```diff
@@ -13,17 +13,10 @@
...
 
-  #define Albany_MPI_COMM_WORLD MPI_COMM_WORLD
-  #define Albany_MPI_COMM_NULL MPI_COMM_NULL
-#else
-  using Albany_MPI_Comm = int;
-
-  #define Albany_MPI_COMM_WORLD 0  // This is compatible with Dakota
-  #define Albany_MPI_COMM_NULL 99
-#endif
+#define Albany_MPI_COMM_WORLD MPI_COMM_WORLD
```
bartgol, line 18: These macros seem unused. I would prune them.

src/Albany_CommTypes.hpp
```diff
@@ -13,17 +13,10 @@
 // Get Teuchos Comm type
 #include "Teuchos_DefaultComm.hpp"
 
-#ifdef ALBANY_MPI
-  using Albany_MPI_Comm = MPI_Comm;
+using Albany_MPI_Comm = MPI_Comm;
```
bartgol, line 16: This is only used in CommUtils, and given that there's no conditional compilation anymore, I would prune it, replacing it with `MPI_Comm` in the comm utils files.

bartgol: The reason is clarity: a typedef suggests the type might be complex and/or dependent on other stuff, so the user might spend extra time grepping for it, to figure out what's the catch, and why this is not just MPI_Comm.

tests/CMakeLists.txt
```diff
@@ -106,35 +106,28 @@ ENDIF()
...
-    set(Albany8.exe                    ${MPIEX} ${MPIPRE} ${MPINPF} ${MAX_MPI_RANKS} ${MPIPOST} ${AlbanyPath})
-    set(AlbanyAnalysis.exe             ${PARALLEL_CALL} ${AlbanyAnalysisPath})
-  ENDIF() 
-ELSE()
-  set(SerialAlbany.exe                 ${AlbanyPath})
-  set(SerialAlbanyAnalysis.exe         ${AlbanyAnalysisPath})
-  set(Albany.exe                       ${AlbanyPath})
-  set(AlbanyAnalysis.exe               ${AlbanyAnalysisPath})
-ENDIF()
+set(SerialAlbany.exe                 ${SERIAL_CALL} ${AlbanyPath})
```
bartgol, line 109: Question: why is our serial call `mpirun -np 1 Albany` rather than just `Albany`? The mpirun does not add anything here, besides (at most) some freedom when it comes to binding the forked thread to a resource. Other than that, it is the same as running the raw executable itself...

It seems a bit of an overcomplication, with associated reading/parsing slowdown, and no tangible added value.

tests/CMakeLists.txt
```diff
@@ -106,35 +106,28 @@ ENDIF()
...
+    SET(MAX_MPI_RANKS ${MPIMNP})
+  ELSE(MPIMNP LESS 8)
+    SET(MAX_MPI_RANKS 8)
+  ENDIF(MPIMNP LESS 8)
+ELSE(DEFINED MPIMNP)
+    SET(MAX_MPI_RANKS 4)
+ENDIF(DEFINED MPIMNP)
+IF (ALBANY_ENABLE_CUDA) 
+  set(Albany.exe                     ${PARALLEL_CALL} ${AlbanyPath} ${KOKKOS_NDEVICES})
+  set(Albany8.exe                    ${MPIEX} ${MPIPRE} ${MPINPF} ${MAX_MPI_RANKS} ${MPIPOST} ${AlbanyPath} ${KOKKOS_NDEVICES})
```
bartgol, line 124: @mperego I see that Albany8 is only used for AIS tests and the 20km GIS test. What's the reason for this peculiarity? Why can't all test use the same number of ranks for the parallel call?

bartgol (approved):
Looks good, thanks for doing this! I have a suggestion for further MPI comm cleanup.

Also, I have a couple of questions on our CMake logic, which may or may not be included in this PR if we agree on further cleanup.

bartgol (approved):

jewatkins (approved):
Looks good, thanks for this!

ikalash:
Can we merge this in?

mperego:
Oh, yeah, go for it. hope @mcarlson801  was not waiting for me to approve.

Issue #873, opened 2022-12-20
SteadyHeatConstrainedOpt2D_Scalar_And_Dist_Param_Hessian_Tpetra failing


jewatkins:
[SteadyHeatConstrainedOpt2D_Scalar_And_Dist_Param_Hessian_Tpetra](https://sems-cdash-son.sandia.gov/cdash/test/3157534)
The Trilinos/Albany commit ids for the point of failure are:
Trilinos passed: cc8c8b2599d
Trilinos failed: cc8c8b2599d
Albany passed: 79eab1196
https://sems-cdash-son.sandia.gov/cdash/test/3152291?graph=status
Albany failed: 9f6fe5c88
https://sems-cdash-son.sandia.gov/cdash/test/3154462?graph=status
So no change to trilinos and I don't see any changes in Albany that could have caused this. @ikalash made some changes but it looks like it was just testing scripts not related to camobap. Did you change openmp num threads here? How many threads are running currently?

I don't see anything obvious here. It looks like the case produces a nan. @mperego mentioned that it might related to one of his changes a while ago so maybe it's a race condition that just decided to pop up now.

ikalash:
No, I did not change # threads.  Here is what I'm using:

ikalash:
export OMP_NUM_THREADS=2
export OMP_PLACES=threads
export OMP_PROC_BIND=spread

Based on the dates, I don't think the error is from my changes.  I finished reworking the camobap nightlies on 12/12, based on git log.

jewatkins:
It looks like this is failing on the serial intel build now too? https://sems-cdash-son.sandia.gov/cdash/test/3158793 so not a race condition. very odd

jewatkins:
And it passed on the openmp build: https://sems-cdash-son.sandia.gov/cdash/test/3158944?graph=status so it seems sensitive.

mperego:
I'll look into that. It's good it's failing on a serial build

ikalash:
Just to clarify, this is in an OpenMP build (KokkosNode = OpenMP).

jewatkins:
@ikalash I think this one is a serial build as it's labeled as serial? https://sems-cdash-son.sandia.gov/cdash/test/3158793

mperego:
But this is serial
 https://sems-cdash-son.sandia.gov/cdash/test/3158793
 It started failing today

ikalash:
[...]

PR #899, opened 2023-01-25
Fix derivative size for scalar parameters

mperego:
This PR should fix randomly failing tests reported in issue #873 due to a mismatch in derivative sizes for `Tangent` and `HessianVec` evaluations.
Also pruned some unused code and files.


src/Albany_Application.cpp
```diff
@@ -1996,8 +1993,15 @@ Application::computeGlobalTangent(
   if (Teuchos::nonnull(params)) {
     TanFadType p;
     int        num_cols_tot = param_offset + num_cols_p;
```
bartgol, line 1995: So this, more than a number of columns, now represent the upper bound for the derivatives for this parameter, right? That is, the derivs for this parameter will be at entries `[param_offset,num_cols_tot)` in the Fad, which will _always_ have length equal to the total number of parameters. Correct?

src/Albany_Application.cpp
```diff
@@ -1996,8 +1993,15 @@ Application::computeGlobalTangent(
...
+
+    TEUCHOS_TEST_FOR_EXCEPTION(
+      num_cols_tot >  tangent_deriv_dim,
```
bartgol, line 1998: Actually, I now see that `param_offset` is not the offset of the current parameter, but rather 0 or num_cols_x, depending on `sum_derivs`. However, I wonder how this can work if `sum_derivs=false`, since `calcTangentDerivDimension` (which computes `tangent_deriv_dim`)  does not seem to consider this. So I am afraid that, if `sum_derivs=false`, we would trigger this exception.

bartgol (commented):
It looks good. I only have one doubt regarding `sum_derivs` (which, to be honest I don't fully understand)

src/Albany_Application.cpp
bartgol: It actually seems that `sum_derivs=false` always. So now I'm confused.

mperego: I think `offset` is always zero because `sum_derivs` it's always `false`, we should clean this up at some point. so `num_cols_tot` should always be less or equal to `tangent_deriv_dim`.

src/Albany_Application.cpp
mperego: yes, I think so.

src/Albany_Application.cpp
bartgol: If `sum_derivs=false`, then `param_offset=num_cols_x`. I guess we don't really ever use Vx/Vxdot/Vxdotdot then, otherwise the offset would be nonzero...

bartgol: Btw, if `sum_derivs=true`, then we need to ensure that the fad length is >=max(num_cols_p, num_cols_x).

I agree, this needs some cleanup at some point. I'm very wary of code that we may be not using, and might be super buggy.

mperego: Oh right sorry, yes you are right. If I remember correctly Vx/Vxdot/Vxdotdot  are needed for some software optimization, but we never use them and we probably should clean up the code at some point, maybe after your PR.

src/Albany_Application.cpp
```diff
@@ -2954,11 +2966,19 @@ Application::evaluateResidual_HessVecProd_pp(
...
+            "\nError in Albany::Application::evaluateResidual_HessVecProd_pp  "
+            << "Number of parameters columns cannot be larger than the derivative size\n ");
+
     HessianVecFad p_val;
     for (unsigned int i = 0; i < num_cols_p_l1; i++) {
-      p_val = HessianVecFad(num_cols_p_l1, params_l1[i].baseValue);
+      p_val = HessianVecFad(deriv_size, params_l1[i].baseValue);
       p_val.fastAccessDx(i).val() = 1.0;
-      params_l1[i].family->setValue<PHAL::AlbanyTraits::HessianVec>(p_val);
+      params_l1[i].family->setValue<EvalT>(p_val);
```
kliegeois, line 2981: Why did you change this line?
If this is not needed, I would be in favor of keeping it as before or to change the other lines such as :

https://github.com/sandialabs/Albany/blob/d89979ca25d07aa7007b938f76742f3968c56883/src/responses/Albany_FieldManagerScalarResponseFunction.cpp#L532
and
https://github.com/sandialabs/Albany/blob/d89979ca25d07aa7007b938f76742f3968c56883/src/responses/Albany_FieldManagerScalarResponseFunction.cpp#L623
.

If those lines are consistent, it is easier to modify them all together if needed in the future.

kliegeois (commented):
It globally look excellent to me; I just have a minor comment above.

src/Albany_Application.cpp
mperego: I can change it back, it's just that you have
`using EvalT =  PHAL::AlbanyTraits::HessianVec`
 at the beginning of the function and I thought it would make sense to use it.

bartgol: I'm in favor of using `EvalT` (shorter lines when you multiple terminal windows). But I agree that it's best if it's consistent. So I vote for using the typedef everywhere that it's possible. :)

kliegeois: I +1 Luca's comment.

mperego: OK, I did that. Merging...

PR #900, opened 2023-01-26
Remove SFAD4 and SFAD8 builds

jewatkins:
These builds are no longer needed since we are no longer supporting tets for LandIce (see https://github.com/sandialabs/Albany/pull/888 and https://github.com/sandialabs/ali-perf-tests/pull/6).


mperego (approved):
Lgtm

jewatkins:
I also updated the files in the blake/weaver project space so these builds should no longer exist tomorrow.

PR #903, opened 2023-01-31
Removed three unused LandIce evaluators

bartgol:
`Viscosity` is a leftover from the Stokes cleanup. `StackFields` is something I thought I could use when I was fiddling with sideset eqns, but then abandoned it. I'm not 100% sure if MapThickness will be needed at some point, but as of now it is unused.

Note: still need to run tests.


mperego (approved):
Lgtm.  Thanks, sorry I missed them.

bartgol:
All tests pass on mappy. Merging.

PR #909, opened 2023-02-03
LandIce: reduce execution time of a couple of tests.

mperego:
Trying to avoid timeout in debug builds

### DIFF
--- a/tests/landIce/FO_GIS/input_fo_gis_analysis_stiffening_memT.yaml
+++ b/tests/landIce/FO_GIS/input_fo_gis_analysis_stiffening_memT.yaml
@@ -75,7 +75,7 @@ ANONYMOUS:
     Cubature Degree: 4
     Exodus Output File Name: gis_analysis_stiffening_mem.exo
     Columnwise Ordering: true
-    NumLayers: 5
+    NumLayers: 3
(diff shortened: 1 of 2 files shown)


tests/landIce/FO_GIS/input_fo_gis_analysis_stiffening_memT.yaml
```diff
@@ -75,7 +75,7 @@ ANONYMOUS:
     Cubature Degree: 4
     Exodus Output File Name: gis_analysis_stiffening_mem.exo
     Columnwise Ordering: true
-    NumLayers: 5
+    NumLayers: 3
```
jewatkins, line 78: I don't think the memoized case times out if you want to keep this one the same. It looks like this will be different than the unmemoized case anyways?

mperego: I know.. but it's probably good to make it faster anyway. As you prefer.

jewatkins: Okay sounds good, we can keep it as is.

jewatkins (approved):

PR #912, opened 2023-02-09
Fix fad offsets when dealing with column-coupled equations

bartgol:
There were a few inconsistencies in how we dealt with column couplings. In particular, in some places we offset fad entries for column coupling by `neq*numCellNodes`, while in others we didn't. But then, after adding the offset to all gather/scatter places, I realized we don't need it, since we never use simultaneously the "volume" fads (the first `neq*numCellNodes` entries) and the "column" fads (the following `neq*numSideNodes*numLevels` entries), so we can reduce the fad length of column-coupled responses to `neq*max(numCellNodes,numSideNodes*numLevels)`.

I spent some time thinking about whether I was going to mix up things, but I think I'm not. SO LONG as the 2D solution field is gathered via `Gather2DField`, then all Fad use the convention that 2d quantities Fads are located at Fad entry `numSideNodes*field2dLevel + $node2d*neq + eq`.

For a 10 layers mesh with wedges, for the FOThicknes sproblem, this brings the Fad length from `3*6+3*11*3=117` to `3*11*3=99`. Not a huge gain, but still a 20% memory reduction for the response FM.

With this mod, I get sensitivities for the FO_GIS_Coupled test that are close to what I get in the DOFManager branch (diff ~ 3e5). I am hoping the diff can now be made smaller by simply cranking down the solvers tolerances (I will check).

Note: I still have to push changes to the input files. I need to run all tests first, to see what changes.

### DIFF
--- a/src/PHAL_Utilities.cpp
+++ b/src/PHAL_Utilities.cpp
@@ -59,7 +59,8 @@ template<> int getDerivativeDimensions<PHAL::AlbanyTraits::Jacobian> (
         int side_node_count = ms->ctd.side[3].topology->node_count;
         int node_count = ms->ctd.node_count;
         int numLevels = app->getDiscretization()->getLayeredMeshNumbering()->numLayers+1;
-        dDims = app->getNumEquations()*(node_count + side_node_count*numLevels);
+        int neq = app->getNumEquations();
+        dDims = neq*std::max(node_count, side_node_count*numLevels);
       }
   }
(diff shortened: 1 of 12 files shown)


bartgol:
On mappy, I get all pass except for the CoupledThickness tests, which are the only ones that really use the column coupling evaluators. I will update the regression values in the input file, and push a new version. Then, this can be merged.

mperego (approved):

Issue #933, opened 2023-04-15
Albany link errors having to do with blas/mkl with Intel compiler


ikalash:
Starting on 4/11, Albany stopped building on CEE and attaway with an intel compiler: 

https://sems-cdash-son.sandia.gov/cdash/viewBuildError.php?buildid=47333
https://sems-cdash-son.sandia.gov/cdash/viewBuildError.php?buildid=47330

There are link errors having to do with mkl and blas libraries.  I originally thought the problem was in Trilinos but it appears it's not (see https://github.com/trilinos/Trilinos/issues/11788 and discussion therein).  @jewatkins it would be great if you could have a look at this.  My attempts to fix / triage it have been unsuccessful and I won't have more time to devote to this before my upcoming trip.

jewatkins:
So I noticed you swapped boost lib: https://github.com/sandialabs/Albany/commit/dcb2dd380c0779ac7006d335e5ca3d2669275ccd I actually dropped the need for boost libraries in my build (we don't need them, just the include dir) and I forgot to push.

I'm going to try this build myself right now.

jewatkins:
I get the failure too. So looking at the commit history, this issue coincides with Luca's big PR: https://github.com/sandialabs/Albany/commit/3ded89bf8043c5b11b041a7e38e44753a06d3fd6. @bartgol any reason to believe your changes may have stirred up some intel linking issues? I'll keep looking.

jewatkins:
Ah I bet it's this: https://github.com/sandialabs/Albany/commit/3ded89bf8043c5b11b041a7e38e44753a06d3fd6#diff-1e7de1ae2d059d21e1dd75d5812d5a34b0222cef273b7c3a2af62eb747f9d20aL192

Let me try adding `-mkl`

bartgol:
I removed that line. In the spirit of CMake, I don't think grabbing a var containing all flags, and stuffing them inside Albany, is the way to go.

My driving motivation for rm-ing it was that I wanted to build albany in DEBUG mode, but using trilinos in RELEASE mode (or else compiling takes too long). Since the DEBUG build does not add `-O0` in CMake (by default), the last opt flag came from trilinos flags, so even a DEBUG build would end up being RELEASE, when linking against Trilinos.

Notice that all important flags _should_ be attached to _some_ trilinos target, so we should get them (provided they are in the PUBLIC interface of the tgt). The only flags that CMake does not attach to the target are the debug/opt flags.

[...]

PR #938, opened 2023-04-20
Fix attaway build

jewatkins:
@ikalash I'm reverting the attaway build to add `-mkl` because the other approach didn't work (maybe an old mkl?). https://github.com/sandialabs/Albany/issues/933

Could you point me to where these files need to be copied?


bartgol:
It would be helpful if the build scripts used `-DCMAKE_VERBOSE_MAKEFILE=ON`, so we can see the compile line. Alternatively, can we make the build dir have 755 permissions, so we can go in and manually check the flags.txt and link.txt files?

jewatkins:
I agree. These directories should be group readable like on blake/weaver. And executable so everyone can use the final binary.

ikalash:
Sorry about the permissions issue.  THe problem was that permissions were not auto reset to wg-albany after the nightly build, like they are on say the CEE machines.  THe problem has been corrected.  I can add this option as well ```-DCMAKE_VERBOSE_MAKEFILE=ON```.

ikalash:
I just updated the files based on what was in the repo.  THe place the nightlies happen is /projects/albany/nightlyCDash.

jewatkins:
I just merged the fix. The files likely weren't part of the copy. I can do that tomorrow once permissions are autofixed.

ikalash:
I changed the perms manually so you should be able to copy the files.

jewatkins:
Can you try one more time?
`rwx------ 4 ikalash ikalash 4096 Apr 21 14:21 nightlyCDash`
Thanks!

ikalash:
Looks like now there is an issue having to do with Amesos2/superlu: https://sems-cdash-son.sandia.gov/cdash//viewBuildError.php?buildid=47660 .

ikalash:
Also, sorry about the perms, they have been corrected.

jewatkins:
The new files weren't in. I just copied them over so we should be good tomorrow.

jewatkins:
Forget to set group permissions... hopefully tomorrow this is working.

Issue #942, opened 2023-04-28
PyAlbany tests failing on blake-gcc


jewatkins:
I fixed the mpi issue with pyalbany on blake-gcc but I'm now getting the following error:
```
Traceback (most recent call last):
  File "thermal_steady.py", line 44, in test_all
    theta_star, I_star, F_star, P_star = ee.evaluateThetaStar(l, problem, n_params)
  File "/ascldap/users/jwatkin/LandIce/Albany/build-blake-gcc-serial/pyAlbany/PyAlbany/ExtremeEvent.py", line 511, in evaluateThetaStar
    error = problem.performAnalysis()
RuntimeError: /ascldap/users/jwatkin/LandIce/Albany/src/utility/Albany_ThyraUtils.cpp:996:
Throw number = 1
Throw test that evaluated to true: true
Error in getLocalData! Could not cast Thyra_Vector to any of the supported concrete types.
```
I suspect pyAlbany hasn't been tested with the DOFManager changes in #861 since blake-gcc has been down since Feb. and it's the only build that tests it. Also it looks like maybe there was some thyra changes in that PR. @bartgol @kliegeois Any ideas what the issue could be? The nightlies should produce the same error tomorrow, I will link it then.

kliegeois:
Oh yes, this is most likely due to the PR that I merged today! I forgot to update PyAlbany. I will fix this tomorrow.

kliegeois:
I reopened the issue to track its status tomorrow or later this week.
This should be fixed but let's wait for the NB.

kliegeois:
Those tests passed last night so I am closing this issue.

Issue #943, opened 2023-04-28
Weaver SolverFactory error


jewatkins:
Error on weaver related to solver factory:
```
/.../repos/Albany/src/Albany_SolverFactory.cpp(199): error: expected a type specifier
```
@kliegeois Possibly related to #939?

kliegeois:
This should be fixed; let's wait for the NB results.

kliegeois:
The error is still present; I will investigate this further as soon as I can.

ikalash:
Seems this happened last night again.  Thanks for looking at it, @kliegeois !

kliegeois:
The compilation should pass correctly now.
I am currently testing the unit tests on weaver.

kliegeois:
The tests passed.

kliegeois:
The nightly builds associated to this issue passed.

PR #945, opened 2023-05-02
Fix PyAlbany tests 

kliegeois:
This PR updates both Albany and PyAlbany to fix https://github.com/sandialabs/Albany/issues/942 .
I tried this on Blake and on my workstation.


kliegeois:
I added an extra commit to fix #943 too.

bartgol (approved):

PR #946, opened 2023-05-06
Resolve warnings to build warning free

kliegeois:
This PR fixes the warnings treated as errors introduced since the Blake warning free build as been down.

The tests passed on my workstation.


ikalash:
Thanks for doing this.  I'm fine to merge it in.  We can discuss Tues. at the Albany meeting if we want to convert one of the existing Albany tests into a warning-free so that the capability doesn't get broken again.

bartgol (approved):

PR #949, opened 2023-05-08
Cleaning, removing code/text related to unused Trilinos packages

mperego:
Trilinos packages Rythmos, TriKota and Intrepid are no longer used in Albany and will be removed from Trilinos soon. MiniTensor is no longer used. In this PR we remove code, text related to those packages. In particular, we remove cmake options from cmake scripts enabling those packages. Also removed some old cmake scripts.


doc/developersGuide/gettingStarted.tex
```diff
@@ -185,14 +185,14 @@ \section{Distinguishing Capabilities}
 propogations get automatically computed with no development
 effort. This approach uses Phalanx for rapid and flexible addition of
 physics, which works closely with Sacado and Stokhos for automatic
-propagation of derivatives and UQ. Intrepid and Shards packages are
+propagation of derivatives and UQ. Intrepid2and Shards packages are
```
bartgol, line 188: If you really want to: missing space.

bartgol (approved):
I love celanup prs. Looks good.

doc/developersGuide/gettingStarted.tex
mperego: Actually, this is the tex of a Sandia report, so we should not modify this, I reverted the changes to this file

bartgol: If it's outdated, then we should either remove it or fix it no? It seems confusing to find instructions that are old...

mperego: Yeah, but it would take some time to fix it properly. I'm OK removing it, or we can put it under `doc/legacy/developersGuide`.

bartgol: True, that document is very long. I quickly scrolled it and saw a lot of old information (all the other physics packages, trilinos dependencies, MP/stokhos capabilities, ...). I would vote to remove it. As a new user or person intereted in the code, I would find it more confusing if there was a document describing something that is not there; I'd much rather not find any document than a confusing one.

mperego: OK, I removed it.

PR #950, opened 2023-05-08
Remove some unused stuff related to states

bartgol:
This is part of a bigger PR, but I figured splitting it up in smaller segments allows for faster review.

This PR removes some flavors of states registration in the StateManager that were not used anymore. The goal is to simplify what code needs to be fixed/maintained when subsequent modifications are made to states/state-manager.


kliegeois:
It looks like a cleanup day for Albany repo ! 😄

PR #952, opened 2023-05-09
Fix logic for creating adjoint model

mperego:
At the moment, we do not create an adjoint model when `Compute Sensitivities` is `false`. While this makes sense for forward runs, it does not for Analysis runs, where the needed derivatives are computed regardless of the `Compute Sensitivities` parameter.


kliegeois (approved):
I agree with the comment and the changes look good, thanks!

mperego:
@kliegeois Thanks!

PR #953, opened 2023-05-12
Work on cubature rules

mperego:
In this PR:

-   We move cubature degree from meshSpecs class to the abstract problem class. Test input files are modified accordingly, moving `Cubature Degree` parameter from the `Discretization` sublist to the `Problem` sublist. LandIce problems based on `StokesFOBase` class have two additional parameters in the `Problem` sublist: `Basal Cubature Degree `(for integrals on the ice bed) and  `Surface Cubature Degree` (for integrals on the ice surface).
- Check for inconsistencies (e.g. if we are specifying a cubature degree for the basal BC that is different than the one associated to the basal side)
- For FO Stokes problems, we introduce tensor-product cubature rules. In fact, for Extruded Meshes with Wedge or Hexaedral cells, it makes sense to use tensor product quadrature rules (tria x line or quad x line) with different degrees for the horizontal (tria or quad) and vertical (line) axes. The tensor-product rule is specified as 
`Cubature Degrees (Horiz Vert): [4,2]`.


src/disc/stk/Albany_GenericSTKMeshStruct.cpp
```diff
@@ -1545,7 +1545,10 @@ GenericSTKMeshStruct::getValidGenericSTKParameters(std::string listname) const
   validPL->set<int>("Exodus Write Interval", 3, "Step interval to write solution data to Exodus file");
   validPL->set<std::string>("Method", "",
     "The discretization method, parsed in the Discretization Factory");
-  validPL->set<int>("Cubature Degree", 3, "Integration order sent to Intrepid2");
+
+  //TODO: remove the following after updating MPAS
+  //validPL->set<int>("Cubature Degree", 3, "Integration order sent to Intrepid2");
```
bartgol, line 1550: So it can be removed already?

src/landIce/problems/LandIce_StokesFOBase.cpp
```diff
@@ -120,7 +120,30 @@ void StokesFOBase::buildProblem (Teuchos::ArrayRCP<Teuchos::RCP<Albany::MeshSpec
   cellType = rcp(new shards::CellTopology (cell_top));
 
   Intrepid2::DefaultCubatureFactory cubFactory;
-  cellCubature = cubFactory.create<PHX::Device, RealType, RealType>(*cellType, meshSpecs[0]->cubatureDegree);
+
+  int defaultCubDegree = 3;
+  bool tensorProductCell = (cellType->getKey() == shards::Wedge<6>::key) || (cellType->getKey() == shards::Hexahedron<8>::key);
```
bartgol, line 125: Technically, a hexa may not be an extruded mesh. We need the mesh to be extruded, since we need the elements to always be oriented in the same way. I think stokesFO has access to the disc param list, so we could ping that.. Alternatively, we could add a bool flag in the mesh specs (whether the mesh is extruded or not, is a true global info that deserves to be in the MS).

mperego: Oh right. We talked about this. I'll fix that.

src/landIce/problems/LandIce_StokesFOBase.cpp
```diff
@@ -143,13 +173,17 @@ void StokesFOBase::buildProblem (Teuchos::ArrayRCP<Teuchos::RCP<Albany::MeshSpec
...
 
-      // If there's no side discretiation, then sideMeshSpecs.cubatureDegree will be -1, and the user need to specify a cubature degree somewhere else
-      int sideCubDegree = sideMeshSpecs.cubatureDegree;
+      int sideCubDegree = (sideCubatureDegree.find(ssName) != sideCubatureDegree.end()) ? sideCubatureDegree[ssName] : -1;
       if (pl->isParameter("Cubature Degree")) {
-        sideCubDegree = pl->get<int>("Cubature Degree");
+        if(sideCubDegree == -1 )
+          sideCubDegree = pl->get<int>("Cubature Degree");
+        else {
+          TEUCHOS_TEST_FOR_EXCEPTION (sideCubDegree != pl->get<int>("Cubature Degree"), std::runtime_error,
```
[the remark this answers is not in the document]
mperego: It's possible but it's a bit of a pain, especially to fix all the MMS tests that have several sides. Also it's tricky to set the list of valid parameters.

src/disc/stk/Albany_GenericSTKMeshStruct.cpp
mperego: oh, I need to uncomment that for now.

src/landIce/problems/LandIce_StokesFOBase.cpp
```diff
@@ -131,6 +154,13 @@ void StokesFOBase::buildProblem (Teuchos::ArrayRCP<Teuchos::RCP<Albany::MeshSpec
   dl = rcp(new Albany::Layouts(worksetSize,numCellVertices,numCellNodes,numCellQPs,numDim,vecDimFO));
 
   unsigned int sideDim = numDim-1;
+  
+  std::map<std::string,int> sideCubatureDegree;
+  if (!isInvalid(surfaceSideName))
+    sideCubatureDegree[surfaceSideName] = params->isParameter("Surface Cubature Degree") ? params->get<int>("Surface Cubature Degree") : -1;
```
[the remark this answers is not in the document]
mperego: This goes back to whether we allow to set cubature degrees in the landice bcs. I tried to somewhat minimize the modifications to the test input files.

src/disc/stk/Albany_GenericSTKMeshStruct.cpp
mperego: Actually I removed this comment. I just checked and in all MPAS tests we rely on the default cubature degrees hard coded in the Interface

mperego:
@bartgol I committed a fix so that we now use tensor-product cubatures only for extruded meshes (checking the `Discretization` parameter list). Is that OK if I merge this and we postpone other changes? We probably need to revisit this anyway when we introduce blocks, as each block should probably have different cubature rules.

bartgol (approved):

PR #955, opened 2023-05-12
LandIce: ViscosityFO: Allow evaluating temperature at quadrature points

mperego:
Until now viscosity used P0 temperature fields.
Modified a couple of tests to exercise new capability.

Also, fixed Enthalpy problem so that it uses the pressure corrected temperature for computing the viscosity.

### DIFF
--- a/src/landIce/problems/LandIce_StokesFOBase.hpp
+++ b/src/landIce/problems/LandIce_StokesFOBase.hpp
@@ -254,6 +254,7 @@ protected:
 
   // Whether to use corrected temperature in the viscosity
   bool viscosity_use_corrected_temperature;
+  bool viscosity_use_p0_temperature;
   bool compute_dissipation;
 
   //Whether to compute rigid body modes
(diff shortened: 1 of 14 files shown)


src/landIce/problems/LandIce_StokesFOBase.hpp
```diff
@@ -431,7 +434,9 @@ constructStatesEvaluators (PHX::FieldManager<PHAL::AlbanyTraits>& fm0,
 
     // Get data layout
     if (rank == FRT::Scalar) {
-      state_dl = loc == FL::Node ? dl->node_scalar : dl->cell_scalar2;
+      state_dl = loc == FL::Node ? dl->node_scalar :
+                        loc == FL::Cell ? dl->cell_scalar2 :
+                        dl->qp_scalar;
```
bartgol, line 439: States are then used in the GenericSTKMeshStruct, in case we need to load something from file. QuadPoint is not handled there, but I think we only need qp states for output, right? If we need them for input, we would get errors in the STK mesh struct.

mperego: What I need to do, and it works, is to put in the mesh a temperature field on quadrature points. 
Somethig like this
```
using QPScalarFieldType = Albany::AbstractSTKFieldContainer::QPScalarFieldType;
QPScalarFieldType* temperature_field = meshStruct->metaData->get_field <QPScalarFieldType> (stk::topology::ELEMENT_RANK, "temperature");
```
...
//loop over elements
```
        double* temperature = stk::mesh::field_data(*temperature_field, elem);
        for(int qp=0; qp<numQPs; qp++) {
          temperature[qp] = myData(cell,qp);
        }
```

It might be that that modification is not needed.. I kind of added the `FL::Cell` case in every place I saw `FL:Node`.

mperego: So yes, we do not need to  load the QP field from file.

bartgol: Great, then we're good. I don't think we need to add check, if the user specifies an input on QP. There will already be an error somewhere in the STK mesh.

bartgol (approved):

mperego:
Thanks @bartgol !

PR #956, opened 2023-05-16
LandIce: adding Depth Integrated Model

mperego:
This PR adds the` Depth Integrated Model` obtained from the `FO` model by using only one mesh layer and replacing the reference trial and test functions for wedges with a tensor-product basis (triangle X line) having the usual P1 bases for the triangle and the basis functions `{1-(1-z)^4, (1-z)^4}` instead of `{z, 1-z}` in the vertical direction (for `z in [0,1]`). 
While our implementation differs, this model is equivalent to other models in the literature, e.g. the MOLHO model https://tc.copernicus.org/articles/16/179/2022/.  

Added a couple of tests using the new capability. We'll also modify MPAS tests to use this capability.

The model is enabled by adding `Depth Integrated Model: true` to the `Problem` sublist of a `LandIce Stokes First Order 3D` or a `LandIce Stokes FO Thermo Coupled 3D` problem. 
At the moment we use the same basis functions for temperature and velocity, but we might change this in the future.
Also, we plan to allow prescribing different Lagrangian polynomial basis functions in alternative to `{1-(1-z)^4, (1-z)^4}`.

### DIFF
--- a/src/evaluators/utility/PHAL_ComputeBasisFunctions_Def.hpp
+++ b/src/evaluators/utility/PHAL_ComputeBasisFunctions_Def.hpp
@@ -15,19 +15,18 @@ template<typename EvalT, typename Traits>
 ComputeBasisFunctions<EvalT, Traits>::
 ComputeBasisFunctions(const Teuchos::ParameterList& p,
                               const Teuchos::RCP<Albany::Layouts>& dl) :
-  coordVec      (p.get<std::string>  ("Coordinate Vector Name"), dl->vertices_vector ),
-  cubature      (p.get<Teuchos::RCP <Intrepid2::Cubature<PHX::Device> > >("Cubature")),
-  intrepidBasis (p.get<Teuchos::RCP<Intrepid2::Basis<PHX::Device, RealType, RealType> > > ("Intrepid2 Basis") ),
-  weighted_measure (p.get<std::string>  ("Weights Name"), dl->qp_scalar ),
-  jacobian_det (p.get<std::string>  ("Jacobian Det Name"), dl->qp_scalar ),
-  BF            (p.get<std::string>  ("BF Name"), dl->node_qp_scalar),
-  wBF           (p.get<std::string>  ("Weighted BF Name"), dl->node_qp_scalar),
-  GradBF        (p.get<std::string>  ("Gradient BF Name"), dl->node_qp_gradient),
-  wGradBF       (p.get<std::string>  ("Weighted Gradient BF Name"), dl->node_qp_gradient)
+  coordVec          (p.get<std::string>  ("Coordinate Vector Name"), dl->vertices_vector ),
+  cubature          (p.get<Teuchos::RCP <Intrepid2::Cubature<PHX::Device> > >("Cubature")),
+  intrepid2FEBasis  (p.get<Teuchos::RCP<Intrepid2::Basis<PHX::Device, RealType, RealType> > > ("Intrepid2 FE Basis") ),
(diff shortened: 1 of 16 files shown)


src/evaluators/utility/PHAL_ComputeBasisFunctions_Def.hpp
```diff
@@ -100,12 +99,12 @@ evaluateFields(typename Traits::EvalData workset)
   typedef typename Intrepid2::CellTools<PHX::Device>   ICT;
   typedef Intrepid2::FunctionSpaceTools<PHX::Device>   IFST;
 
-  ICT::setJacobian(jacobian, refPoints, coordVec.get_view(), intrepidBasis);
+  ICT::setJacobian(jacobian, refPoints, coordVec.get_view(), intrepid2MapBasis);
```
bartgol, line 102: Is there a reason not to use the FE basis for this?

src/landIce/interfaceWithMPAS/Interface.cpp
```diff
@@ -200,18 +221,70 @@ void velocity_solver_solve_fo(int nLayers, int globalVerticesStride,
...
+        }
+      }
+    } else {  //P0 temperature
+      ScalarFieldType* temperature_field = meshStruct->metaData->get_field<ScalarFieldType>(stk::topology::ELEMENT_RANK, "temperature");
+      for(int ib=0; ib <indexToTriangleID.size(); ++ib ) {
+        stk::mesh::Entity elem = meshStruct->bulkData->get_entity(stk::topology::ELEMENT_RANK, indexToTriangleID[ib]);
+        double* temperature = stk::mesh::field_data(*temperature_field, elem);
+        for(int il=0; il<nLayers; il++) {
+          int lId = il * lElemColumnShift + elemLayerShift * ib;  
+          double tempFraction = temperatureDataOnPrisms[lId]*(levelsNormalizedThickness[il+1]-levelsNormalizedThickness[il]);
```
bartgol, line 269: I'm not surewhat this calculation and the if statment below are about

src/landIce/interfaceWithMPAS/Interface.cpp
```diff
@@ -68,6 +69,10 @@ double MPAS_gravity(9.8), MPAS_rho_ice(910.0), MPAS_rho_seawater(1028.0), MPAS_s
     MPAS_ClausiusClapeyoronCoeff(9.7546e-8);
 bool MPAS_useGLP(true);
 
+bool depthIntegratedModel(false);
+
+std::vector<int> dirichletNodesIdsSepthInt;
```
bartgol, line 74: I'm assuming this was meant to be DepthInt? Fixing it might help to read the code.

mperego: Oh, right, I'll fix this.

src/evaluators/utility/PHAL_ComputeBasisFunctions_Def.hpp
mperego: Yes, the Jacobian is based on the mapping between the reference and the physical element, which is different in general from the finite element field defined on the element. You could have an affine mapping and then use P2 as finite element, or, like in the depth-integrated case a weird forth order basis.

src/landIce/interfaceWithMPAS/Interface.cpp
mperego: This code computes an average of the MPAS P0 temperature along the column and saves it as a P0 quantity in the Albany one-layer field. I'll add a comment.

mperego:
@bartgol added more comments and renamed the dirichlet field.

bartgol (approved):

mperego:
Thanks a lot @bartgol !

PR #959, opened 2023-05-25
Remove warnings

kliegeois:
@bartgol can you check if the line:
```
dim1 = stk::mesh::field_extent1_per_entity(*node_field,buck);
```
is correct now?

Before, `dim1` was used uninitialized.


bartgol (approved):
Good catch!

mperego:
Thanks @kliegeois !

PR #960, opened 2023-05-25
Download and build omegah at config time if not found

bartgol:
This PR allows to build Omega_h on the fly when configuring Albany, if an existing installation is not found.

Note: if you have an installation, the common way to tell CMake where to find MyPkg is to set the env var `MyPkg_ROOT`, so in this case, `Omega_h_ROOT` is the env var that CMake checks first. If that's not set, we check `${CMAKE_BINARY_DIR}/tpls/omegah/install`, since that's where we would automatically install it if no installation was found. If this fails too, we download, configure, build, and install Omega_h manually.

@ikalash With this PR, we should be able to add `ENABLE_OMEGAH` to all our builds without having to manually install Omega_h on all machines. The nightlies would simply build omegah on the fly when configuring Albany. In time, we will likely go on the important machines, install a properly tuned version of omegah, and set the `Omega_h_ROOT` env var to point to that location. This PR makes that less of a priority though.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,3 +1,23 @@
+function(dump_cmake_variables)
+  set (opts)
+  set (args1v REGEX)
+  set (argsMv)
+  cmake_parse_arguments (dcv "${opts}" "${args1v}" "${argsMv}" ${ARGN})
+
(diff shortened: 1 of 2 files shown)


mperego (approved):
Looks good to me

Issue #930, opened 2023-04-12
Failing tests in clang CEE build


ikalash:
All the tests in the clang CEE build are failing due to seg faults: https://sems-cdash-son.sandia.gov/cdash/viewTest.php?onlyfailed&buildid=47293 .  Curiously the debug clang build is OK.  Could be some system related thing...  I will see what happens if I build/run the code from scratch.

ikalash:
Building from scratch / running has no problems.  Will see what happens in subsequent nights...

ikalash:
I looked at this some more today, including moving to a different machine.  When the nightly test scripts create the Albany build, all tests seg fault.  When I build from scratch using the Trilinos build created by the scripts, the tests pass.  I am stumped on this, and am afraid I will have to punt on it for now b/c I don't have time to troubleshoot it...

I have deactivated the clang CEE release build.  The clang CEE debug build is still up.  If anyone is willing to take this on in the next couple of weeks while I'm traveling, that would be great.  I'm happy to provide scripts to reproduce the problem.

ikalash:
So I've finally had a chance to look into this problem, and have some additional info.  Here is what I have observed: 

- The following PR by @bartgol merged on 4/11 is what broke the code.  I verified this through bisection.  https://github.com/sandialabs/Albany/commit/3ded89bf8043c5b11b041a7e38e44753a06d3fd6 
- The problem (segfault) happens only when Albany is built using the nightly testing cdash builds with clang + release.  If I build the code straightup with cmake and clang + release, things are fine.
- The issue does not show up in any other builds, including the clang + debug CEE build.
- The issue is not specific to any of the cee-compute001 nodes - seems to be happening everywhere.

I attempted to debug the problem on the SteadyHeat2D/inputT.yaml test case using print statements and traced it down to this line: https://github.com/sandialabs/Albany/blob/master/src/InitialCondition.cpp#L267 .  When I print elemLID, in the "good" builds, I get 0.  In the clang cee-compute001 compute build, I get some crazy number, e.g., 192267392, which is clearly wrong.  Interestingly every time I run, I get a different number!

I am not sure how we want to proceed here...  I have a worry that there is an underlying problem somewhere in the code (Trilinos or Albany) that is causing this (an initiated value perhaps), but it is weird that it doesn't show up in the other builds.  Thougthts?  @kliegeois , @mperego , @bartgol .

[...]

PR #962, opened 2023-06-01
Fix clang release build

jewatkins:
See #930.

Some optimization in the clang release build is causing the data in elem_lids to go out of scope. This fixes it.

This also gets rid of the warning associated with the dual view:
```
[Albany/src/Albany_DualView.hpp:88](https://blob/master/Albany/src/Albany_DualView.hpp#L88):3: warning: conversion function converting 'Albany::DualView<const int *>' to itself will never be used [-Wclass-conversion]
  operator const_type() const {
```

@bartgol @mcarlson801 Can you remind me again why we're not using a `Kokkos::DualView` for this?

### DIFF
--- a/src/Albany_DualView.hpp
+++ b/src/Albany_DualView.hpp
@@ -84,8 +84,8 @@ struct DualView {
   }
 
   // Allow implicit conversion to DualView<const DT>;
-  template<typename SrcDT = DT, typename = typename std::enable_if<not std::is_same<SrcDT,const_DT>::value>::type>
-  operator const_type() const {
+  template<typename SrcDT = DT>
+  operator DualView<std::enable_if_t<not std::is_same_v<SrcDT,const_DT>, const_DT>>() const {
     const_type ct;
     ct.d_view = d_view;
     ct.h_view = h_view;
--- a/src/InitialCondition.cpp
+++ b/src/InitialCondition.cpp
@@ -48,7 +48,8 @@ void InitialConditions (const Teuchos::RCP<Thyra_Vector>& soln,
 {
   auto soln_data = Albany::getNonconstLocalData(soln);
   const auto& dof_mgr = disc->getDOFManager();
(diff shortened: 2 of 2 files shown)


bartgol (approved):

bartgol:
@jewatkins My original thought was that I did not need all the complicated syncing mechanisms of DualView. All I needed was a simple struct with a host and dev view (and I threw a couple of utility methods on top). But perhaps my understanding of the kokkos impl is wrong, and/or I was overthinking it. I can take a look at swapping in Kokkos::DualView.

ikalash:
Thanks for doing this!  The release clang build still had failures last night (https://sems-cdash-son.sandia.gov/cdash/viewTest.php?onlyfailed&buildid=49066).  It's possible that it started running before the changes were pushed though...  I can check again tomorrow.

jewatkins:
Yeah I think it was pushed after the nightlies pulled `7594145de` which was Luca's changes that don't exist anymore?

mperego:
> which was Luca's changes that don't exist anymore

That was an Italian job.. Luca added some commits by mistake and then reverted them, but he thought it would be better to clean the history so I gave him right to force push to master to remove the commits.. but we were caught!

bartgol:
Yes, I deleted those changes cause they were not supposed to be in master. With @mperego help, we temporarily removed the branch protection on master, and I reset master so that those 5 commit and their corresponding revert were gone.

I pushed the current branch thinking I was on a branch, but I was accidentally working on master. Sorry for the confusion.

bartgol:
> > which was Luca's changes that don't exist anymore
> 
> That was an Italian job.. Luca added some commits by mistake and then reverted them, but he thought it would be better to clean the history so I temporarily gave him right to force push to master to remove the commits.. but we were caught!

We forgot to check the CCTV footage! Rookie mistake...

PR #963, opened 2023-06-02
Fixes to comply with changes in STK

mperego:
STK `Field<>` is no longer templated with the `stk::mesh::Cartesian` type:
```
using VectorField = stk::mesh::Field<double, stk::mesh::Cartesian>; //deprecated
using VectorField = stk::mesh::Field<double>; //new
```
Similarly, `get_field<>` and `declare_field<>` methods are now only templated on the data type.
More details in the attached document provided by Dave Glaze [simple_fields.pptx](https://github.com/sandialabs/Albany/files/11638824/simple_fields.pptx).

### DIFF
--- a/src/disc/stk/Albany_AbstractSTKFieldContainer.hpp
+++ b/src/disc/stk/Albany_AbstractSTKFieldContainer.hpp
@@ -32,44 +32,19 @@ namespace Albany {
 class AbstractSTKFieldContainer
 {
 public:
-  // Tensor per Node/Cell  - (Node, Dim, Dim) or (Cell,Dim,Dim)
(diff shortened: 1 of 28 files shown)


bartgol (approved):
This is so much cleaner. I really disliked having to handle scalar and vector fields separately just b/c of the template arg. 

Thanks for doing this!

jewatkins (approved):
lgtm. much cleaner

mperego:
Thanks @bartgol and @jewatkins! I tried to get rid of the different states `cell_scalar_states`, `cell_vector_states`, `qpscalar_states`.. etc and use only  `stk_state`  and use the field rank to differentiate them, but `qptensor_state` dimensions are different in stk and in the state arrays (because the `dimxdim` tensor is saved as one dimension in STK), so it made things complicated.

Issue #917, opened 2023-03-07
blake builds down due to Kokkos 4.0 promotion


jewatkins:
blake build is down due to https://github.com/trilinos/Trilinos/pull/11600

`gcc 7.2.0` is the newest version that has openmpi. `intel 19.3.199` is showing up as `Compiler: Intel 19.0.0.20190206` for some reason. And when I run `mpicxx --version`, I get `19.0.3.199`...

@ikalash @mperego We should probably discuss how we want to proceed. I could probably try to get the intel build working with a newer compiler (I think we'd have to build some tpls ourselves) but we'd have to start from scratch with the gcc build (possibly with spack although I'm not sure if I'd trust the mpi build). Not sure if we'll get much support from admins, my ticket for cmake still has no response https://github.com/sandialabs/Albany/issues/914

ikalash:
I was going to open the same issue.  I was going to suggest you try to figure out how to get the build working again with newer compilers.  Can we request that a newer gcc module get installed on blake?  If it's required for Kokkos, it will be not just us who need it.

jewatkins:
There are newer gcc modules but there are no corresponding openmpi builds. They haven't responded to my cmake ticket so I'm not sure if they would respond to a new openmpi build. Then we also need to build all the tpls. So at that point, maybe it makes sense to just do everything with the spack build?

I could try building openmpi and test it with hello world. I'm mostly just worried that it won't work across nodes if I build it myself.

ikalash:
How did you open the cmake ticket?  I opened one a couple days ago and they replied immediately, saying it was already installed.  I emailed ***@***.******@***.***> .

I would not try to push your own compilers/openmpi quite honestly...  the modules should be used for this when possible.  I think it's worth checking if they can accommodate.  Feel free to CC me when you email the help line.


From: Jerry Watkins ***@***.***>
Sent: Wednesday, March 8, 2023 7:52 PM
To: sandialabs/Albany ***@***.***>
Cc: Tezaur, Irina ***@***.***>; Mention ***@***.***>
Subject: [EXTERNAL] Re: [sandialabs/Albany] blake builds down due to Kokkos 4.0 promotion (Issue #917)


There are newer gcc modules but there are no corresponding openmpi builds. They haven't responded to my cmake ticket so I'm not sure if they would respond to a new openmpi build. Then we also need to build all the tpls. So at that point, maybe it makes sense to just do everything with the spack build?

[...]

PR #965, opened 2023-06-07
Update blake builds

jewatkins:
- part of #917
 - remove all intel builds
 - add gcc release, sfad builds
 
 I can work on updating the performance tests once this goes through.


mperego:
thanks @jewatkins! Looks good to me. I'a wondering if it would make sense for the release build to be Epetra-free.

jewatkins:
> thanks @jewatkins! Looks good to me. I'm wondering if it would make sense for the release build to be Epetra-free.

We should probably make all builds epetra-free? Maybe we can do that in a separate PR once we know this works.

mperego:
> > thanks @jewatkins! Looks good to me. I'm wondering if it would make sense for the release build to be Epetra-free.
> 
> We should probably make all builds epetra-free? Maybe we can do that in a separate PR once we know this works.

Sounds good. I would probably keep testing Epetra in a couple of builds.

mperego (approved):

PR #967, opened 2023-07-13
STK fixes

jewatkins:
Looks good to me, thanks Luca


lbertagna:
Thanks for opening the PR. I forgot I could have forked albany in my personal account, and issue a pr from the fork...

jewatkins:
is this good to push?

lbertagna:
I am still running the testsuite (corePDEs passed). However, given that this was a matter of compilation errors, and given that we don't check output files content in our tests (something we should do at some point), I don't foresee any failure, so we could probably merge right away.

jewatkins:
sounds good thanks

mperego:
thanks @lbertagna and @jewatkins !

PR #969, opened 2023-08-02
Add support for UVM-free Trilinos (Velocity)

mcarlson801:
This PR makes a number of changes to Albany to allow for a velocity problem to be solved when Albany is built against Trilinos that has Cuda UVM disabled. The following changes were made in order to ensure data is accessed only in the proper execution space during runtime:

1) Kokkos views (mostly related to sidesets) were replaced with Kokkos DualView's and are synced to device after being populated on host.
2) Added a sync for a DualView in GatherSolution that caused a bug only when UVM is disabled.
3) Fully ported some evaluators that had some dangling code that still ran on host

The changes from this branch have been tested successfully (with UVM disabled) on Greenland and Humboldt velocity solves and on the full suite of tests with UVM enabled. The only tests that currently fail are the two SHMIP tests that are known to be failing.

Notes:
-Responses are not currently ported to device so they will need to be disabled if running a velocity problem with UVM disabled.
-Some exception checks related to invalid parameters were lost when porting BasalFrictionCoefficient to device, can these exceptions be thrown within a device kernel? The values live on device so if the exceptions can only be thrown on host, I'll need to figure something out.

### DIFF
--- a/src/evaluators/gather/PHAL_GatherSolution_Def.hpp
+++ b/src/evaluators/gather/PHAL_GatherSolution_Def.hpp
@@ -182,6 +182,13 @@ postRegistrationSetup(typename Traits::SetupData d,
         device_sol.val_dotdot_kokkos[i]=this->val_dotdot[i].get_static_view();
       }
     }
+    device_sol.val_kokkos.host_to_device();
+    if (enableTransient){
+      device_sol.val_dot_kokkos.host_to_device();
+    }
+    if (enableAcceleration){
+      device_sol.val_dotdot_kokkos.host_to_device();
+    }
 #endif
   } else if (tensorRank == 1) {
     this->utils.setFieldData(valVec,fm);
--- a/src/evaluators/state/PHAL_LoadStateField_Def.hpp
+++ b/src/evaluators/state/PHAL_LoadStateField_Def.hpp
@@ -54,46 +54,11 @@ void LoadStateFieldBase<EvalT, Traits, ScalarType>::evaluateFields(typename Trai
   auto stateData   = stateToLoad.dev();
   const int stateToLoad_size = stateToLoad.size();
 
-  std::vector<size_t> data_dims;
-  data.dimensions(data_dims);
-
-  const int d1 = data_dims[0];
-  const int d2 = (data_dims.size() > 1) ? data_dims[1] : 0;
-  const int d3 = (data_dims.size() > 2) ? data_dims[2] : 0;
-  const int d4 = (data_dims.size() > 3) ? data_dims[3] : 0;
-  const int d5 = (data_dims.size() > 4) ? data_dims[4] : 0;
-
-  const int data_dims_size = data_dims.size();
+  PHAL::MDFieldVectorRight<ScalarType> dataVec(data);
 
   Kokkos::parallel_for(Kokkos::RangePolicy(0,data.size()),
                        KOKKOS_CLASS_LAMBDA(const int i) {
-    if (data_dims_size == 1) {
-      data(i) = (i < stateToLoad_size) ? stateData(i) : 0.0;
-    } else if (data_dims_size == 2) {
-      int i2 = i % d2;
-      int i1 = i / d2;
-      data(i1,i2) = (i < stateToLoad_size) ? stateData(i) : 0.0;
-    } else if (data_dims_size == 3) {
-      int i3 = i % d3;
-      int i2 = (i / d3) % d2;
-      int i1 = i / (d3 * d2);
-      data(i1,i2,i3) = (i < stateToLoad_size) ? stateData(i) : 0.0;
-    } else if (data_dims_size == 4) {
-      int i4 = i % d4;
-      int i3 = (i / d4) % d3;
-      int i2 = (i / (d4*d3)) % d2;
-      int i1 = i / (d4*d3*d2);
-      data(i1,i2,i3,i4) = (i < stateToLoad_size) ? stateData(i) : 0.0;
-    } else if (data_dims_size == 5) {
-      int i5 = i % d5;
-      int i4 = (i / d5) % d4;
-      int i3 = (i / (d5*d4)) % d3;
-      int i2 = (i / (d5*d4*d3)) % d2;
-      int i1 = i / (d5*d4*d3*d2);
-      data(i1,i2,i3,i4,i5) = (i < stateToLoad_size) ? stateData(i) : 0.0;
-    } else {
-      // Exception
-    }
+    dataVec[i] = (i < stateToLoad_size) ? stateData(i) : 0.0;
   });
 }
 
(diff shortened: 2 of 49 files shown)


src/evaluators/gather/PHAL_GatherSolution_Def.hpp
```diff
@@ -182,6 +182,13 @@ postRegistrationSetup(typename Traits::SetupData d,
         device_sol.val_dotdot_kokkos[i]=this->val_dotdot[i].get_static_view();
       }
     }
+    device_sol.val_kokkos.host_to_device();
+    if (enableTransient){
+      device_sol.val_dot_kokkos.host_to_device();
+    }
+    if (enableAcceleration){
+      device_sol.val_dotdot_kokkos.host_to_device();
+    }
```
jewatkins, line 191: I didn't know kokkos vector had a dual view underneath. 👍

src/evaluators/state/PHAL_LoadStateField_Def.hpp
```diff
@@ -50,11 +50,51 @@ void LoadStateFieldBase<EvalT, Traits, ScalarType>::evaluateFields(typename Trai
...
+      int i5 = i % d5;
+      int i4 = (i / d5) % d4;
+      int i3 = (i / (d5*d4)) % d3;
+      int i2 = (i / (d5*d4*d3)) % d2;
+      int i1 = i / (d5*d4*d3*d2);
+      data(i1,i2,i3,i4,i5) = (i < stateToLoad_size) ? stateData(i) : 0.0;
+    } else {
+      // Exception
+    }
+  });
```
jewatkins, line 97: Are you sure the state data indices always unroll the same? Also, this looks like it should be `RangePolicy(0, stateToLoad_size)`? Also you could create 5 lambda kernels here. An exception could also be thrown outside the kernel for `data_dims_size >5`.

src/evaluators/state/PHAL_LoadStateField_Def.hpp
```diff
@@ -91,13 +131,52 @@ void LoadStateField<EvalT, Traits>::evaluateFields(typename Traits::EvalData wor
...
```
jewatkins, line 179: This looks the same as the base evaluator, you could probably consolidate into one functor/kernel.

src/landIce/evaluators/LandIce_BasalFrictionCoefficient_Def.hpp
```diff
@@ -504,69 +528,6 @@ evaluateFields (typename Traits::EvalData workset)
...
-    if (lambda_type==FIELD_TYPE::CONSTANT && printedLambda!=lambda) {
-      *output << "[Basal Friction Coefficient" << PHX::print<EvalT>() << "] lambda = " << lambda << "\n";
-      printedLambda = lambda;
-    }
-#endif
-
-    TEUCHOS_TEST_FOR_EXCEPTION ((lambda_type == FIELD_TYPE::CONSTANT) && lambda<0, Teuchos::Exceptions::InvalidParameter,
-                                "\nError in LandIce::BasalFrictionCoefficient: \"Bed Roughness\" must be >= 0.\n");
-  }
-
```
jewatkins, line 569: You'll have to deep copy the values onto host to test the exception.

jewatkins (changes requested):
Looks like a good first step. We'll have to work on responses later.

src/evaluators/state/PHAL_LoadStateField_Def.hpp
mcarlson801: According to the MDFieldIterator documentation, this should be consistent with how the indices are unrolled. As for the RangePolicy size, I did it this way to be consistent with what we had before. Anything beyond stateToLoad_size will be set to 0.0. This can be changed if we know MDFields get initialized to zero.

I'll go ahead and add that exception.

src/landIce/evaluators/LandIce_BasalFrictionCoefficient_Def.hpp
mcarlson801: Sounds good, I'll add this shortly.

src/evaluators/state/PHAL_LoadStateField_Def.hpp
mcarlson801: Yep, they are the same. I'll consolidate.

src/evaluators/state/PHAL_LoadStateField_Def.hpp
jewatkins: Ah okay I see `MDFieldIterator` is our own thing. I would modify that to work on device so that it's consistent across all code that uses it.

jewatkins: It's defined here: https://github.com/sandialabs/Albany/blob/master/src/PHAL_Utilities.hpp#L58 . It needs to be made device friendly or completely removed so that others won't use it.

jewatkins: Also, I don't think you should rewrite this code in other evaluators. It should be made generic so that other evaluators can call the same code.

src/landIce/evaluators/LandIce_BasalFrictionCoefficient_Def.hpp
bartgol: +1 for adding back the tests. Also, if possible we should leave the possibility to print the parameters (it helps some times to debug stuff, to ensure the params are indeed what they should be).

bartgol (approved):
This looks good. Nothing particular stands out to me.

bartgol:
> -Responses are not currently ported to device so they will need to be disabled if running a velocity problem with UVM disabled.

I don't think we can run anything without responses. I mean, yes, Albany can run, but we cannot do, for instance, any analysis, since we do need responses for those. I'm not going to block this PR on the count of responses not working without UVM, but we should definitely hurry up and port them, if UVM-free builds are going to be our default.

> -Some exception checks related to invalid parameters were lost when porting BasalFrictionCoefficient to device, can these exceptions be thrown within a device kernel? The values live on device so if the exceptions can only be thrown on host, I'll need to figure something out.

I think we can probably check/print params at the beginning of evaluateFields, no?

mperego:
> > -Some exception checks related to invalid parameters were lost when porting BasalFrictionCoefficient to device, can these exceptions be thrown within a device kernel? The values live on device so if the exceptions can only be thrown on host, I'll need to figure something out.
> 
> I think we can probably check/print params at the beginning of evaluateFields, no?

Agreed, we should check the parameters at the beginning of evaluateFields.

mcarlson801:
I've run some performance comparisons for this branch with UVM on and off against master. It seems like uvm-off is consistently faster but there is enough variability in all three cases that it's hard to say for sure. Full performance run logs can be found on Perlmutter at `/pscratch/sd/m/mcarlson/AISWeakScaling/antarctica/PerlmutterData/ant-16km-20L/uvmfree_logs` and `/pscratch/sd/m/mcarlson/AISWeakScaling/antarctica/PerlmutterData/ant-1km-20L/uvmfree-logs`

bartgol:
@mcarlson801 can give world read permissions on the txt files?

mcarlson801:
@bartgol My bad, the 1km files didn't get set to fanssie but it is fixed now. Let me know if you still can't access and I'll make them readable by all.

mcarlson801:
@bartgol Oh I put the wrong locations in. The logs are located at `/global/cfs/cdirs/fanssie/users/mcarlson/AISWeakScaling_June2023/logs/ant-1km-20L/uvmfree-logs` and `/global/cfs/cdirs/fanssie/users/mcarlson/AISWeakScaling_June2023/logs/ant-16km-20L/uvmfree_logs`

jewatkins (approved):
looks good, thanks!

PR #972, opened 2023-08-16
Fix bug in stk rebalance

### LINKED ISSUE #968: SHMIP tests failing
Starting today, 2 of the SHMIP tests began failing in the nightly builds:

https://sems-cdash-son.sandia.gov/cdash/viewTest.php?onlyfailed&buildid=50583

The error looks like this: 

```
p=0: *** Caught standard std::exception of type 'std::logic_error' :

 Requirement( false ) FAILED
 Error occurred at: Albany/src/disc/stk/percept/stk_rebalance/GeomDecomp.cpp:143
 
 Error: Spatial Dimension not 1, 2, or 3, can not apply rotation.
```
@jewatkins I think this is due to the recent STK changes and your PR to fix the compilation errors that occurred.  Can you please have a look?

bartgol:
The bug was introduced in 6cb65751 while attempting to fix compilation errors due to some STK update. It was utlimately due to a misunderstanding on my end of what the routine was trying to do.

I verified that SHMIP tests now run on my laptop.

Fixes #968


bartgol:
@ikalash I'm going to merge this before I forget, so it can be tested tonight.

PR #975, opened 2023-08-18
omegah: cmake opts for kokkos, enable mpi, fix source and test paths

cwsmith:
This PR: 
- adds CMake args for Omegah via `ExternalProjectAdd(...)` to 
  - enable Kokkos (only tested with serial backend so far)
  - (hopefully) make enabling MPI for Omegah more robust
  - enabled shared libs
- sets the branch of Omegah to `reducedThrust` which has a Kokkos backend that supports AMD, Intel and Nvidia GPUs
- fixes Omegah test source and mesh paths
- one line compile error fix in `Albany_OmegahDiscretization.hpp`
- merged sandialabs/master (there was a fix for a compile error in `PHAL_Utilities_Def.hpp`)

I'll submit a follow-up PR if CMake argument changes are needed to build when the Kokkos CUDA backend is enabled.


CMakeLists.txt
```diff
@@ -810,6 +810,38 @@ if (ENABLE_KERNELGETMEMORYSIZE)
   message("-- Memory: Kernel_GetMemorySize() Enabled.")
 endif()
 
+# Set MPI args
```
bartgol, line 813: Since you mived them here, we can prob delete these settings down below (near file end).

cwsmith: Whoops.  Yeah, I meant to do that.  Good catch.

cwsmith: @bartgol Fixed with faae04967b87bcea753e79881861463321f0e247.

bartgol (approved):
Thanks!

PR #978, opened 2023-08-24
Improving computeNodeSets

OscarAntepara:
PR improves performance of computeNodeSets, making Albany Setup approx. 3x faster.
Also, adding some time monitors to Albany Setup to get more profiler info.

### DIFF
--- a/src/disc/stk/Albany_STKDiscretization.cpp
+++ b/src/disc/stk/Albany_STKDiscretization.cpp
@@ -2089,19 +2089,17 @@ STKDiscretization::computeNodeSets()
     stk::mesh::get_selected_entities(ns_selector, bulkData->buckets(NODE_RANK), nodes);
 
     // Remove nodes that are not owned according to Tpetra/Epetra
-    bool done = false;
-    while (not done) {
-      done = true;
-      auto dof_mgr = getNodeDOFManager();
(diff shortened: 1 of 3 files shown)


src/disc/stk/Albany_STKDiscretization.cpp
```diff
@@ -2089,19 +2096,17 @@ STKDiscretization::computeNodeSets()
...
-          nodes.erase(it);
-          done = false;
-          break;
-        }
+    std::vector<stk::mesh::Entity>::iterator it = nodes.begin();
+    auto dof_mgr = getNodeDOFManager();
+    while(it != nodes.end()){
+      auto gid = stk_gid(*it);
+      if (not dof_mgr->indexer()->isLocallyOwnedElement(gid)){
+        it = nodes.erase(it);
```
bartgol, line 2104: Ah, nice. I was scared by the fact that erasing invalidates pointers, but I suspect the compiler does not caches nodes.end() in the while condition, so this is fine.

bartgol (approved):

mperego:
@OscarAntepara thanks for the fix! When you say that this makes Albany setup 3x faster, are you referring to GPU runs or CPU runs?

bartgol:
This is host code, it affects all builds.

PR #984, opened 2023-09-25
Fix how Albany gets trilinos, plus some cleaning

### LINKED ISSUE #982: Failing Albany configures due to incorrect Trilinos version
All the Albany and Albany-LCM nightlies are failing as a result of Albany finding the "wrong" Trilinos version.  I checked the version of Trilinos that was built and it is 14.5.  I then tried to chance Albany/CMakeLists.txt to look for 14.5 Trilinos instead of 14.1, but it did not fix the problem.  My guess is this is due to a Trilinos change.  Does anyone have any insight into this?  Perhaps @jewatkins @bartgol @mcarlson801 @mperego @kliegeois ?

Tagging @lxmota since this impact LCM too.

bartgol:
Notably:
- albany decides its own compilers, regardless of what trilinos used
- `find_package()` must be called _after_ `project()`
- don't assume structure of trilinos install folder (if possible)
- remove unused options

Fixes #982


mperego:
Thanks @bartgol. Is this going to solve also the issues that Jim is seeing in E3SM? Note that E3SM is linking an old version of Albany.

bartgol:
> Thanks @bartgol. Is this going to solve also the issues that Jim is seeing in E3SM? Note that E3SM is linking an old version of Albany.

I don't think so. I need to re-do the installation of albany. It's not super long, but since he found a workaround, I don't think it's a priority.

bartgol:
This works on my workstation. Merging, according to discussion in today's telecon.

@ikalash We'll need to revert the hack of setting GNUInstallDirs. At the telecon we were suggesting to roll back a couple, to make sure we're still getting coverage if this fix ends up not working as expected. How do you want to handle? I can simply revert the commits you did for the dashboards, if you think that's ok.

PR #993, opened 2023-10-11
Update blake nightly testing to use new modules

jewatkins:
This updates the blake nightlies to work with the new modules. I tested the release build and all tests pass.

A couple notes:
 - I had to edit the cmakelists to use mpirun when executing tests. blake srun seems to be configured to only work with the openmpi package that's installed through the modules. Maybe there's a better workaround...
 - The python module on blake does not include pip so that needs to be installed locally and added to PATH...

 @mperego @kliegeois fyi


bartgol (approved):
Yeah, in order to use srun, mpi must have slurm support compiled in. I have the same issue on mappy (only the acme-openmpi module has slurm compiled in, sems modules don't). It's not a big deal if we are already on a compute node.

I don't see where you replaced srun with mpirun though...

jewatkins:
we actually don't use srun, we default to mpiexec which works fine but running the binary directly without mpi throws an error about srun so I assume it tries to use srun in that case?

bartgol:
> we actually don't use srun, we default to mpiexec which works fine but running the binary directly without mpi throws an error about srun so I assume it tries to use srun in that case?

That's something I've never seen actually. I suppose that, if mpi was compiled with srun support, the call to MPI_Initialize may check that slurm/mpi env vars are set, and throw if they are not (which may be the case if running without mpiexec). I'm not really sure. So long as it works with mpiexex/srun, that's ok.

PR #997, opened 2023-10-17
Remove unused/unreachable code related to nodal fields/states

bartgol:
This PR removes some code related to mesh states that was not only unused, but unreachable. That is, it was either never called, or only called by code that could not be triggered.

I did this cleanup as a preliminary step before adding full omegah support in the discretization, to reduce the amount of code I need to follow...


mperego (approved):
Nice cleaning. Looks good to me.

ikalash:
I tested it and it's good.  I will merge now.  Thanks!

Issue #998, opened 2023-10-20
Compilation errors coming from Omegah in warning-free blake build


ikalash:
I turned on Omega_h in the blake nightlies and there are failures in the warning-free one: 

```
[Albany/src/disc/omegah/OmegahPermutation.hpp:103](https://blob/master/Albany/src/disc/omegah/OmegahPermutation.hpp#L103):25: error: comparison of integer expressions of different signedness: 'int' and 'unsigned int' [-Werror=sign-compare]
```
https://sems-cdash-son.sandia.gov/cdash/viewBuildError.php?buildid=53394

@cwsmith you may be the best one to look at this, but please let me know if I am mistaken.

cwsmith:
@ikalash Taking a look now.  Thank you.

cwsmith:
@ikalash A fix for the `OmegahPermutation.hpp` warnings (there were a few more in my local build) are here https://github.com/sandialabs/Albany/pull/1000.  

I saw a few more warnings (gcc 10 `-Wall`) in some of the other new Omegah source files.  Do you see any others in the dashboard?  I can't access that site.

PR #1000, opened 2023-10-20
omegah: silence warnings in OmegahPermutation.hpp

cwsmith:
Addresses #998.
Compiled locally with gcc 10.1.0 and `-Wall`.


ikalash:
I will merge this.  I tested it and it gets past the Omega_h error but hits another error that is happening in the warning-free build (https://github.com/trilinos/Trilinos/issues/12418).

PR #1007, opened 2023-11-07
Change cmake logic to find/build omegah

bartgol:
* Use FetchContent rather than ExternalProject: the resulting cmake code is much smaller and simpler to maintain.
* When the found installation is not compatible with the needed configuration, we can no longer install locally, since FetchContent creates a project within the project (while ExternalProject creates a separate project), and the Omega_h targets (first defined when imported by find_package) cannot be redefined.

@cwsmith this requires SCOREC/omega_h#71.


cwsmith (commented):
Very nice!  Thank you.

Does this set omegah to configure/build with shared libs if an existing install is not found?

bartgol:
> Very nice! Thank you.
> 
> Does this set omegah to configure/build with shared libs if an existing install is not found?

It builds it just like Albany. If you configure Albany with `BUILD_SHARED_LIBS=ON`, it will build omegah with shared libs, otherwise static.

cwsmith (approved):

PR #1010, opened 2023-11-09
Cleaning up the set field/bulk data stage in the mesh structures

bartgol:
I decided to do this while trying to add extruded meshes in omegah. Cleaning up a bit this stuff helps me with the omegah effort.

Summary
- States are not needed when setting up the bulk data. This phase is in charge to set up the mesh topological information (entities, coords, ...)
- Worksetsize not needed, since it's already stored in the meshSpecs.
- Ioss was doing certain things twice (first in setFieldData, then again in setBulkData).
- Most STK meshes were doing the same thing in setFieldData, so there was a bit of code duplication.


mperego (approved):
It looks good to me.
Is the "Use Serial Mesh" case still supported?

Now that you are actively working on the meshes, we really should have some tests that check the output meshes to be sure we are not breaking some capability.

bartgol:
> It looks good to me. Is the "Use Serial Mesh" case still supported?
> 
> Now that you are actively working on the meshes, we really should have some tests that check the output meshes to be sure we are not breaking some capability.

I don't know. But I agree that we should add tests that compare output files... I'll think of something.

mperego:
@bartgol , would you mind having a look at https://github.com/sandialabs/Albany/blob/master/src/landIce/interfaceWithMPAS/Albany_MpasSTKMeshStruct.cpp and see if it needs some clean up as well?

bartgol:
I think this cleanup has uncovered some bugs related to how we handled stk native objects. I'm not entirely sure what I'm saying is true, but I think that, despite our recent cleanup, we were still using stk fields with tags (rather than "simple" fields) when reading input meshes (e.g., `Field<double,Cartesian>` instead of simply `Field<double>`). This was causing issues in this branch, since we later try to declare fields just with the data type, causing incompatibility of tags.

I fixed a few issues, but I'm still getting a couple of errors. I will try to fix this.

bartgol:
Update: everything is almost ready, but I have one final problem, with hte PopulateMesh tests. In particular, this test puts a solution on the mesh, which has a certain number of equations. When later other problems load the mesh and attempt to declare the solution field, stk complains b/c the layout is wrong (b/c of a different number of equations).

I have two strategies:
 - I quickly wrote a bash script that uses nco to remove the solution field, leaving a pristine exo file. We can then run this script as a post-processing script after the PopulateMeshes test runs, but we have to run it _for all the exo files_, since we may be running in parallel.
 - Figure out a way to _not_ add the solution field to the output mesh if a certain parameter is set via input file. I think this should be doable, and maybe easier, I just need to try it out.

bartgol:
I think I found a solution: I added an option for STK meshes to _not_ save the solution field(s). This option is only used by the PopulateMesh problem, to avoid polluting the populated mesh with a trivial solution field, and running in the issue described above.

I am currently testing the branch. If everything works, I think I'm inclined to merge...

bartgol:
Update: corePDEs tests now all pass. Still have fails in LandIce, which I need to debug.

Note: there were quite a few inconsistencies with how we used the stk::mesh::Field objects. Sometimes we used them as "simple fields" (without array dim tags) and sometimes as "not simple fields" (with tags). I don't know excatly which part of this PR triggered the new behavior (maybe I added a flag while building trilinos), but I started to expose these inconsistencies. Since stk may at some point move to "only simple fields", all these changes are probably for the best...

bartgol:
@mperego All tests pass now. I know it's a relatively large PR, so no rush on the review.

tests/landIce/Enthalpy/input_kleiner_B.yaml
```diff
@@ -124,7 +124,7 @@ ANONYMOUS:
             Field Origin: File
             Field Value: 0.0
           Field 4:
-            Number Of Layers: 41
+            Number Of Layers: 81
```
mperego, line 127: why did you change this?

tests/landIce/ExoMeshes/square_quad4.jou
mperego, line 1: Why are you removing this?

mperego (commented):
Thanks Luca! 
Looks good to me, but I want to try it out a little bit with MALI.

tests/landIce/Enthalpy/input_kleiner_B.yaml
bartgol: Because the actual data (the ascii file) has 81 layers. I am not entirely sure how things worked before, but some changes I made ended up exposing this.

tests/landIce/ExoMeshes/square_quad4.jou
bartgol: Cause we don't use it? I can bring it back, if you think there's value in it.

bartgol:
@mperego The mali smoke test runs to completion now, and the albany log show that the solver converged. Should I run some other test?

mperego:
Let me try it out with the full compass suite

mperego:
Compass tests pass but the results are not BFB with previous baseline (although errors are of the order of machine precision or close to). Is this expected?

mperego (approved):

PR #1015, opened 2023-11-20
Fix CUDA build error related to Omega_h disc

### LINKED ISSUE #1011: Failing Perlmutter build with Omega_h enabled
It looks like the Perlmutter build is failing with Omega_h enabled: https://my.cdash.org/viewBuildError.php?buildid=2438487 .  

```
/.../repos/Albany/src/disc/omegah/Albany_OmegahBoxMesh_Def.hpp(103): error: The enclosing parent function ("OmegahBoxMesh") for an extended __host__ __device__ lambda must allow its address to be taken
```

@mcarlson801 : I also noticed that the MALI and compass builds never have a configure post to the CDash site.  Do you know why?

bartgol:
This PR fixes a few things:

- The typedef used for the coords views in Albany Omega_h meshes did not match what Omega_h objects.
- DeviceView2d was hard-coding layout left, but this is not right for CPU builds. @jewatkins I don't know if our perf on CPU has looked bad recently, but using a layout left for 2d views (for 1d it's irrelevant) may have made us a tad slower. This would only be used inside the gather/scatter kernels for the Jacobian, so the impact is limited.
- Require Trilinos 15.1. @ikalash I am not sure why I get the error while nightlies don't, but basically, on my laptop the current (15.1) installation of Trilinos is rejected. My guess was that 15.1 is somewhat marked as incompatible with previous versions (indeed, some packages are removed), so our call to `find_package (Trilinos 14.1 REQUIRED)` would fail. But the fact that nightlies happily built makes me second guessing my guess. That said, I think changing the required version to 14.1 should not break anything, so maybe it's ok?

Fixes #1011 

Edit: @mperego do you know if Trilinos 15.1 was supposed to break code requesting an old Trilinos version (like 14.1)?

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -59,7 +59,7 @@ if (ALBANY_TRILINOS_DIR)
 endif()
 message ("-- Looking for trilinos installation ...")
 message ("     ALBANY_TRILINOS_DIR: ${ALBANY_TRILINOS_DIR}") 
-find_package(Trilinos 14.1 REQUIRED)
+find_package(Trilinos 15.1 REQUIRED)
 message("-- Looking for trilinos installation ... Found.\n")
 
 # Trilinos_BIN_DIRS probably should be defined in the Trilinos config. Until it is, set it here.
--- a/src/Albany_KokkosTypes.hpp
+++ b/src/Albany_KokkosTypes.hpp
@@ -39,8 +39,6 @@ using ViewLR = Kokkos::View<DT,Kokkos::LayoutRight,MemSpace>;
 // NOTE: Tpetra may use a different LO type (Albany uses int32, while tpetra uses int). When extracting local views/matrices,
 //       be careful about this. At worst, you may need to extract pointers and reinterpret_cast them.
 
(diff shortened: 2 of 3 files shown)


mcarlson801 (commented):
This looks good to me. I'm interested to see how significant of a performance impact the memory layout change will have, I imagine it should be positive. Thanks!

mcarlson801 (approved):

jewatkins (approved):
lgtm thanks! We haven't tracked CPU performance since the summer when blake received new hardware so we won't be able to see if this makes a difference. We could potentially look at old history to see what happened before this was changed to left layout but probably not worth looking into it.

cwsmith (approved):

mperego:
> @mperego do you know if Trilinos 15.1 was supposed to break code requesting an old Trilinos version (like `14.1)?

I don't know.

jewatkins:
> But the fact that nightlies happily built makes me second guessing my guess.

The nightlies hadn't built with https://github.com/trilinos/Trilinos/commit/3108744c056ae05d3b0c6069227677c4b71470e4 yet. Nothing ran last night because of the version change.

Issue #1014, opened 2023-11-16
Albany warning-free array-bounds error on Blake


mcarlson801:
The blake warning-free build is failing with an `array-bounds` warning-as-error.

```/home/projects/albany/nightlyCDashTrilinosBlake/build-gcc/TrilinosDebugInstallGccNoWarn/include/Sacado_StaticArrayTraits.hpp:87:20: error: 'void* memset(void*, int, size_t)' offset [136, 137438953416] is out of the bounds [0, 128] of object 'val' with type 'LandIce::EnthalpyBasalResid<PHAL::AlbanyTraits::HessianVec, PHAL::AlbanyTraits, Sacado::Fad::Exp::GeneralFad<Sacado::Fad::Exp::DynamicStorage<Sacado::Fad::Exp::GeneralFad<Sacado::Fad::Exp::StaticFixedStorage<double, 1> >, Sacado::Fad::Exp::GeneralFad<Sacado::Fad::Exp::StaticFixedStorage<double, 1> > > > >::ScalarT [4]' {aka 'Sacado::Fad::Exp::GeneralFad<Sacado::Fad::Exp::DynamicStorage<Sacado::Fad::Exp::GeneralFad<Sacado::Fad::Exp::StaticFixedStorage<double, 1> >, Sacado::Fad::Exp::GeneralFad<Sacado::Fad::Exp::StaticFixedStorage<double, 1> > > > [4]'} [-Werror=array-bounds]```

The error output points to the Enthalpy_BasalResid kernel when specialized for EvalT = HessianVec and the array-bounds error seems to be related to the Sacado type.

`In file included from [Albany/src/landIce/evaluators/LandIce_EnthalpyBasalResid.cpp:11](https://blob/master/Albany/src/landIce/evaluators/LandIce_EnthalpyBasalResid.cpp#L11):
[...]

PR #1016, opened 2023-11-21
Fix for issue #1014

mcarlson801:
This fixes the improper initialization of an array of ScalarT in Enthalpy_BasalResid and should fix the error found in issue #1014.

### DIFF
--- a/src/landIce/evaluators/LandIce_EnthalpyBasalResid_Def.hpp
+++ b/src/landIce/evaluators/LandIce_EnthalpyBasalResid_Def.hpp
@@ -81,9 +81,9 @@ operator() (const Enthalpy_Basal_Residual_Tag& tag, const int& sideSet_idx) cons
   const int cell = sideSet.ws_elem_idx.d_view(sideSet_idx);
   const int side = sideSet.side_pos.d_view(sideSet_idx);
 
-  ScalarT val[maxNumNodesPerSide];
+  ScalarT val[maxNumNodesPerSide] = {};
   for (unsigned int node = 0; node < numSideNodes; ++node) {
-      val[node] = 0;
+      val[node] = ScalarT(0);
       for (unsigned int qp = 0; qp < numSideQPs; ++qp) {
       val[node] += basalMeltRateQP(sideSet_idx,qp) 
                  * BF(sideSet_idx,node,qp) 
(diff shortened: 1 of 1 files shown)


src/landIce/evaluators/LandIce_EnthalpyBasalResid_Def.hpp
```diff
@@ -81,9 +81,9 @@ operator() (const Enthalpy_Basal_Residual_Tag& tag, const int& sideSet_idx) cons
   const int cell = sideSet.ws_elem_idx.d_view(sideSet_idx);
   const int side = sideSet.side_pos.d_view(sideSet_idx);
 
-  ScalarT val[maxNumNodesPerSide];
+  ScalarT val[maxNumNodesPerSide] = {};
   for (unsigned int node = 0; node < numSideNodes; ++node) {
-      val[node] = 0;
+      val[node] = ScalarT(0);
```
mperego, line 86: I would remove this line.

mcarlson801: Good call, it's removed now

mperego (approved):
Looks good to me. Thanks!

src/landIce/evaluators/LandIce_EnthalpyBasalResid_Def.hpp
```diff
@@ -81,9 +81,8 @@ operator() (const Enthalpy_Basal_Residual_Tag& tag, const int& sideSet_idx) cons
   const int cell = sideSet.ws_elem_idx.d_view(sideSet_idx);
   const int side = sideSet.side_pos.d_view(sideSet_idx);
 
-  ScalarT val[maxNumNodesPerSide];
+  ScalarT val[maxNumNodesPerSide] = {};
   for (unsigned int node = 0; node < numSideNodes; ++node) {
-      val[node] = 0;
```
bartgol, line 86: Are we 100% sure that the default ctor of a sacado type is to init it as if with ScalarT(0)? If so, then this is fine.

bartgol (approved):

jewatkins (approved):

src/landIce/evaluators/LandIce_EnthalpyBasalResid_Def.hpp
mcarlson801: Looking at the docs for DFad and SimpleFad, it looks like the default constructors do indeed initialize to 0.

PR #1021, opened 2023-11-30
Add gcc/intel python env on blake

jewatkins:
The intel build requires it's own mpi4py installation so I added two separate gcc/intel python environments on blake. This should fix the intel pyAlbany tests.


mperego (approved):
lgtm.Thanks!

PR #1022, opened 2023-12-01
Enable uvm-free testing on Perlmutter

mcarlson801:
This PR does the following:

1) Adds the `ALBANY_ENABLE_CUDA_UVM` flag so that CUDA testing can be separated into uvm-on and uvm-off
2) For uvm-off CUDA builds, enables a small selection of tests that either are currently passing when uvm is off or will be passing in the very near future
3) Adds portable (_kokkos) versions of MueLu FO_AIS and FO_GIS tests. (I need to port solution average response to get these to pass so that will be the next step)
4) Updates to Perlmutter automated testing scripts (some modules needed to be updated after this week's maintenance and some minor general fixes to automation scripts)
5) Added uvm-off build and test to Perlmutter dashboard (uvm-on build is now named `pm-gpu-uvm` and uvm-off build is named `pm-gpu`)


tests/landIce/FO_GIS/input_fo_gis_coupledT_muelu_kokkos.yaml
```diff
@@ -0,0 +1,353 @@
+%YAML 1.1
+---
+ANONYMOUS:
+  Build Type: Tpetra
+  Debug Output:
+    Write Solution to MatrixMarket: 0
+  Problem:
+    Phalanx Graph Visualization Detail: 1
+    Solution Method: Steady
+    Name: LandIce Coupled FO H 3D
```
jewatkins, line 10: This one is currently not tested, should you remove it for now? I'm not sure if we've tried coupled with muelu even on CPU.

jewatkins (approved):
looks good, thanks!

tests/landIce/FO_AIS/inputMueLuKokkos.yaml
```diff
@@ -0,0 +1,342 @@
...
+      BC 1:
+        Type: Lateral
+        Cubature Degree: 3
+        Side Set Name: lateralside
+    Parameters:
+      Number Of Parameters: 1
+      Parameter 0:
+        Type: Scalar
+        Name: Glen's Law Homotopy Parameter
+    LandIce Physical Parameters:
```
mperego, line 42: Max, would you mind fixing the format of the floating point parameters in this input file? I'm trying to get rid of things like `9.99999999999999954e-08` in favor of `1e-07` in all input files.

mperego (approved):
Looks good, I just left a minor comment. Thanks for doing this!

CMakeLists.txt
```diff
@@ -85,6 +85,10 @@ IF(EXISTS "${Trilinos_INCLUDE_DIRS}/KokkosCore_config.h")
   STRING(REGEX MATCH "#define KOKKOS_ENABLE_CUDA" ALBANY_ENABLE_CUDA ${KOKKOS_CORE_CONFIG_FILE})
   IF(ALBANY_ENABLE_CUDA)
     MESSAGE("-- Kokkos is configured to use CUDA, Albany will also.")
+    STRING(REGEX MATCH "#define KOKKOS_ENABLE_CUDA_UVM" ALBANY_ENABLE_CUDA_UVM ${KOKKOS_CORE_CONFIG_FILE})
```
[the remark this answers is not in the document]
bartgol: That said, since we can simply access Kokkos cmake vars, I'm not sure we need these variables. We can simply use `Kokkos_XYZ`...

CMakeLists.txt
```diff
@@ -85,6 +85,10 @@ IF(EXISTS "${Trilinos_INCLUDE_DIRS}/KokkosCore_config.h")
...
+    IF(ALBANY_ENABLE_CUDA_UVM)
+      MESSAGE("-- Kokkos is configured to use CUDA UVM, Albany will also.")
+    ENDIF()
     STRING(REGEX MATCH "#define KOKKOS_COMPILER_CUDA_VERSION ([0-9]*)" _ ${KOKKOS_CORE_CONFIG_FILE})
```
bartgol, line 92: I know it's not part of your PR, but since we're dealing with cuda...what is the purpose of this line? It seems pointless to me. If it is indeed pointless, can you remove it?

tests/CMakeLists.txt
```diff
@@ -180,6 +180,18 @@ IF(${PYTHON_TEST})
   set(PYTHON_TEST_HELPERS_DIR ${CMAKE_CURRENT_SOURCE_DIR}/python)
 ENDIF()
 
+IF (ALBANY_ENABLE_CUDA AND NOT ALBANY_ENABLE_CUDA_UVM)
```
bartgol, line 183: This doesn't seem right, since it would skip landice tests on non-cuda platforms. If you're trying not to run tests with UVM on, you should simply to
```
if (NOT ALBANY_ENABLE_CUDA_UVM)
```

tests/landIce/CMakeLists.txt
```diff
@@ -1,3 +1,10 @@
+IF (ALBANY_ENABLE_CUDA AND NOT ALBANY_ENABLE_CUDA_UVM)
```
bartgol, line 1: Same here: you only need to check the latter, I think.

tests/landIce/FO_AIS/CMakeLists.txt
```diff
@@ -38,6 +38,25 @@ if (NOT ALBANY_PARALELL_EXODUS)
   set_tests_properties (${testNameRoot}_decompMesh PROPERTIES FIXTURES_SETUP antarcticaMeshSetup)
 endif()
 
+if (ALBANY_ENABLE_CUDA AND NOT ALBANY_ENABLE_CUDA_UVM)
```
bartgol, line 41: Same.

bartgol (changes requested):

CMakeLists.txt
jewatkins: I thinkt it's to print the cuda version used to build at runtime. similar to the other compiler prints at runtime.

tests/CMakeLists.txt
mcarlson801: It's a bit awkward but notice there's a return in this if-statement. If this specific condition isn't met, this block won't run and the old behavior will happen (including adding the landice tests later in a subsequent if-block). 

This seemed like the least intrusive way to enable the specific tests I want without adding `and not ALBANY_ENABLE_CUDA_UVM` to every if statement. I've confirmed the same tests run on CUDA builds, I will double check the same tests are enabled for a serial build.

bartgol: Ah, I missed that! So are we just trying to exclude code/demo if CUDA=ON and UVM=OFF, right? Then, since landice/unit subdirs are added anyways, maybe we should do something like

```
# Unit tests (always run)
if (ALBANY_UNIT_TESTS)
  add_subdirectory(unit)
endif()

# Do not run code/demo tests on GPU unless UVM is on
if (NOT ALBANY_ENABLE_CUDA OR ALBANY_ENABLE_CUDA_UVM)
  add_subdirectory(corePDEs)
  
  if (ENABLE_DEMO_PDES)
    add_subdirectory(demoPDEs)
  endif()
endif()

# Landice needs SEACAS
if(ENABLE_LANDICE AND ALBANY_SEACAS )
  add_subdirectory(landIce)
endif()
```

bartgol: Or you could still use return, but I would use it like this
```cmake
<add all subdirs that are added anyways>

if (ALBANY_ENABLE_CUDA AND NOT ALBANY_ENABLE_CUDA_UVM)
  return()
endif()

<add all remaining subdirs that are only executed if CUDA=OFF or UVM=ON
```

CMakeLists.txt
mcarlson801: What does it mean for the output variable to be `_` here? I don't think this line even does anything.

CMakeLists.txt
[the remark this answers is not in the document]
mcarlson801: I just tried out setting these variables directly like your example (`set(ALBANY_ENABLE_CUDA ${KOKKOS_ENABLE_CUDA})`) and `ALBANY_ENABLE_CUDA` was empty. Looks like we might still need to use the regexs unless I'm missing something.

CMakeLists.txt
bartgol: `_` is a legit CMake variable name. I suppose it's used to signal that this is just a placeholder. However, I'm not sure what's the point of the `STRING` call then, since the output var is the only thing that matter here...

I think we can prune this line.

CMakeLists.txt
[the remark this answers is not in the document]
bartgol: Ah, Kokkos vars are case sensitive. 
```cmake
message ("Kokkos_ENABLE_SERIAL: ${Kokkos_ENABLE_SERIAL}")
message ("KOKKOS_ENABLE_SERIAL: ${KOKKOS_ENABLE_SERIAL}")
```
prints
```
Kokkos_ENABLE_SERIAL: ON
KOKKOS_ENABLE_SERIAL: 
```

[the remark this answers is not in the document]
mcarlson801: Ah ha, that worked! Alright, I've replaced the CUDA related `ALBANY_ENABLE_whatever` variables with the corresponding `Kokkos_` variables directly and avoided calling `set` entirely.

tests/CMakeLists.txt
mcarlson801: Good call, I just updated the logic accordingly to reduce code duplication and be less confusing.

CMakeLists.txt
```diff
@@ -78,14 +78,14 @@ ENDIF()
 
 IF(EXISTS "${Trilinos_INCLUDE_DIRS}/KokkosCore_config.h")
```
bartgol, line 79: We can delete this if statement (and the else clause) and the `file(READ ...)` lines completely, I think, since we simply rely on CMake vars.

bartgol (approved):

jewatkins:
Just fyi, since you removed `KOKKOS_COMPILER_CUDA_VERSION`, you should remove it from `src/Albany_Utils.cpp` where the value dumps at runtime. I still find it useful (I sometimes use it to make sure my sim was using the correct cuda version) so it would be great to have a replacement for that line.

jewatkins:
wait... I guess we're already pulling this from kokkos? so maybe it's okay. I guess I'll find out the next time I look at a cuda run.

jewatkins:
I did a little more digging and it looks like this has been broken for a while (not sure when it stopped working, sometime in the last year). I'll probably revive it at some point.

PR #1024, opened 2024-01-08
Fix for pyAlbany parallelEnv initialization

mcarlson801:
Since the last time we tested PyAlbany on GPU it looks like Kokkos has changed slightly how it is initialized via the `InitializationSettings` struct. Instead of supplying a default <0 value for these arguments, they should just not be set.


kliegeois (approved):
It looks good to me, thanks for the fix!

PR #1026, opened 2024-02-17
Fixes and improvements to landice code and sensitiity logic

mperego:
This PR contains these changes:
1. Fixed sign in smoothing length in pressurized bed boundary condition
2. Modified the Landice viscosity computation to allow to scale it by a damage factor
3. Improved the Quadratic Linear Operator used for the Laplacian regularization. It is now possible to use a non diagonal mass matrix and optimizations are performed, e.g. when the matrix is symmetric.
4. Fixed logic that caused sensitivities not to be computed for problems with multiple parameters. This is related to https://github.com/trilinos/Trilinos/pull/12749.


src/Albany_SolverFactory.cpp
```diff
@@ -167,36 +167,29 @@ Teuchos::RCP<Thyra::ResponseOnlyModelEvaluatorBase<ST>>
...
-    }
-  } else {
-    for(int i=0; i<num_parameters; ++i) {
-      p_indices[i] = i;
-    }    
-  }
-
   if ( model_tmp->Np() > 1) {
+    std::vector<int> p_indices;
+    if (!forwardMode) {
```
bartgol, line 179: I find it easier to read "positive" if statements rather than "negative" ones, when possible. Can we swap the two branches, and use the "positive" check `if (forwardMode)`?

src/landIce/problems/LandIce_LaplacianSampling.hpp
```diff
@@ -145,29 +145,63 @@ LandIce::LaplacianSampling::constructEvaluators (PHX::FieldManager<PHAL::AlbanyT
...
-    std::string parameter_sublist_name = util::strint("Parameter", p_index+num_param_vecs);
-    Teuchos::ParameterList param_list = parameterParams.sublist(parameter_sublist_name);
-    param_name = param_list.get<std::string>("Name");
-    if(param_name == stateName) {
-      isParameter = true;
-      break;
+  {
+    stateName = "weighted_normal_sample";
+    bool isParameter = false;
+    for (unsigned int p_index=0; p_index< (unsigned int) num_dist_param_vecs; ++p_index) {
```
bartgol, line 151: Using `int` rather than `unsigned int` for `p_index` would allow to avoid the cast, and keep the line more readable...
```suggestion
    for (int p_index=0; p_index<num_dist_param_vecs; ++p_index) {
```

bartgol: I know you didn't change this, but while at it...

bartgol (commented):
Looks good. I only had a couple of "style" comments, but feel free to merge as is.

bartgol (approved):

mperego:
Thanks! I've done the requested changes. I rebased to keep the branch clean.

PR #1029, opened 2024-03-11
Cleanup ENABLE_KOKKOS_UNDER_DEVELOPMENT code

mcarlson801:
In the spirit of spring cleaning, this PR fully removes ALBANY_KOKKOS_UNDER_DEVELOPMENT and associated code.

It should be noted that PHAL_Neumann has a block of kokkos code (only for Jacobian specialization) that has been commented out since 2016 since it apparently did not work with an OpenMP backend. If we need a device-capable PHAL_Neumann in the future, we will need to re-implement this block anyway so I've removed it for now. Any objections?


doc/LandIce/cori/do-cmake-albany-cori-gcc-knl-serial-sfad12
```diff
@@ -16,7 +16,6 @@ cmake \
       -D CMAKE_VERBOSE_MAKEFILE:BOOL=OFF \
       -D ENABLE_DEMO_PDES:BOOL=OFF \
       -D ENABLE_LANDICE:BOOL=ON \
-      -D ENABLE_KOKKOS_UNDER_DEVELOPMENT:BOOL=ON \
```
bartgol, line 19: Do you guys think we should keep cori config files? It's been decommissioned a long time ago...

bartgol (approved):
Looks good, thanks!

doc/LandIce/cori/do-cmake-albany-cori-gcc-knl-serial-sfad12
jewatkins: We could remove cori but maybe as part of another PR that removes all the other junk in "doc".

jewatkins (approved):
Looks good, thanks!

mperego (approved):

doc/LandIce/cori/do-cmake-albany-cori-gcc-knl-serial-sfad12
mperego: Agreed. There's a lot of cleaning up needed int the doc folder after we got rid of Epetra.

PR #1031, opened 2024-03-13
Disable saving solution in the PopulateMesh problem

bartgol:
No need to set the yaml entry, the problem already hard-codes it


mperego (approved):
lgtm

PR #1032, opened 2024-03-20
Gather/scatter porting towards uvm-optional Albany

mcarlson801:
This PR ports a number of gather and scatter specializations to run on device and some additional utilities in ThyraUtils for getting device views from Thyra_Multivectors. Additionally, unit tests were updated to avoid ambiguous host/device accesses causing them to fail with uvm-free builds.

### DIFF
--- a/src/Albany_KokkosTypes.hpp
+++ b/src/Albany_KokkosTypes.hpp
@@ -39,13 +39,11 @@ using ViewLR = Kokkos::View<DT,Kokkos::LayoutRight,MemSpace>;
 // NOTE: Tpetra may use a different LO type (Albany uses int32, while tpetra uses int). When extracting local views/matrices,
 //       be careful about this. At worst, you may need to extract pointers and reinterpret_cast them.
 
-using DevLayout = PHX::Device::array_layout;
-
 // kokkos 1d and 2d views to be used for on-device kernels
 template<typename Scalar, typename MemoryTraits = Kokkos::MemoryUnmanaged>
-using DeviceView1d = Kokkos::View<Scalar*, DevLayout, PHX::Device, MemoryTraits>;
+using DeviceView1d = Kokkos::View<Scalar*, Kokkos::LayoutLeft, PHX::Device, MemoryTraits>;
 template<typename Scalar, typename MemoryTraits = Kokkos::MemoryUnmanaged>
-using DeviceView2d = Kokkos::View<Scalar**, DevLayout, PHX::Device, MemoryTraits>;
+using DeviceView2d = Kokkos::View<Scalar**, Kokkos::LayoutLeft, PHX::Device, MemoryTraits>;
 
 // Kokkos types for local graphs/matrices, to be used for on-device kernels
 using DeviceLocalGraph  = Kokkos::StaticCrsGraph<LO, Kokkos::LayoutLeft, KokkosNode::device_type, void, size_t>;
--- a/src/Albany_KokkosTypes.hpp
+++ b/src/Albany_KokkosTypes.hpp
@@ -39,11 +39,19 @@ using ViewLR = Kokkos::View<DT,Kokkos::LayoutRight,MemSpace>;
 // NOTE: Tpetra may use a different LO type (Albany uses int32, while tpetra uses int). When extracting local views/matrices,
 //       be careful about this. At worst, you may need to extract pointers and reinterpret_cast them.
 
+using DevLayout = PHX::Device::array_layout;
+
 // kokkos 1d and 2d views to be used for on-device kernels
 template<typename Scalar, typename MemoryTraits = Kokkos::MemoryUnmanaged>
-using DeviceView1d = Kokkos::View<Scalar*, Kokkos::LayoutLeft, PHX::Device, MemoryTraits>;
+using DeviceView1d = Kokkos::View<Scalar*, DevLayout, PHX::Device, MemoryTraits>;
 template<typename Scalar, typename MemoryTraits = Kokkos::MemoryUnmanaged>
-using DeviceView2d = Kokkos::View<Scalar**, Kokkos::LayoutLeft, PHX::Device, MemoryTraits>;
+using DeviceView2d = Kokkos::View<Scalar**, DevLayout, PHX::Device, MemoryTraits>;
+
+// Thyra view types for underlying Tpetra kokkos views
+template<typename Scalar>
+using ThyraVDeviceView = Kokkos::View<Scalar*, Kokkos::LayoutLeft, PHX::Device>;
+template<typename Scalar>
+using ThyraMVDeviceView = Kokkos::View<Scalar**, Kokkos::LayoutLeft, PHX::Device>;
 
 // Kokkos types for local graphs/matrices, to be used for on-device kernels
 using DeviceLocalGraph  = Kokkos::StaticCrsGraph<LO, Kokkos::LayoutLeft, KokkosNode::device_type, void, size_t>;
--- a/src/evaluators/gather/PHAL_GatherScalarNodalParameter_Def.hpp
+++ b/src/evaluators/gather/PHAL_GatherScalarNodalParameter_Def.hpp
@@ -80,17 +80,13 @@ evaluateFields(typename Traits::EvalData workset)
   const auto elem_lids  = workset.disc->getWsElementLIDs();
   const auto elem_lids_dev = Kokkos::subview(elem_lids.dev(),ws,ALL);
 
-  const int nnodes = this->numNodes;
-
-  auto& val = this->val;
-
   Kokkos::parallel_for(RangePolicy(0,workset.numCells),
-                       KOKKOS_LAMBDA(const int& cell) {
+                       KOKKOS_CLASS_LAMBDA(const int& cell) {
     const auto elem_LID = elem_lids_dev(cell);
     const auto dof_lids = Kokkos::subview(p_elem_dof_lids,elem_LID,ALL);
-    for (int node=0; node<nnodes; ++node) {
+    for (int node=0; node<this->numNodes; ++node) {
       const LO lid = dof_lids(node);
-      val(cell,node) = lid>=0 ? p_data(lid) : 0;
+      this->val(cell,node) = lid>=0 ? p_data(lid) : 0;
     }
   });
 }
@@ -466,7 +462,6 @@ evaluateFields(typename Traits::EvalData workset)
   Albany::DeviceView1d<const ST> vvec_data;
   if (is_p_direction_active) vvec_data = Albany::getDeviceData(vvec->col(0).getConst());
 
-  const int num_nodes = this->numNodes;
   const int ws = workset.wsIndex;
 
   // Parameter/nodes dof numbering info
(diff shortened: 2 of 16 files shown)


src/evaluators/gather/PHAL_GatherScalarNodalParameter_Def.hpp
```diff
@@ -70,23 +70,29 @@ evaluateFields(typename Traits::EvalData workset)
...
   const auto ws = workset.wsIndex;
-  const auto elem_lids = workset.disc->getElementLIDs_host(ws);
+  const auto elem_lids  = workset.disc->getWsElementLIDs();
+  const auto elem_lids_dev = Kokkos::subview(elem_lids.dev(),ws,ALL);
 
-  for (std::size_t cell = 0; cell<workset.numCells; ++cell) {
-    const auto elem_LID = elem_lids(cell);
+  const int nnodes = this->numNodes;
+
+  auto& val = this->val;
```
bartgol, line 85: Does the ref work here? A ref is like a pointer, so I expect `KOKKOS_LAMBDA` to not work correctly on CUDA... I would just use a copy here, since it's a shallow copy anyways.

src/evaluators/gather/PHAL_GatherScalarNodalParameter_Def.hpp
```diff
@@ -457,38 +463,43 @@ evaluateFields(typename Traits::EvalData workset)
...
   const int ws = workset.wsIndex;
 
   // Parameter/nodes dof numbering info
-  const auto p_elem_dof_lids = p->get_dof_mgr()->elem_dof_lids().host();
-  const auto elem_lids    = workset.disc->getElementLIDs_host(ws);
+  const auto p_elem_dof_lids = p->get_dof_mgr()->elem_dof_lids().dev();
+  const auto elem_lids  = workset.disc->getWsElementLIDs();
+  const auto elem_lids_dev = Kokkos::subview(elem_lids.dev(),ws,ALL);
+
+  auto& val = this->val;
```
bartgol, line 477: Here too.

bartgol (approved):
Looks good. I have only one comment on a reference usage.

src/Albany_KokkosTypes.hpp
```diff
@@ -39,13 +39,11 @@ using ViewLR = Kokkos::View<DT,Kokkos::LayoutRight,MemSpace>;
...
 
-using DevLayout = PHX::Device::array_layout;
-
 // kokkos 1d and 2d views to be used for on-device kernels
 template<typename Scalar, typename MemoryTraits = Kokkos::MemoryUnmanaged>
-using DeviceView1d = Kokkos::View<Scalar*, DevLayout, PHX::Device, MemoryTraits>;
+using DeviceView1d = Kokkos::View<Scalar*, Kokkos::LayoutLeft, PHX::Device, MemoryTraits>;
 template<typename Scalar, typename MemoryTraits = Kokkos::MemoryUnmanaged>
-using DeviceView2d = Kokkos::View<Scalar**, DevLayout, PHX::Device, MemoryTraits>;
+using DeviceView2d = Kokkos::View<Scalar**, Kokkos::LayoutLeft, PHX::Device, MemoryTraits>;
```
jewatkins, line 46: Does this layout get used on host in a serial build? I think that would affect cpu performance.

src/evaluators/gather/PHAL_GatherSolution_Def.hpp
```diff
@@ -485,67 +485,75 @@ evaluateFields(typename Traits::EvalData workset)
...
-    const auto dof_lids = Kokkos::subview(elem_dof_lids,elem_LID,ALL);
-    for (int node = 0; node < this->numNodes; ++node) {
-      for (int eq = 0; eq < numFields; ++eq) {
+  auto nnodes = this->numNodes;
+  auto nfields = this->numFields;
+  auto sol = this->device_sol;
+  auto ncolsx = workset.num_cols_x;
+  auto jcoeff = workset.j_coeff;
+  auto mcoeff = workset.m_coeff;
+  auto ncoeff = workset.n_coeff;
```
jewatkins, line 519: `const auto` for these.

src/evaluators/gather/PHAL_GatherSolution_Def.hpp
```diff
@@ -692,29 +700,35 @@ evaluateFields(typename Traits::EvalData workset)
...
 
-  const auto& fields_offsets = m_fields_offsets.host();
+  const auto& fields_offsets = m_fields_offsets.dev();
   const auto& first_dof = this->offset;
-  for (std::size_t cell=0; cell < workset.numCells; ++cell ) {
-    const auto elem_LID = elem_lids(ws,cell);
+
+  const auto elem_lids_dev = Kokkos::subview(elem_lids.dev(),ws,ALL);
+  auto nnodes = this->numNodes;
+  auto nfields = this->numFields;
```
jewatkins, line 720: `const` here too

src/evaluators/scatter/PHAL_ScatterResidual_Def.hpp
```diff
@@ -355,43 +355,60 @@ evaluateFields(typename Traits::EvalData workset)
...
+
+  const auto& fields_offsets = m_fields_offsets.dev();
   const auto eq_offset = this->offset;
-  for (size_t cell=0; cell<workset.numCells; ++cell) {
+  auto nnodes = this->numNodes;
+  auto nfields = this->numFields;
+  auto resid = this->device_resid;
+  auto ncolsx = workset.num_cols_x;
+  auto ncolsp = workset.num_cols_p;
+  auto paramoffset = workset.param_offset;
```
jewatkins, line 384: `const`

src/evaluators/scatter/PHAL_ScatterResidual_Def.hpp
```diff
@@ -355,43 +355,60 @@ evaluateFields(typename Traits::EvalData workset)
...
-            fp_data[col][lid] += res.dx(col + workset.param_offset);
+          for (int col=0; col<ncolsp; ++col) {
+            KU::atomic_add<ExecutionSpace>(&fp_data(lid,col), res.dx(col + paramoffset));
           }
         }
       }
     }
-  }
+  });
+  cudaCheckError();
```
jewatkins, line 411: This might be a no-op now? I don't think it's needed anymore now that we have kokkos-tools.

jewatkins (changes requested):
looks good, thanks! Just a few things

src/evaluators/gather/PHAL_GatherScalarNodalParameter_Def.hpp
mcarlson801: So this worked on Perlmutter and Weaver without warning but I'm not against changing this. Should it just be `const auto` instead?

src/Albany_KokkosTypes.hpp
mcarlson801: Not sure how this would work with `getNonconstDeviceData` since that's basically just grabbing a write-accessible view. We'd have to copy the data in and then copy the data back at some point with another call.

src/evaluators/gather/PHAL_GatherScalarNodalParameter_Def.hpp
jewatkins: Actually, I think you can use `KOKKOS_CLASS_LAMBDA` and you won't have to do these copies at all. Just `auto` is probably best for a view but I think `const auto` is okay too in this case because you're not modifying the view, just the values in the view.

src/Albany_KokkosTypes.hpp
jewatkins: Ah that's weird that left layout is used always. Since the phalanx and thyra layouts are different, I think we should distinguish them here too. Is `DeviceView1d`/2d used only for thyra views? If it is, I would name this `ThyraDeviceView 1d` and grab the layout directly from Thyra.

bartgol: I had the same thought. But then realized that Tpetra uses exclusively LayoutLeft. I agree that there may be some confusion though, so something along the line of what Jerry suggested may help.

jewatkins: I dug a bit more and I think these layouts should be the same. It seems like the get routines return a tpetra view which is defined by `KokkosNode`: https://github.com/sandialabs/Albany/blob/master/src/Albany_KokkosTypes.hpp#L29
which use `PHX::Device`. Maybe we can discuss the issue you're seeing offline, I don't think this should be LayoutLeft all the time.

src/evaluators/gather/PHAL_GatherScalarNodalParameter_Def.hpp
bartgol: Yes, we can use KOKKOS_CLASS_LAMBDA, now that C++17 is mandatory in Trilinos.

src/Albany_KokkosTypes.hpp
jewatkins: a lot of this seems overdefined too. shouldn't we be able to define these with just `Kokkos::View<Scalar*, PHX::Device>`?

src/evaluators/gather/PHAL_GatherScalarNodalParameter_Def.hpp
mcarlson801: Okay, I'll use KOKKOS_CLASS_LAMBDA. I think that should let me clean up some of the other variables that are just pointing to class members too.

src/Albany_KokkosTypes.hpp
bartgol: The Tpetra MV's use LayoutLeft to ensure that enries of the same column are contiguous, which ensures we can extract a col from an MV, and get a contiguous array.

jewatkins: ah okay I remember that now. It looks like `DeviceView` is getting used for other things so we should definitely make a distinction to avoid others using it for anything other than tpetra MV.

jewatkins: This is probably what we should be using in that case: https://github.com/trilinos/Trilinos/blob/master/packages/tpetra/core/src/Tpetra_MultiVector_decl.hpp#L473
via Tpetra_MultiVector::device_view_type

mcarlson801: I found this when looking at the Tpetra docs:

https://docs.trilinos.org/dev/packages/tpetra/doc/html/classTpetra_1_1MultiVector.html#ae9efb48f9e44a6376862b4d5e436ee87

If I understand this correctly, the underlying dual view of a Tpetra_Multivector will always have LayoutLeft.

jewatkins: right, you can see it in the link I sent as well. I would make the generic `DeviceView = Kokkos::View<Scalar*, PHX::Device>` and create a new definition for the unique `TpetraMVDeviceView = Tpetra_MultiVector::device_view_type`

jewatkins: or `ThyraMVDeviceView` since we still use thyra.

mcarlson801: > right, you can see it in the link I sent as well

Ah sorry, I should have hit refresh before I commented. I didn't see all the follow up comments

jewatkins (approved):
Thanks!

mperego:
Thanks @mcarlson801!

Issue #925, opened 2023-04-10
PyAlbany: known bug when writing (multi)vectors to files with unconsistent ordering


kliegeois:
Since 356f67b, @mperego and I observed a issue when writing (multi)vectors to files when the ordering of the local IDs is not the same as the ordering of the global IDs.

kliegeois:
@mperego can we close this issue?

mperego:
Yes, let's close it.

PR #1033, opened 2024-03-21
Added some Tpetra Matrix support to PyAlbany

mperego:
Also revisited the functions to load and write MultiVectors from/to files. There was an issue with serial (1 MPI rank) runs, where the vector was not exported using the Map. I also tried to simplified them a bit removing some capabilities that I don't think are used. @kliegeois , let me know if it's OK otherwise I can add them back.

Still need to add a test that uses Tpetra Matrices.


mperego:
I think this PR fixes #925.

kliegeois (approved):
Everything looks great to me, thanks @mperego !
I like the simplifications of the logic inside the IO functions.

mperego:
Thanks for the review, @kliegeois !

PR #1035, opened 2024-03-22
Remove Albany timers and add kernel names

jewatkins:
We should use kokkos-tools instead. This also fixes a warning about non virtual destructor.


mcarlson801 (approved):
This looks good to me, thanks @jewatkins !

PR #1036, opened 2024-03-26
Fix bug related to Kokkos::vector replacement

mcarlson801:
This PR fixes a bug that only shows up in debug builds and was introduced by the recent changes to replace Kokkos::vector. DualView's `resize` sets the modify flag on device and then a later call to `modify_host` causes the modify flag to be true on both host and device. With a debug build, this situation causes an exception to be thrown.

This fix adds a sync after the resize to clear the modify flags. Looking at the implementation of DualView's `resize`, it looks like there might be situations where the host modify flag is set instead. Therefore, I've added syncs for host and device just in case. It's not an elegant solution but whichever one is unnecessary will be a no-op.

### DIFF
--- a/src/evaluators/gather/PHAL_GatherSolution_Def.hpp
+++ b/src/evaluators/gather/PHAL_GatherSolution_Def.hpp
@@ -138,10 +138,16 @@ GatherSolutionBase(const Teuchos::ParameterList& p,
 
   if ( tensorRank == 0 ) {
     device_sol.val_kokkos.resize(numFields);
+    device_sol.val_kokkos.sync_host();
+    device_sol.val_kokkos.sync_device();
     if (enableTransient)
       device_sol.val_dot_kokkos.resize(numFields);
+      device_sol.val_dot_kokkos.sync_host();
+      device_sol.val_dot_kokkos.sync_device();
     if (enableAcceleration)
       device_sol.val_dotdot_kokkos.resize(numFields);
+      device_sol.val_dotdot_kokkos.sync_host();
+      device_sol.val_dotdot_kokkos.sync_device();
   }
 
   if (p.isType<int>("Offset of First DOF"))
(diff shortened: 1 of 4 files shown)


src/evaluators/gather/PHAL_GatherSolution_Def.hpp
```diff
@@ -138,10 +138,16 @@ GatherSolutionBase(const Teuchos::ParameterList& p,
 
   if ( tensorRank == 0 ) {
     device_sol.val_kokkos.resize(numFields);
+    device_sol.val_kokkos.sync_host();
+    device_sol.val_kokkos.sync_device();
```
bartgol, line 142: Why do you need to sync twice? Isn't `sync_host` doing "copy device to host", and hence ending with a fully synced view?

bartgol (commented):
I have a question on the double sync calls.

Also, I noticed that the typedef is still `using KV = ...`, which is a bit confusing since KV is not a vector but a DualView. Would you mind chaning that to something like `using DV = ...` or `using DualView = ...`, which is a bit more self-documenting?

mcarlson801:
> Also, I noticed that the typedef is still `using KV = ...`, which is a bit confusing since KV is not a vector but a DualView. Would you mind chaning that to something like `using DV = ...` or `using DualView = ...`, which is a bit more self-documenting?

Good call, I'll go ahead and update that.

src/evaluators/gather/PHAL_GatherSolution_Def.hpp
mcarlson801: If we could guarantee that the result of `resize` sets the modify flag on the device side, that would be all that is needed.

I looked at the implementation (https://github.com/kokkos/kokkos/blob/master/containers/src/Kokkos_DualView.hpp#L1092) and it seems like it is possible that in certain configurations, the modify flag could get set on host instead. Although maybe in that case, there is no device to sync to?

bartgol: Ok, that's fine. This is not that crucial anyways.

bartgol (approved):

Issue #1038, opened 2024-03-28
PyAlbany MatrixOperations test failing in spack nightly


ikalash:
In the Spack nightly with PyAlbany, the MatrixOperations test is failing with the following error: 

```
135/156 Test #135: PyAlbany_MatrixOperations .............................................   Passed    0.83 sec
test 136
        Start 136: PyAlbany_MatrixOperations_MPI

136: Test command: /nightlySpackBuild/spack/opt/spack/linux-rhel8-haswell/gcc-11.1.0/openmpi-4.1.6-knz4yflnjpyldplqsv2ljcm4sxqejcbd/bin/mpiexec "-np" "2" "/nightlySpackBuild/spack/opt/spack/linux-rhel8-haswell/gcc-11.1.0/python-3.9.18-cmndxf3ue5nncpqpdojfnfvvmd5qvmdb/bin/python3.9" "/tmp/ikalash/spack-stage/spack-stage-albany-develop-ycy2yfkfsfl46rpqbsaol4mebadmegnt/spack-build-ycy2yfk/pyAlbany/tests/linearAlgebra/matrixOperations.py"
136: Working Directory: /tmp/ikalash/spack-stage/spack-stage-albany-develop-ycy2yfkfsfl46rpqbsaol4mebadmegnt/spack-build-ycy2yfk/pyAlbany/tests/linearAlgebra
136: Test timeout computed to be: 1200
136: Traceback (most recent call last):
136:   File "/tmp/ikalash/spack-stage/spack-stage-albany-develop-ycy2yfkfsfl46rpqbsaol4mebadmegnt/spack-build-ycy2yfk/pyAlbany/tests/linearAlgebra/matrixOperations.py", line 2, in <module>
136:     import numpy as np
136: ModuleNotFoundError: No module named 'numpy'
136: Traceback (most recent call last):
136:   File "/tmp/ikalash/spack-stage/spack-stage-albany-develop-ycy2yfkfsfl46rpqbsaol4mebadmegnt/spack-build-ycy2yfk/pyAlbany/tests/linearAlgebra/matrixOperations.py", line 2, in <module>
136:     import numpy as np
136: ModuleNotFoundError: No module named 'numpy'
```
The error is surprising b/c there are other PyAlbany tests that use numpy and they are passing.  @kliegeois do you have any insight into this?

kliegeois:
@mperego this is related to the recently added test in your PR is that right ?

mperego:
[...]

PR #1039, opened 2024-04-08
Fix PyAlbany_MatrixOperations_MPI

kliegeois:
This PR should fix our nightly issue: #1038

The environment property of the MPI version of the test was not set correctly and therefore the test was unable to find the required Python libraries.

@mperego @ikalash


mperego (approved):
Thanks @kliegeois

PR #1042, opened 2024-04-26
Store Omega_h::Mesh objects via pointers

bartgol:
While trying to hack the code to get an output file from an Omega_h test, I realized that we were copying around Omega_h::Mesh objects, but that's not safe, since they are not fully shallow objects. E.g., adding a tag to one copy of the mesh will not add it to the original mesh object.

This PR stores the raw Omega_h::Mesh in our wrappers as an RCP, and passes it around as an RCP. With this, I was able to save the mesh in vtk format, and visualize the solution on paraview.

Note: I am not committing the code that saves the mesh, since it was a bit of a hack. In particular, I haven't figured out how to add multiple time slices of fields. I will tackle that another time.


cwsmith (approved):
Good catch.  LGTM.  Thank you.

PR #1046, opened 2024-05-06
PyAlbany: upgrade pybind11 and replace assertTrue

kliegeois:
This PR upgrades Pybind11 version that was getting older and replaces `assertTrue` with the assertion corresponding to the used comparisons.


mperego (approved):
It looks good to me. This is a nice improvement. Thanks!

PR #1049, opened 2024-05-17
Port responses to kokkos for uvm-free access

mcarlson801:
This PR updates `ResponseGLFlux` and `ResponseSurfaceVelocityMismatch` so that they can be used in uvm-off GPU builds. This includes porting a number of evaluators for scattering and communicating responses to kokkos for device-friendly access.

Additionally, this PR also fixes some existing kokkos ported evaluators to avoid some unintended host accesses that showed up in a uvm-off build. I've also added evaluator names to as many of the `parallel_for` regions as I could find to aid in performance profiling and debugging.

After this PR is merged, the following tests will now be passing in a uvm-off build:

```
landIce_FO_GIS_ForwardSensitivity
landIce_FO_GIS_Unstructured
landIce_FO_GIS_Unstructured_Restart
landIce_FO_GIS_Unstructured_Memoization
landIce_FO_GIS_Unstructured_FROSch
landIce_FO_GIS_Manifold
landIce_FO_MMS_FO_GLF_TRI_CORNER_CASE
landIce_FO_MMS_FO_GLF_QUAD
landIce_FO_MMS_FO_GLF_TRI_NONCONST_VELOCITY
landIce_FO_MMS_FO_GLF_TRI
landIce_FO_Thermo_Dry_Bed
landIce_FO_Thermo_Wet_Bed
landIce_FO_Thermo_Wet_Bed_FEA
landIce_FO_Thermo_Wet_Bed_FROSch
landIce_Enthalpy_Humboldt
landIce_Enthalpy_Kleiner_A
landIce_Enthalpy_Kleiner_B
```

The next target for the uvm-free porting process is to port `L2ProjectedBoundaryLaplacian` to kokkos which will get a large number of additional tests to start passing. (I think this will get most everything remaining in FO_GIS to pass)

Finally, I've re-enabled all of the cuda tests for the uvm-free build and I will instead start an issue to track the status of which tests and expected to pass/fail with uvm disabled.

### DIFF
--- a/src/PHAL_Utilities.cpp
+++ b/src/PHAL_Utilities.cpp
@@ -107,37 +107,6 @@ int getDerivativeDimensions(const Albany::Application* app, const int ebi)
 
 namespace {
 template<typename ScalarT>
-struct A2V {
-  std::vector<ScalarT>& v;
-  A2V (std::vector<ScalarT>& v) : v(v) {}
-  void operator() (typename Ref<const ScalarT>::type a, const int i) {
-    v[i] = a;
-  }
-};
-
-template<typename ScalarT>
-struct V2A {
-  const std::vector<ScalarT>& v;
-  V2A (const std::vector<ScalarT>& v) : v(v) {}
-  void operator() (typename Ref<ScalarT>::type a, const int i) {
-    a = v[i];
-  }
-};
-
-template<typename ScalarT>
-void copy (const PHX::MDField<ScalarT>& a, std::vector<ScalarT>& v) {
-  v.resize(a.size());
-  A2V<ScalarT> a2v(v);
-  loop(a2v, a);
-}
-
-template<typename ScalarT>
-void copy (const std::vector<ScalarT>& v, PHX::MDField<ScalarT>& a) {
-  V2A<ScalarT> v2a(v);
-  loop(v2a, a);
-}
-
-template<typename ScalarT>
 void myReduceAll (
   const Teuchos_Comm& comm, const Teuchos::EReductionType reduct_type,
   std::vector<ScalarT>& v)
@@ -212,15 +181,6 @@ void reduceAll (
   ScalarT b = a;
   Teuchos::reduceAll(comm, reduct_type, 1, &a, &b);
   a = b;
-}
-
-template<typename ScalarT>
-void broadcast (const Teuchos_Comm& comm, const int root_rank,
-                PHX::MDField<ScalarT>& a) {
-  std::vector<ScalarT> v;
-  copy<ScalarT>(a, v);
-  Teuchos::broadcast<int, ScalarT>(comm, root_rank, v.size(), &v[0]);
-  copy<ScalarT>(v, a);
 }
 
 template int getDerivativeDimensions<PHAL::AlbanyTraits::Jacobian>(
--- a/src/PHAL_Utilities.cpp
+++ b/src/PHAL_Utilities.cpp
@@ -186,6 +186,11 @@ void reduceAll (
   const Teuchos_Comm& comm, const Teuchos::EReductionType reduct_type,
   PHX::MDField<ScalarT>& a)
 {
+  // Old reduceAll implementation is not uvm-friendly and untested
+  //  for rank > 1
+  TEUCHOS_TEST_FOR_EXCEPTION(a.rank() > 1, std::logic_error,
+        "PHAL::reduceAll not implemented for MDFields with rank > 1.\n");
+        
   MDFieldHostMirror<ScalarT> a_host = Kokkos::create_mirror_view(a.get_view());
   Kokkos::deep_copy(a_host, a.get_view());
   std::vector<ScalarT> v;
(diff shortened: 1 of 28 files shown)


src/PHAL_Utilities.hpp
```diff
@@ -94,18 +94,23 @@ class MDFieldIterator {
 template<typename T>
 class MDFieldVectorRight {
 public:
+  using array_type = typename PHX::MDField<T>::array_type;
```
bartgol, line 97: This seems unused.

src/landIce/evaluators/LandIce_ResponseSurfaceVelocityMismatch_Def.hpp
```diff
@@ -239,21 +255,23 @@ void LandIce::ResponseSurfaceVelocityMismatch<EvalT, Traits>::evaluateFields(typ
...
             t += sum * w_measure(sideSet_idx,qp);
           }
-          this->local_response_eval(cell, 0) += t*scaling*alpha;//*50.0;
-          this->global_response_eval(0) += t*scaling*alpha;//*50.0;
-          p_reg += t*scaling*alpha;
-        }
+
+          ScalarT result = t*scaling*alpha;
+          KU::atomic_add<ExecutionSpace>(&(this->local_response_eval(cell, 0)), result);//*50.0
+          KU::atomic_add<ExecutionSpace>(&(this->global_response_eval(0)), result);//*50.0
```
bartgol, line 261: Would it make sense to move the update of the global resp and p_reg to outside the kernel? Like
```c++
// kernel to update local_resp(cell,0)
parellel_reduce ( ...); // ||reduce to reduce local resp into global resp
deep_copy(p_ref, g_resp);
```
I'm not sure if atomic adds are going to be noticeable though. If not, then maybe it's ok to leave as is.

bartgol (approved):
Looks great!

src/landIce/evaluators/LandIce_ResponseSurfaceVelocityMismatch_Def.hpp
mcarlson801: Yeah, this kernel is pretty tiny in terms of overall runtime (and is faster than what we have now) so I figured it's probably fine.

src/PHAL_Utilities.hpp
mcarlson801: Ah actually, looks like it is used in the return type on the next line.

bartgol: Jeez...time to up my prescription glasses...

src/evaluators/scatter/PHAL_ScatterScalarNodalParameter_Def.hpp
```diff
@@ -86,22 +86,27 @@ evaluateFields(typename Traits::EvalData workset)
   if (this->memoizer.have_saved_data(workset,this->evaluatedFields()))
     return;
 
+  constexpr auto ALL = Kokkos::ALL();
```
jewatkins, line 89: I would rename this. It might get confused with a macro.

jewatkins, line 89: I would rename this. It might get confused with a macro.

src/PHAL_Utilities.cpp
```diff
@@ -186,10 +186,22 @@ void reduceAll (
...
-  copy<ScalarT>(a, v);
+  for (size_t i = 0; i < a.size(); ++i) {
+    v.push_back(a_host(i));
+  }
   myReduceAll<ScalarT>(comm, reduct_type, v);
-  copy<ScalarT>(v, a);
+  for (size_t i = 0; i < a.size(); ++i) {
+    a_host(i) = v[i];
+  }
+  Kokkos::deep_copy(a.get_view(), a_host);
```
jewatkins, line 204: Is this only used for size == 1 too (global response)? If yes, I think we can remove/simplify a lot of this.

jewatkins (changes requested):
thanks Max!

src/PHAL_Utilities.cpp
mcarlson801: Good call, I removed the relevant code that is no longer used or tested.

jewatkins (approved):
thanks!

src/evaluators/interpolation/PHAL_DOFGradInterpolation_Def.hpp
```diff
@@ -141,6 +141,7 @@ template< typename Traits>
 KOKKOS_INLINE_FUNCTION
 void FastSolutionGradInterpolationBase<PHAL::AlbanyTraits::Jacobian, Traits, typename PHAL::AlbanyTraits::Jacobian::ScalarT>::
 operator() (const FastSolutionGradInterpolationBase_Jacobian_Tag& tag, const int& cell) const {
+  const int num_dof = this->val_node(0,0).size();
   for (size_t qp=0; qp < this->numQPs; ++qp) {
     for (size_t dim=0; dim<this->numDims; dim++) {
       this->grad_val_qp(cell,qp,dim) = ScalarT(num_dof, this->val_node(cell, 0).val() * this->GradBF(cell, 0, qp, dim));
```
mperego, line 147: We can probably optimize this similarly to how Oscar did it in the other PR. This can be done in in a subsequent PR.

jewatkins: Oscar is planning to look at other kernels starting with the most expensive ones showing up in the profiler. I think this is in the top 10.

PR #1050, opened 2024-05-18
Adding logic to over-write ALBANY_PARALLEL_EXODUS  if desired.

ikalash:
Needed for Attaway nightly tests.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -149,6 +149,7 @@ include(CTest)
 MESSAGE("Checking Trilinos build for required and optional packages")
 
 OPTION(ENABLE_ALBANY_PYTHON "Flag to turn on building of code depending on Python" OFF)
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -431,6 +438,8 @@ ELSE()
   MESSAGE("-- ALBANY_PARALELL_EXODUS set to False")
 ENDIF()
 
+SET(ALBANY_PARALELL_EXODUS FALSE)
```
mperego, line 441: This line is setting `ALBANY_PARALELL_EXODUS` support to false in all cases. I think it should be removed.

ikalash: Good catch!  I had this in there initially for testing purposes and forgot to remove it!  I just checked in the corrected CMakeLists.txt file.

mperego (approved):

PR #1052, opened 2024-05-23
Add generic infrastructure to handle extruded meshes

bartgol:
This PR adds some infrastructure to handle extruded meshes. In particular, it adds a `ExtrudedMesh` class, and a `ExtrudedConnManager` class. Neither of these two classes stores 3d information. Instead, they store the basal mesh/conn-mgr information, along with information about the vertical extrusion. At runtime, they combine these two to provide full 3d data.

I added initial testing for the ExtrudedConnManager, but more needs to be added. The conn manager has some baked in assumptions, including

- numbering of dofs is done layer-by-layer or column-by-column, depending on the structure of the layered mesh
- all fields in the FieldPattern aggregate passed to buildConnectivity have the *same* field pattern.

We may be able to remove the second assumption, but we would have to decide how to order dofs in that case.


bartgol:
@mperego I am thinking of integrating this as is for now (assuming you have no objections), and then proceed to add more complex tests later (e.g., a FO_GIS test that uses this instead of ExtrudedSTKMeshStruct). Let me know if you disagree.

mperego:
@bartgol did you tested this with MALI?

bartgol:
> @bartgol did you test this with MALI?

Not yet. I don't expect any issue though, since it's not used anywhere except the unit tests though..

bartgol:
@mperego I'm closing this PR, since I found out several issues (which have been fixed) while working on an ExtrudedDiscretization class. Adding together the two pieces of work makes it for a big PR. When I get to a point where things work, I'll figure out a way to break it into smaller PRs, so that we can review things more easily.

PR #1058, opened 2024-06-26
Fix handling of side disc when saving sol in STKDiscretization

bartgol:
The side discretizations were only processed in writeSolutionToFile, and not in writeSolutionToMeshDatabase (which is were the stk in-memory mesh data is updated). As a result, exo files for side discretizations had zero solution.

This PR fixes that, moving the handling of side discretizations into the main method (writeSolution), which calls the two methods mentioned above. Also, I got rid of some 2-liner methods in STKDiscretization, inlining them at the call site.


mperego (approved):
It looks good to me.
Thanks for doing this. I run into the issue of zero solution on the boundary before!

PR #1059, opened 2024-06-27
Miscellanea small fixes/cleanups in disc folder

bartgol:
Small summary

- clean up unused code/files
- clean up disc factory interface
- add mesh type enum, in view of adding ExtrudedMesh generic interface (coming in a follow-up PR)
- fix bug in STKConnManager: should not affect existing code, and was exposed only by the upcoming Extruded mesh work. But I figured I added it to this PR, since it's a bugfix.
- move impl of common conn mgr methods in base class (reduces code duplication)


src/disc/Albany_DiscretizationFactory.hpp
```diff
@@ -63,29 +63,20 @@ class DiscretizationFactory {
...
-      const int neq,
-      const std::map<int,std::vector<std::string> >& sideSetEquations,
-      const Teuchos::RCP<RigidBodyModes>& rigidBodyModes);
-    
     /* This function overwrite previous discretization parameter list */
     void
     setDiscretizationParameters(Teuchos::RCP<Teuchos::ParameterList> disc_params);
     
   protected:
+    Teuchos::RCP<AbstractDiscretization> createDiscretizationFromMeshStruct(
```
mperego, line 71: Maybe add a brief description of this function?

src/disc/stk/STKConnManager.cpp
```diff
@@ -126,7 +120,7 @@ STKConnManager::getConnectivityMask (const std::string& sub_part_name) const
         int count = m_bulkData->num_connectivity(e,rank);
         for (int irel=0; irel<count; ++irel) {
           for (int cnt=0; cnt<m_idCnt[rank]; ++cnt) {
-            elem_mask[offsets[rank] + cnt] = 1;
+            elem_mask[offsets[rank] + irel*m_idCnt[rank] + cnt] = 1;
```
mperego, line 123: Was this a bug?

mperego (approved):
It looks good to me. I left a couple of comments.

src/disc/stk/STKConnManager.cpp
bartgol: Yes, but only showed up for a particular case. Namely, when we restricted to a sideset. Currently, that's never the case: we only call getConnectivityMask when restricting the DOF mgr to a nodeset (e.g., bottom for extruded meshes). However, when building an extruded discretization (coming in a follow up PR), I needed to restrict to a side set, and started getting seg faults.

Issue #1054, opened 2024-06-17
Failing landIce_FO_GIS_AdjointSensitivity_Thickness test in nightly spack build


ikalash:
The ```landIce_FO_GIS_AdjointSensitivity_Thickness``` test has been failing in the nightly spack build.  This started a few weeks ago, but I did not get a chance to look at it until today due to travel.

The attachment contains output from the test.  Why is so much info about the MDFields dumped?  That makes the output very difficult to parse through.  Curiously, it looks like none of the regular nightlies are running this test, just the spack one.  Is there a particular configure option that triggers the test to be turned on?

@mperego @bartgol

ikalash:
Thanks for looking at it, @bartgol .  Yes, it looks like ```ENABLE_MESH_DEPENDS_ON_PARAMETERS``` is only turned on in the Intel CEE build: https://github.com/sandialabs/Albany/blob/master/doc/dashboards/cee-compute011.sandia.gov/ctest_nightly_tmp.cmake#L489-L518 .  It's odd to me that it's not on in the others.  I can certainly turn it on in more nightlies - I will do this.  Let me know if there are any specific ones you would like it enabled in.  The Spack build does not push to the CDash site, so that is a separate thing.  I bet if the option is enabled, we will see the same failure as in the spack build in the nightlies.  We can see tomorrow.

Unfortunately the Intel build has been broken due to another issue, and right now the CDash site it down so I cannot check the history of this test in that build.  I will try again later today.

bartgol:
The only drawback of that option is to make the build/run time longer. So I would maybe enable it on machines that are not already slow, to make sure we get a report in a reasonable time...

bartgol:
Looking at the git history, there are two things that happened on the 23rd, which could affect testing:
- PR #1051 was merged
- you pushed a couple of commits to master, related to ceecompute dashboards.

It's hard to believe any of those had an impact on the pass/fail of the test. We can look at the test log of the last passing test, and compare against the current log, to see if any of the parameters has somehow changed. But given the date you pinpointed, I doubt we'll see a difference. Maybe something changed in Trilinos? Like a default value in Piro or some other solver package? @mperego ?

ikalash:
Ok I can try to do bisection to figure out what broke things on 5/23.  Please stay tuned.

ikalash:
It is the commits on 5/23 made by @mcarlson801 that are the culprit.  I checked out the following shas:

[...]

Issue #1060, opened 2024-06-27
Failing landIce_FO_MMS_FO tests on some platforms


ikalash:
It looks like the following tests are failing in some of the nighties (camobap and weaver).

landIce_FO_MMS_FO_GLF_QUAD
landIce_FO_MMS_FO_GLF_TRI
landIce_FO_MMS_FO_GLF_TRI_CORNER_CASE
landIce_FO_MMS_FO_GLF_TRI_NONCONST_VELOCITY

https://sems-cdash-son.sandia.gov/cdash/viewTest.php?onlyfailed&buildid=104669

It looks like comparisons are failing, but the comparison is for a very large value, which seems fishy.

Tagging @mperego and @bartgol since you likely know the most about these.

bartgol:
@mcarlson801 longshot: could this be a similar issue to the one you fixed in some responses a caouple of days ago? I recall that issue had very large diffs in the response/sensitivities.

mcarlson801:
Ah, I didn't see this had an issue opened for it. Yes, these are the tests I mentioned in the other issue that started failing after I fixed an issue with the GLFlux evaluator. I'm going to revert all my changes to this evaluator to see if it fixes and then debug from there.

mcarlson801:
I reverted the uvm-free stuff in ResponseGLFlux and the tests are now passing on my local build on weaver. Looking at the diff, it's not really clear to me what could be causing the issue though. Nightlies should be clean at least while I debug.

ikalash:
Thanks for working on this @mcarlson801 !

ikalash:
These are back to normal, I can confirm.  Should I close this, or keep it open?  It seemed like you might still be investigating the root cause @mcarlson801 .

mcarlson801:
This should be fixed now. If no FO_GIS tests start failing tomorrow, we can close this.

mcarlson801:
These tests are passing now, closing this and moving to the AdjointSensitivity_Thickness issue.

PR #1064, opened 2024-07-11
Fix for LandIce_ResponseGLFlux when using DistParamDeriv

mcarlson801:
This PR fixes issues #1054 and #1060. Looks like `coeff` needs to be a ScalarT instead of a double in order to get the correct derivative values in DistParamDeriv local responses.

This also restores the uvm-free kokkos implementation of ResponseGLFlux.


jewatkins (approved):
lgtm. static arrays make more sense for this I think.

mperego:
thanks @mcarlson801 !!

PR #1067, opened 2024-07-24
Make CreateExportAlbany less restrictive

mcarlson801:
We're currently unable to build Albany with the MPAS interface enabled on Frontier due to an overhaul of Frontier's module system. This is due to the `CreateExportAlbany` script throwing an error if a `-L` flag points to a directory that doesn't exist. (After the recent module overhaul on Frontier, `rocm/lib64` doesn't exist but Trilinos still adds it to the list of linker flags)

Is there a reason we strictly enforce that `-L` points to a valid directory? I propose that instead of throwing an error here, we just print a warning to the log and skip adding the non-existent paths to `export_albany.in`.

### DIFF
--- a/cmake/CreateExportAlbany
+++ b/cmake/CreateExportAlbany
@@ -98,10 +98,11 @@ def run (bin_dir,install_lib_dir, cmake_generator):
         elif item.startswith("-L"):
             # We want to get an abs path, with symlinks resolved (if any)
             lib_dir_full = pathlib.Path(item[2:]).resolve()
+            # If path doesn't exist or it isn't a directory, don't add it to the "libs_dirs" list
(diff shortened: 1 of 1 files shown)


jewatkins (approved):

jewatkins:
I'd wait for Luca's input on this just in case.

bartgol (commented):
The reason for throwing was that this seemed like an error. Given the situation, I think we can allow for this to happen, but I would print a nice WARNING in the message, so that the user does not miss it.

mcarlson801:
Good call, I added a "WARNING" to the print statements so they don't get lost in the log.

bartgol (approved):

PR #1070, opened 2024-07-31
Fix for E3SM-Mali uvm-free error

mcarlson801:
Tpetra host views were not going out of scope before being accessed on device in `SDirichlet` and `SDirichletField` evaluators. This fixes the error that was showing up in CUDA E3SM-Mali runs when UVM is disabled.

@mperego This error didn't show up in the handful of compass tests that I'm able to run on GPU. Is there a way to know when `SDirichletField` is used based on the Albany input or timer output?

### DIFF
--- a/src/evaluators/bc/PHAL_SDirichletField_Def.hpp
+++ b/src/evaluators/bc/PHAL_SDirichletField_Def.hpp
@@ -136,19 +136,21 @@ set_row_and_col_is_dbc(typename Traits::EvalData dirichlet_workset)
   row_is_dbc_->assign(0.0);
   col_is_dbc_->assign(0.0);
 
-  auto row_is_dbc_data = Albany::getNonconstLocalData(row_is_dbc_);
+  {
+    auto row_is_dbc_data = Albany::getNonconstLocalData(row_is_dbc_);
 
-  const auto& ns_node_elem_pos = dirichlet_workset.nodeSets->at(this->nodeSetID);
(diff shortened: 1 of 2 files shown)


jewatkins (approved):
thanks Max!

mperego:
Thanks. MALI hard codes the dirichlet conditions to be SDirichletField. However, this can be overwritten in the ` albany_input.yaml` file with a line like
 `DBC on NS dirichlet ... `

Thanks for fixing this!

mperego:
Does this mean that we have a UVM free E3SM-MALI build on Perlmutter GPU?

jewatkins:
> Does this mean that we have a UVM free E3SM-MALI build on Perlmutter GPU?

yes :)

I think we're still a bit confused on why Albany/MALI tests didn't catch this since we think they also use SDBC. For example,
https://my.cdash.org/tests/184095086
errors out much later (not where e3sm was erroring out). Ideally, we would want to try to figure out how to reproduce this error in albany.

jewatkins:
the input file that doesn't seem to reproduce (but probably should?): https://github.com/sandialabs/Albany/blob/master/tests/landIce/FO_GIS/input_fo_humboldt_muelu.yaml

mperego:
Um. I don't know. One difference might be that E3SM we are solving for a full ice sheet, so the Dirichlet sideset should be empty, whereas most of MALI Albany tests are on glaciers with nonempty dirichlet sidesets.

PR #1075, opened 2024-09-09
Fix compilation error

### LINKED ISSUE #1074: Failing builds due to ParameterList routine call
We are getting an error due to an alleged erroneous call to a PL routine:
```
/.../repos/Albany/src/Albany_StateManager.cpp(65): error: no instance of overloaded function "Teuchos::ParameterList::set" matches the argument list
```
Curiously, this is showing up right now in the Weaver build only on the CDash (https://sems-cdash-son.sandia.gov/cdash/viewBuildError.php?buildid=204332) but also in both my spack nightly builds, which are non-GPU.  

Could someone please look into this?  Did some PL routines get deprecated?

bartgol:
Fixes #1074


ikalash:
I just tried building this and it's not building.  First, in Albany/CMakeLists.txt, you need to set a 16.1 required Trilinos instead of 15.1 - that is minor.  There are more serious issues regarding STK:

```
/nightlyCDash/repos/Albany/src/disc/stk/percept/stk_rebalance/Rebalance.cpp:103:16: error: 'get_entities_through_relations' is not a member of 'stk::mesh'
  103 |     stk::mesh::get_entities_through_relations(bulk_data, elems, dep_rank, related_entities);
      |                ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
[9/26] Building CXX object src/landIce/CMakeFiles/landIce.dir/problems/LandIce_StokesFOBase.cpp.o
FAILED: src/landIce/CMakeFiles/landIce.dir/problems/LandIce_StokesFOBase.cpp.o 
```

The fact that your CMakeLists.txt did not have the right Trilinos number tells me that maybe you forked from the wrong version of Albany - is that possible?  Did the code build for you?

mperego:
Maybe we should just merge this. The changes are minor and seem benign.

ikalash:
Ok, I will merge, and we can see what happens tomorrow in the nightlies.

src/Albany_StateManager.cpp
```diff
@@ -62,9 +62,9 @@ StateManager::registerStateVariable(
   Teuchos::RCP<Teuchos::ParameterList> p =
       Teuchos::rcp(new Teuchos::ParameterList(
           "Save or Load State " + stateName + " to/from field " + stateName));
-  p->set<const std::string>("State Name", stateName);
-  p->set<const std::string>("Field Name", stateName);
-  p->set<const Teuchos::RCP<PHX::DataLayout>>("State Field Layout", dl);
```
bartlettroscoe, line 67: Yea, it definingly did not make sense to use `const` with this template parameter to the PL::set() function.  (The value is already passed in as `const T&` so there was no need to specify `T` as `const`, which would make `const const std::string&`?)

Now, if you wanted to be sure to control the type used to store the parameter value, you could have just dropped `const` and called:

```
p->set<std::string>("State Name", stateName);
```

That makes 100% sense.

I am kind of surprised this compiled even with the old implementation for PL::set().  (C++ templates are so complex.)

PR #1076, opened 2024-09-11
Add extruded discretizations that only store a basal mesh

bartgol:
This branch does a few things:

- Add interface for an Extruded mesh that does not store full 3d mesh info
- Add interface for an Extruded conn mgr, that uses a basal mesh conn-mgr and layers data to create full 3d connectivity
- Add interface for an Extruded discretization, which relies on the above two

The extruded disc is still lacking some functionalities. E.g., it does not support saving the solution to the mesh or to file (it would likely have to save fields layer-by-layer, as N fields, or as a vector field at each basal DOF, depending on layered mesh ordering). The extruded disc is called "NewExtruded", to differentiate from the current "Extruded", which uses the ExtrudedSTKMeshStruct mesh.

I also found a bug in the STKConnManager::getConnectivityMask, which was creating the wrong mask.

I added a test, where I run the same Heat3D problem with either an Extuded disc or a good old STK3D disc. The test is only run as extruded, but there's a yaml file ready to run as monolithic, for debug purposes.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -232,6 +232,7 @@ list (APPEND SOURCES
   disc/Albany_MeshSpecs.cpp
   disc/Albany_DOFManager.cpp
   disc/Albany_DiscretizationUtils.cpp
+  disc/Albany_ExtrudedMesh.cpp
   )
 list (APPEND HEADERS
   disc/Albany_DiscretizationUtils.hpp
@@ -243,6 +244,7 @@ list (APPEND HEADERS
   disc/Albany_MeshSpecs.hpp
   disc/Albany_ConnManager.hpp
   disc/Albany_DOFManager.hpp
+  disc/Albany_ExtrudedMesh.hpp
   )
 
 #stk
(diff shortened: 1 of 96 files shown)


mperego (approved):
This looks great! Thanks @bartgol. Thanks also for adding unit tests.

Can we chage the name `NewExtruded` with something more descriptive? Maybe `LogicallyExtruded`?

bartgol:
> This looks great! Thanks @bartgol. Thanks also for adding unit tests.
> 
> Can we chage the name `NewExtruded` with something more descriptive? Maybe `LogicallyExtruded`?

Yes, we can. Ideally, though, I'd like to change the old Extruded into something like STKExtruded, and use Extruded for this disc type.

mperego:
> > This looks great! Thanks @bartgol. Thanks also for adding unit tests.
> > Can we chage the name `NewExtruded` with something more descriptive? Maybe `LogicallyExtruded`?
> 
> Yes, we can. Ideally, though, I'd like to change the old Extruded into something like STKExtruded, and use Extruded for this disc type.

That's fine with me.

PR #1081, opened 2024-09-25
Warnings eradication

bartgol:
This PR aims to remove a TON of warnings, exposed by "-Wall -Wextra", in the hope of getting the warning free build back to normal. I am not yet done (and the branch does not compile as of now).

### DIFF
--- a/src/corePDEs/problems/Albany_PopulateMesh.cpp
+++ b/src/corePDEs/problems/Albany_PopulateMesh.cpp
@@ -49,7 +49,6 @@ void PopulateMesh::buildProblem (Teuchos::ArrayRCP<Teuchos::RCP<MeshSpecsStruct>
   cellCubature  = cubFactory.create<PHX::Device, RealType, RealType>(*cellTopology, cubDegree);
 
   const int worksetSize     = meshSpecs[0]->worksetSize;
-  const int numCellSides    = cellTopology->getFaceCount();
   const int numCellVertices = cellTopology->getNodeCount();
   const int numCellNodes    = cellBasis->getCardinality();
   const int numCellQPs      = cellCubature->getNumPoints();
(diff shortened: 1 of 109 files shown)


src/corePDEs/problems/Albany_PopulateMesh.cpp
```diff
@@ -82,7 +82,7 @@ void PopulateMesh::buildProblem (Teuchos::ArrayRCP<Teuchos::RCP<MeshSpecsStruct>
       const int numSideVecDim   = -1;
 
       dl->side_layouts[ss_name] = Teuchos::rcp(new Layouts(numSideVertices,numSideNodes,numSideQPs,
-                                                           numSideDim,numCellDim,numCellSides,numSideVecDim,ss_name));
```
bartgol, line 85: This parameter was never used, so I decided to change the signature of the method. It used to be used when we had layouts of the form <Cell,Side,...>, but after Max's work, we just have <Side,...>, so this is now pointless.

src/problems/Albany_ResponseUtilities.hpp
```diff
@@ -38,18 +38,16 @@ class ResponseUtilities
   constructResponses(
     PHX::FieldManager<PHAL::AlbanyTraits>& fm0,
     Teuchos::ParameterList& responseList,
-    Teuchos::RCP<Teuchos::ParameterList> paramsFromProblem,
```
bartgol, line 41: The state mgr was never used, so again, I decided to change the method signature, to keep things slimmer.

bartgol:
@mperego The vast majority of this PR is adding comments to methods inputs to avoid compiler warnings. There's a couple of legitimate bugs, but which we were not hitting with our production codes anyways. I also changed a few methods signatures to remove inputs that were never used anyways.

Feel free to review, if you want. But I think I want to merge this today, so we can verify all warnings are removed tomorrow.

src/landIce/evaluators/LandIce_SimpleOperation.hpp
```diff
@@ -111,7 +111,7 @@ struct Sum
 {
   void setup (const Teuchos::ParameterList& /*p*/) {}
   ScalarT operator() (const ScalarT& x, const ScalarT& y) const {
-    return y;
+    return x+y;
```
mperego, line 114: Good catch!

mperego (approved):
Looks great! Thanks!!

PR #1082, opened 2024-09-27
STK-related fixes to let Albany build correctly

bartgol:
This nano-PR removes some compiler errors as well as linker errors, due to recent changes in STK.

@mperego @ikalash unless you have objections, I will merge this today, so the nightlies can build.


mperego (approved):
Looks good to me.

PR #1086, opened 2024-10-31
Porting some optimization cases to run on GPU without UVM

mcarlson801:
This PR does the following:
1. Adds Kokkos implementations for a number of scatter and gather specializations necessary for optimization cases to work on GPU when there is no unified virtual memory.
2. Adds Kokkos implementation for ResponseSquaredL2DifferenceSide that is used in a number of optimization cases to work on uvm-free GPU builds
3. Updates the scatter unit tests to use Thyra multivectors instead of vectors. There are downstream issues when casting Thyra vectors to multivectors that were exposed by my changes so these tests had to be updated.
4. Makes some of the DOFManager's internal data available as Kokkos Views to be device accessible.

I'm going to do some performance profiling to get an idea of what impact these changes have on performance but the code is ready for review in the meantime.

### DIFF
--- a/src/evaluators/gather/PHAL_GatherScalarNodalParameter_Def.hpp
+++ b/src/evaluators/gather/PHAL_GatherScalarNodalParameter_Def.hpp
@@ -215,20 +215,6 @@ evaluateFields(typename Traits::EvalData workset)
     bool trans = workset.transpose_dist_param_deriv;
     const auto elem_dof_lids = dof_mgr->elem_dof_lids().dev();
 
-    // allocate View for local_Vp
-    int num_cols = 0;
-    if (Vp != Teuchos::null) {
-      num_cols = Vp->domain()->dim();
-      if (trans) {
-        const int num_dofs = elem_dof_lids.extent(1);
-        workset.local_Vp = Kokkos::View<double***, PHX::Device>("local_Vp",workset.numCells,num_dofs,num_cols);
-      } else {
-        workset.local_Vp = Kokkos::View<double***, PHX::Device>("local_Vp",workset.numCells,num_deriv,num_cols);
-      }
-    }
-
-    const auto& local_Vp = workset.local_Vp;
-
     Kokkos::parallel_for(this->getName(),RangePolicy(0,workset.numCells),
                         KOKKOS_CLASS_LAMBDA(const int& cell) {
       const auto elem_LID = elem_lids(cell);
@@ -244,6 +230,18 @@ evaluateFields(typename Traits::EvalData workset)
     });
 
     if (Vp != Teuchos::null) {
+            const int num_cols = Vp->domain()->dim();
+
+      const int num_dofs = elem_dof_lids.extent(1);
+      if (trans) {
+        const int num_dofs = elem_dof_lids.extent(1);
+        workset.local_Vp = Kokkos::View<double***, PHX::Device>("local_Vp",workset.numCells,num_dofs,num_cols);
+      } else {
+        workset.local_Vp = Kokkos::View<double***, PHX::Device>("local_Vp",workset.numCells,num_deriv,num_cols);
+      }
+
+      const auto& local_Vp = workset.local_Vp;
+
       if (trans) {
         for (int eq=0; eq<neq; ++eq) {
           const auto& offsets = dof_mgr->getGIDFieldOffsetsKokkos(eq);
@@ -252,8 +250,9 @@ evaluateFields(typename Traits::EvalData workset)
           Kokkos::parallel_for(this->getName()+"_transvp",RangePolicy(0,workset.numCells),
                         KOKKOS_CLASS_LAMBDA(const int& cell) {
             const auto elem_LID = elem_lids(cell);
-            auto dof_lids = Kokkos::subview(elem_dof_lids,elem_LID,ALL);
-            for (int o=0; o<num_offsets; ++o) {
+            const auto dof_lids = Kokkos::subview(elem_dof_lids,elem_LID,ALL);
+            for (int i=0; i < num_offsets; ++i) {
+              const auto o = offsets(i);
               const LO lid = dof_lids(o);
               for (int col=0; col<num_cols; ++col)
                 local_Vp(cell,o,col) = Vp_data(lid,col);
@@ -348,13 +347,23 @@ evaluateFields(typename Traits::EvalData workset)
   // Are we differentiating w.r.t. this parameter?
   const bool is_active = (workset.dist_param_deriv_name == this->param_name);
 
-  const auto& local_Vp = workset.local_Vp;
-
   // If active, initialize data needed for differentiation
   if (is_active) {
     const int neq = sol_dof_mgr->getNumFields();
     const int num_deriv = this->numNodes;
     const bool trans = workset.transpose_dist_param_deriv;
+
+        if (Vp != Teuchos::null) {
+      const int num_cols = workset.Vp->domain()->dim();
+      if (trans) {
+        const int num_dofs = elem_dof_lids.extent(1);
+        workset.local_Vp = Kokkos::View<double***, PHX::Device>("local_Vp",workset.numCells,num_dofs,num_cols);
+      } else {
+        workset.local_Vp = Kokkos::View<double***, PHX::Device>("local_Vp",workset.numCells,num_deriv,num_cols);
+      }
+    }
+    const auto& local_Vp = workset.local_Vp;
+    
     for (std::size_t cell=0; cell<workset.numCells; ++cell) {
       const auto elem_LID = elem_lids(cell);
       const auto basal_elem_LID = layers_data->getColumnId(elem_LID);
--- a/src/evaluators/gather/PHAL_GatherScalarNodalParameter_Def.hpp
+++ b/src/evaluators/gather/PHAL_GatherScalarNodalParameter_Def.hpp
@@ -318,7 +318,12 @@ evaluateFields(typename Traits::EvalData workset)
   const auto bot = layers_data->bot_side_pos;
   const auto top = layers_data->top_side_pos;
   const auto ws = workset.wsIndex;
-  const auto elem_lids = workset.disc->getElementLIDs_host(ws);
+  const auto elem_lids_ws = workset.disc->getWsElementLIDs();
+  const auto elem_lids = Kokkos::subview(elem_lids_ws.dev(),ws,Kokkos::ALL);
+
+  const auto layerOrd = layers_data->layerOrd;
+  const auto numHorizEntities = layers_data->numHorizEntities;
+  const auto numLayers = layers_data->numLayers;
 
   // Pick element layer that contains the field level
   const auto fieldLayer = fieldLevel==layers_data->numLayers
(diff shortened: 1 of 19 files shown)


jewatkins (approved):
lgtm, thanks!

src/evaluators/gather/PHAL_GatherScalarNodalParameter_Def.hpp
```diff
@@ -211,58 +213,76 @@ evaluateFields(typename Traits::EvalData workset)
...
+                        KOKKOS_CLASS_LAMBDA(const int& cell) {
       const auto elem_LID = elem_lids(cell);
       const auto p_dof_lids = Kokkos::subview(p_elem_dof_lids,elem_LID,ALL);
       for (int node=0; node<num_deriv; ++node) {
         const LO lid = p_dof_lids(node);
 
         // Initialize Fad type for parameter value
-        const auto p_val = lid>=0 ? p_data[lid] : 0;
+        const auto p_val = lid>=0 ? p_data(lid) : 0;
         ParamScalarT v(num_deriv, node, p_val);
```
bartgol, line 227: Does this work on device, and is it performant? With certain fads, doesn't it allocate memory on device at every call?

src/evaluators/gather/PHAL_GatherScalarNodalParameter_Def.hpp
```diff
@@ -330,6 +350,18 @@ evaluateFields(typename Traits::EvalData workset)
     const int neq = sol_dof_mgr->getNumFields();
     const int num_deriv = this->numNodes;
     const bool trans = workset.transpose_dist_param_deriv;
+
+        if (Vp != Teuchos::null) {
```
bartgol, line 354: Indentation seems off here.

src/evaluators/response/PHAL_ResponseSquaredL2DifferenceSide_Def.hpp
```diff
@@ -200,46 +196,52 @@ evaluateFields(typename Traits::EvalData workset)
...
       case 1:
-        for (int sideSet_idx = 0; sideSet_idx < sideSet.size; ++sideSet_idx)
-        {
+        Kokkos::parallel_for(this->getName(),RangePolicy(0,sideSet.size),
+                       KOKKOS_CLASS_LAMBDA(const int& sideSet_idx) {
           // Get the local data of cell
-          const int cell = sideSet.ws_elem_idx.h_view(sideSet_idx);
+          const int cell = sideSet.ws_elem_idx.d_view(sideSet_idx);
+
+          ScalarT diff_1[8] = {0};
```
bartgol, line 209: Another case of an automatic tmp fad inside a kernel: are these ok? For SFad and SLFad, I think the answer is yes, but for generic Fad, I'm not sure. Someone with more Fad knowledge than me may know.

src/evaluators/response/PHAL_ResponseSquaredL2DifferenceSide_Def.hpp
```diff
@@ -200,46 +196,52 @@ evaluateFields(typename Traits::EvalData workset)
...
+              }
             }
             sum += sq / rms2 * w_measure(sideSet_idx,qp);
           }
 
-          this->local_response_eval(cell, 0) = sum*scaling;
-          this->global_response_eval(0) += sum*scaling;
-        }
+          this->local_response_eval(cell,0) = sum*scaling;
+          KU::atomic_add<ExecutionSpace>(&(this->global_response_eval(0)), sum*scaling);
```
bartgol, line 235: Wouldn't it be better to use parallel_reduce, rather than a parallel_for with atomic access? All threads access _the same_ value, so contention is very high here...

tests/unit/evaluators/ScatterScalarResponse.cpp
```diff
@@ -74,31 +74,31 @@ TEUCHOS_UNIT_TEST(evaluator_unit_tester, separableScatterScalarResponseHessianVe
     auto overlapped_p_space = p_dof_mgr->ov_indexer()->getVectorSpace();
     auto overlapped_x_space = x_dof_mgr->ov_indexer()->getVectorSpace();
 
-    const auto hess_vec_prod_g_xx = Thyra::createMember(x_space);
```
bartgol, line 77: I don't follow these changes. Why do we need MV's (with 1 col)?

tests/unit/evaluators/ScatterResidual.cpp
```diff
@@ -75,30 +75,30 @@ TEUCHOS_UNIT_TEST(evaluator_unit_tester, scatterResidualHessianVecTensorRank0)
...
-  const auto hess_vec_prod_f_xx_out = Thyra::createMember(x_space);
-  const auto hess_vec_prod_f_xp_out = Thyra::createMember(x_space);
-  const auto hess_vec_prod_f_px_out = Thyra::createMember(p_space);
-  const auto hess_vec_prod_f_pp_out = Thyra::createMember(p_space);
-
-  const auto ov_hess_vec_prod_f_xx_out = Thyra::createMember(ov_x_space);
-  const auto ov_hess_vec_prod_f_xp_out = Thyra::createMember(ov_x_space);
-  const auto ov_hess_vec_prod_f_px_out = Thyra::createMember(ov_p_space);
-  const auto ov_hess_vec_prod_f_pp_out = Thyra::createMember(ov_p_space);
+  const auto hess_vec_prod_f_xx = Thyra::createMembers(x_space,1);
```
bartgol, line 78: Same here: why MV with 1 col rather than just V?

bartgol (commented):
Looks good! I have a few doubts and questions.

src/evaluators/response/PHAL_ResponseSquaredL2DifferenceSide_Def.hpp
mperego: That's a good point. With DFad it will create temporaries, and it could run out of memory.

jewatkins: it's an issue with dfad on gpu but we don't use dfad.

tests/unit/evaluators/ScatterScalarResponse.cpp
mcarlson801: I'm not 100% sure what the cause is but these unit tests started failing when I tried to grab the HessianWorkset multivector view using `getNonconstDeviceData`. I was getting an exception that the Thyra_Multivector couldn't be cast to a Tpetra one. Looking at it in a debugger made me think something went wrong when casting a Thyra_Vector to a Thyra_Multivector. Changing the unit tests to use MVs from the start fixed it.

src/evaluators/response/PHAL_ResponseSquaredL2DifferenceSide_Def.hpp
mcarlson801: Yeah, this should definitely be a reduce. I ported this early on when I was just trying to get something to work. I'll fix it.

src/evaluators/response/PHAL_ResponseSquaredL2DifferenceSide_Def.hpp
bartgol: We do allow DFad during config though. If we don't want to allow DFad, we should make it clear and throw during config time.

bartgol: I'm not concerned about running out of memory. I'm more concerned with doing lots of small Cuda allocations at run time.

tests/unit/evaluators/ScatterScalarResponse.cpp
bartgol: Ah, these are unit tests. Then it's ok.

src/evaluators/response/PHAL_ResponseSquaredL2DifferenceSide_Def.hpp
jewatkins: yeah that makes sense. we should just set default slfad values (that work with testing) when running with cuda/hip/sycl instead of having those lines in every albany config. And error out if they are explicitly set to dfad.

when using sfad/slfad, they should be static allocations in local memory/registers. worse case, if the memory spills, it should be as performant as a read/write to global memory but not as bad as a device malloc, which is what's attempted with dfad (unless we setup the memory pool thing).

my concern would actually be, how large is the derivative dimension in this scalar? I forgot we're talking about optimization which could have a lot of derivative components... in which case, we may run out of memory...

src/evaluators/response/PHAL_ResponseSquaredL2DifferenceSide_Def.hpp
mcarlson801: @bartgol Is there a trick to doing a parallel_reduce with Sacado FAD types? I created a very simple example program based on https://kokkos.org/kokkos-core-wiki/ProgrammingGuide/Custom-Reductions-Built-In-Reducers-with-Custom-Scalar-Types.html and I'm getting compiler errors:

```
/pscratch/sd/m/mcarlson/IntroToHPC/repos/trilinos-lite/packages/sacado/example/my_custom_reduce_example.cpp(35): error: no instance of constructor "Kokkos::View<DataType, Properties...>::View [with DataType=ScalarT *, Properties=<Kokkos::CudaSpace, Kokkos::MemoryUnmanaged>]" matches the argument list
            argument types are: (ScalarT *, int)
          detected during:
            instantiation of "SumScalarT<ScalarT, Space>::result_view_type SumScalarT<ScalarT, Space>::view() const [with ScalarT=ScalarT, Space=Kokkos::CudaSpace]"
/pscratch/sd/m/mcarlson/IntroToHPC/repos/trilinos-lite/packages/kokkos/core/src/Kokkos_Parallel_Reduce.hpp(1525): here
            instantiation of "void Kokkos::Impl::ParallelReduceAdaptor<PolicyType, FunctorType, ReturnType>::execute_impl(const std::string &, const PolicyType &, const FunctorType &, ReturnType &) [with PolicyType=Kokkos::RangePolicy<Kokkos::DefaultExecutionSpace>, FunctorType=lambda [](int, ValueType &)->void, ReturnType=SumScalarT<ScalarT, Kokkos::CudaSpace>]"
/pscratch/sd/m/mcarlson/IntroToHPC/repos/trilinos-lite/packages/kokkos/core/src/Kokkos_Parallel_Reduce.hpp(1542): here
            instantiation of "std::enable_if_t<<expression>, void> Kokkos::Impl::ParallelReduceAdaptor<PolicyType, FunctorType, ReturnType>::execute(const std::string &, const PolicyType &, const FunctorType &, ReturnType &) [with PolicyType=Kokkos::RangePolicy<Kokkos::DefaultExecutionSpace>, FunctorType=lambda [](int, ValueType &)->void, ReturnType=SumScalarT<ScalarT, Kokkos::CudaSpace>, Dummy=SumScalarT<ScalarT, Kokkos::CudaSpace>]"
/pscratch/sd/m/mcarlson/IntroToHPC/repos/trilinos-lite/packages/kokkos/core/src/Kokkos_Parallel_Reduce.hpp(1798): here
            instantiation of "std::enable_if_t<<expression>, void> Kokkos::parallel_reduce(const size_t &, const FunctorType &, const ReturnType &) [with FunctorType=lambda [](int, ValueType &)->void, ReturnType=SumScalarT<ScalarT, Kokkos::CudaSpace>]"
(51): here
```

bartgol: Uhm, I am not sure. Maybe @etphipp has some advice.

mcarlson801:
Performance profiling and comparison for this branch on ant-4-20km optimization case is up at https://acme-climate.atlassian.net/wiki/spaces/FAN/pages/4936990749/UVM-free+GPU+Optimization+Profiling

Summary:
- No performance loss for serial build (slight improvement in fill time for distparamderiv specialization)
- 2.45x speedup of total time with this branch without UVM over the old cuda build with UVM
- This branch eliminates major bottleneck for optimization cases on GPU, specifically 4.8x speedup for the optimization related fill-time and 11.1x speedup for filling distparamderivs.
- Cuda build using this branch is 2x faster than a serial build using the same number of nodes (Perlmutter-CPU vs Perlmutter-GPU)

Didn't hear anything back on the issue with kokkos_reduce using fad types, but atomic_adds don't seem to hurt performance. Could be something for future performance optimization.

mperego:
This is amazing. Thanks @mcarlson801 !

PR #1091, opened 2024-11-18
Initial work on evolution of ice thickness.

mperego:
This PR:
Modifies StokesFOThickness problem to enable thickness evolution using Tempus implicit schemes.

Adds a few fixes for Albany to run when `MESH_DEPENDS_ON_SOLUTION` is on. 
Adds two tests computing thickness evolution and sensitivities (sensitivity test is enabled when `MESH_DEPENDS_ON_SOLUTION` is on).


src/landIce/evaluators/LandIce_UpdateZCoordinate_Def.hpp
```diff
@@ -22,15 +22,19 @@
...
+template<typename EvalT, typename Traits, typename ScalarT>
+UpdateZCoordinateMovingTopBase<EvalT, Traits, ScalarT>::
+UpdateZCoordinateMovingTopBase (const Teuchos::ParameterList& p,
                             const Teuchos::RCP<Albany::Layouts>& dl) :
   coordVecIn (p.get<std::string> ("Old Coords Name"), dl->vertices_vector),
   bedTopo(p.get<std::string> ("Bed Topography Name"), dl->node_scalar),
   topSurface(p.get<std::string>("Top Surface Name"), dl->node_scalar),
   coordVecOut(p.get<std::string> ("New Coords Name"), dl->vertices_vector)
 {
+  bool lossyConversion = p.isParameter("Allow Loss Of Derivative Terms") ? p.get<bool>("Allow Loss Of Derivative Terms") : false;
```
bartgol, line 34: What is this about? Are we planning to support lossy and non-lossy later on? Also, what's the loss we're talking about?

bartgol (approved):

src/landIce/evaluators/LandIce_UpdateZCoordinate_Def.hpp
mperego: We could allow lossy for building jacobian and sensitivities.
In this case, we need the mesh coords to be FAD type to get an exact Jacobian, but we can also neglect those derivatives and build an approximate Jacobian.

bartgol: You mean derivatives of H wrt u and viceversa?

mperego: Here we are missing the derivatives of the z-coords of the mesh w.r.t. H.

mperego: If you look in that evaluator you'll see a `Albany::convertScalar<MeshScalarT>(zcoord)`. Here `zcoord` is the value of the updated vertical coordinate that depend on `H`, and is a `ScalarT`. If `MeshScalarT` is a double during the `Jacobian` evaluation, we are stripping out the derivatives to be able to assign t to the mesh.

jewatkins (approved):
lgtm

mperego:
Thanks @bartgol and @jewatkins for reviewing this.

PR #1093, opened 2024-11-22
Move exception outside kernel

jewatkins:
@mperego does this look okay? I moved it outside the kernel since teuchos exceptions aren't device friendly and the gpus builds broke last night.


mperego (approved):
Oh, yeah. Sorry about that.

PR #1096, opened 2024-12-04
Fix loadStateField evaluator to address #1092

### LINKED ISSUE #1092: Nightly build failures
Couple builds broken this morning. It looks like it's due to recent PRs. @mperego do you want to take a look?

mperego:
Apparently linear access of Kokkos dynamic rank views is no longer working. 
Not sure if this is a bug (DynRankView is not as widely tested as View) or expected behavior.  
Fixes #1092

### DIFF
--- a/src/evaluators/state/PHAL_LoadStateField_Def.hpp
+++ b/src/evaluators/state/PHAL_LoadStateField_Def.hpp
@@ -56,7 +56,7 @@ void LoadStateFieldBase<EvalT, Traits, ScalarType>::evaluateFields(typename Trai
   Kokkos::parallel_for(this->getName(),
                        Kokkos::MDRangePolicy<ExecutionSpace, Kokkos::Rank<3>>({0,0,0},{stateData.extent(0),stateData.extent(1),stateData.extent(2)}),
                        KOKKOS_CLASS_LAMBDA(const int i, const int j, const int k) {
-    field(i,j,k) = stateData(i,j,k);
+    field.access(i,j,k) = stateData.access(i,j,k);  //works also when rank is less than 3
   });
 }
 
@@ -102,7 +102,7 @@ void LoadStateField<EvalT, Traits>::evaluateFields(typename Traits::EvalData wor
   Kokkos::parallel_for(this->getName(),
                        Kokkos::MDRangePolicy<ExecutionSpace, Kokkos::Rank<3>>({0,0,0},{stateData.extent(0),stateData.extent(1),stateData.extent(2)}),
                        KOKKOS_CLASS_LAMBDA(const int i, const int j, const int k) {
-    field(i,j,k) = stateData(i,j,k);
+    field.access(i,j,k) = stateData.access(i,j,k); //works also when rank is less than 3
   });
 }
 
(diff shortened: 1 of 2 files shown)


src/evaluators/state/PHAL_LoadStateField_Def.hpp
```diff
@@ -51,15 +50,13 @@ void LoadStateFieldBase<EvalT, Traits, ScalarType>::evaluateFields(typename Trai
...
 
-  MDFieldVectorRight<ScalarType> g(data);
-  dataVec = g;
+  ALBANY_ASSERT (stateData.rank() <= 3, "Current implementation supports only views with rank up to 3. If larger rank is needed modify code below");
 
   Kokkos::parallel_for(this->getName(),
-                       Kokkos::RangePolicy<ExecutionSpace>(0,data.size()),
-                       KOKKOS_CLASS_LAMBDA(const int i) {
-    dataVec[i] = (i < stateToLoad_size) ? stateData(i) : 0.0;
+                       Kokkos::MDRangePolicy<ExecutionSpace, Kokkos::Rank<3>>({0,0,0},{stateData.extent(0),stateData.extent(1),stateData.extent(2)}),
```
bartgol, line 57: This works cause we use DynRankView, so usince 3 indices is always ok (we just have extent(1) and extent(2) that may be 1), right?

bartgol (approved):
Sneaky. That was a great find Mauro, thanks!

mcarlson801 (approved):
Nice catch! This change looks good to me.

mperego:
Another alternative would be to deep_copy the state view into the MDfiled view, but this is a bit more fragile and it expects the views to have same properties (e.g. Kokkos layouts).

bartgol (approved):

mcarlson801:
> An alternative would be to deep_copy the state view into the MDfiled view, but this is a bit more fragile and it expects the views to have same properties (e.g. Kokkos layouts).

I can confirm this doesn't work, I tried this awhile back before using an MDField iterator.

src/evaluators/state/PHAL_LoadStateField_Def.hpp
jewatkins: kind of weird that this works for rank<3 but nice that it does since it seems a lot simpler.

jewatkins (changes requested):
this seems to point to some issue with `MDFieldVectorRight`. could you finish removing it from the code? It looks like it's only in a few other places: ScatterScalarResponse, SeparableScatterScalarResponse, and PHAL_Utilities

mcarlson801:
> this seems to point to some issue with `MDFieldVectorRight`. could you finish removing it from the code? It looks like it's only in a few other places: ScatterScalarResponse, SeparableScatterScalarResponse, and PHAL_Utilities

The DynRankView that is causing problems is the one in stateData, I'm pretty sure. MDFieldVectorRight is designed to be accessed linearly.

mperego:
Yes, I agree, stateData was the issue.

jewatkins (approved):
ah okay, then it's fine.

PR #1105, opened 2025-01-10
First iteration of adaptivity hooks

bartgol:
This PR adds a first version of the adaptivity hooks. We may change them later if/when we realize they are insufficient/overkill/wrong/whatever. The goal was to get _something_ working.

Some details:

- Added a version of PiroObserver that is also Tempus-compatible (so it can be called by Tempus during the time integration)
- Modified SolutionManager to
	- provide DxDp (not just x)
	- update vector spaces
- Added interfaces to discretization for calling adaptivity
- Moved AdvDiff test to corePDEs
- Added a simple 1d adaptivity test for AdvDiff: starts with a mesh too coarse (hence, solution oscillates), then refines the mesh to get a stable solution.
- Added a simple adaptivity for STK 1d meshes. If user requested it, we refine the 1d mesh if the solution oscillates AND an approx of the hessian is large enough. This check really only works with 1 MPI rank, so should not be used in general. It was just for the unit test. I could do a few mods for the partitioned case, but didn't want to spend too much effort, since we really won't be doing adaptivity with STK meshes...


src/disc/Albany_AbstractDiscretization.hpp
```diff
@@ -389,6 +392,16 @@ class AbstractDiscretization
       const bool               overlapped = false,
       const bool               force_write_solution = false) = 0; 
 
+  // Check if mesh adaptation is needed, and if so what kind (topological or just mesh-movement)
+  virtual Teuchos::RCP<AdaptationData>
```
bartgol, line 396: I'm thinking I may change this. Namely, this may just return an AdaptationType enum. The reason is that instead of doing
```c++
auto data = disc->checkForAdaptation(x,xdot,xdotdot,dxdp);
if (data->type!=None) {
  disc->adapt(data);
}
```
We could just do
```c++
auto type = disc->checkForAdaptation(x,xdot,xdotdot,dxdp);
if (type!=None) {
  disc->adapt(x,xdot,xdotdot,dxdp);
}
```
without forcing the disc to stuff the OLD x/xdot/xdotdot/dxdp inside the returned type (the caller site already has them anyways).

src/disc/Albany_AbstractMeshFieldAccessor.hpp
```diff
@@ -54,6 +54,9 @@ class AbstractMeshFieldAccessor
   virtual void fillSolnMultiVector (Thyra_MultiVector&   soln,
                                     const dof_mgr_ptr_t& sol_dof_mgr,
                                     const bool           overlapped) = 0;
+  virtual void fillSolnSensitivity (Thyra_MultiVector&   dxdp,
```
bartgol, line 57: This, and all the dxdp changes in the PR, are so that we can run a pb that computes sensitivities and do adaptation. I haven't tested the dxdp functionality though, so more work may be needed down the road.

cwsmith (approved):
LGTM.  Thank you.

mperego (approved):
Looks good to me. Would this work with current version of Trilinos/Piro?

src/Albany_PiroTempusObserver.cpp
```diff
@@ -0,0 +1,90 @@
...
+{
+  auto& integrator_nc = const_cast<Tempus::Integrator<ST>&>(integrator);
+  auto  history_nc = integrator_nc.getNonConstSolutionHistory();
+  auto  state_nc = history_nc->getCurrentState();
+
+  TEUCHOS_TEST_FOR_EXCEPTION (state_nc.is_null(), std::runtime_error,
+      "Error! Unexpectedly found a null current state in the tempus integrator.\n");
+
+  //Don't observe solution if step failed to converge
+  if (state_nc->getSolutionStatus() == Tempus::Status::FAILED) {
```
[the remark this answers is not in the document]
bartgol: I don't think so. For instance, we may be using adaptive tempus timestepping, in which case the step may fail to converge without it being an issue, since tempus will try again with a smaller timestep.

bartgol:
> Looks good to me. Would this work with current version of Trilinos/Piro?

It's been a while since I tested, so I'm rebuilding trilinos and running all tests again

bartgol:
Tests (including landice and demo) pass on my workstation. @mperego I would merge, but the dilemma I have is this: I'm taking off 10days starting tomorrow, so I won't be able to address any nightlies failure until March 24th. Would you like me to hold off the merge?

cwsmith:
> @bartgol if Cameron is not going to need these changes right away, I would wait until you are back. I don't expect any significant change in Albany while you are gone.

Working from the branch for a week or so is fine with me.

PR #1111, opened 2025-04-08
Fix MALI interface gpu

jewatkins:
forgot the deep copy. Thanks @mperego for catching this!

BTW, it looks like this case (https://github.com/sandialabs/ali-perf-tests/tree/master/perf_tests/mali-ais8km) still fails to converge with the gpu solver (even with a serial build). But it is converging slower on gpu (final residual after 50 iterations, gpu: `6.167e-01` vs. cpu:`1.874e-01`). I'm on the queue with this change. We'll see if it reproduces the cpu results at least.


mperego (approved):
thanks!

jewatkins:
> Looks good, sorry I missed it on the first review

my fault for trying to push it in EOD before the nightly builds started.

jewatkins:
Confirmed the result is a little better now `1.939e-01`. We'll still need to look into this.

PR #1113, opened 2025-04-09
Fix scatter of side eqn residual

### LINKED ISSUE #1030: corePDEs_SideSetLaplacian_3D failing on weaver after epetra removal
https://sems-cdash-son.sandia.gov/cdash/test/4156291

It looks like weaver has more tests after https://github.com/sandialabs/Albany/pull/1028 was merged (111 -> 125) so maybe this case never ran before?

@mperego Is this a case that ran with epetra and never ran with tpetra? Were you planning to look into it? Anything obvious that shouldn't work on device? If not, @mcarlson801 can try to see what's going on.

bartgol:
The implementation was quite wrong, I have no clue how this could pass on CPU machines...

Fixes #1030

### DIFF
--- a/src/evaluators/scatter/PHAL_ScatterSideEqnResidual.hpp
+++ b/src/evaluators/scatter/PHAL_ScatterSideEqnResidual.hpp
@@ -81,16 +81,15 @@ protected:
   std::string             sideSetName;        // The side set where the equation(s) are defined
   std::vector<Albany::SideStruct> sideSet;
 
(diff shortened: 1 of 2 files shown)


bartgol:
Wait, I'm having second thoughts. The test passes now, but I think it's still wrong (we are getting right results only cause the 2nd equation is just "x = 0".

bartgol:
Scratch what I said. The original impl looks correct to me now. Maybe it's a device/host issue..

PR #1117, opened 2025-04-16
Pull in all builds and tests groups other than Experimental

bartlettroscoe:
@ikalash 

This is the blacklisting approach to specifying the builds and tests you want to select.

This produced the attached file [albanyNightlyBuilds.html.txt](https://github.com/user-attachments/files/19785053/albanyNightlyBuilds.html.txt).


bartlettroscoe:
@ikalash, I think the list of expected builds in the CSV file needs to be fixed.  They don't seem to match up with what is being posted to CDash.

ikalash:
Thanks, this fixes the problem!

PR #1118, opened 2025-04-18
Move load-field routines out of GenericSTKMeshStruct

bartgol:
Omegah discretization will use these same utils, which had little-to-no dependence on the fact that we were in a STK mesh. I move the utils to `Albany_DiscretizationUtils.hpp`, although they may very well just go into the utility folder, since there's little-to-no mention of the fact that this is related to a mesh.

On mappy, landice tests fail, but I get the same regression values as I get with master, so I am assuming some settings are different on mappy. If approved, I will merge this next week, and if something goes bad with nightlies, fix (or revert) the next day.


mperego (approved):
Looks good to me. Thanks!

PR #1119, opened 2025-04-22
Moved STK field container members to AbstractMeshFieldAccessor

bartgol:
They will be used by Omegah as well.

Also, removed StateInfoStruct from STK field containers constructors. Simply call `addStateStructs(sis)` after construction (avoids the clumsy unused arg in the solution-type consturctor).


bartgol:
Merging this, since it just moves code into base class and cleans up a ctor signature. I have a few PRs pipelined, and I want to merge one at a time, to let nightlies help with finding issues.

PR #1122, opened 2025-04-30
Reworked implementation of writeSolution methods in disc classes

bartgol:
Changes summary:

- Renamed writeSolutionToFile as writeMeshDatabaseToFile
- Implement writeSolution by calling writeSolutionToMeshDatabase followed by writeMeshDatabaseToFile in base class
- Removed time arg in writeSolutionToMeshDatabase (it's only needed when writing to file)
- Removed overlapped arg in writeMeshDatabaseToFile (it's only needed when writing to the mesh db)
- Removed defaults in virtual methods (not a good practice)
- Add _some_ impl for omegah disc

### DIFF
--- a/src/disc/Albany_ExtrudedDiscretization.cpp
+++ b/src/disc/Albany_ExtrudedDiscretization.cpp
@@ -87,60 +87,9 @@ ExtrudedDiscretization::setupMLCoords()
 }
 
 void
-ExtrudedDiscretization::writeSolution(
-    const Thyra_Vector& /* soln */,
-    const Teuchos::RCP<const Thyra_MultiVector>& /* soln_dxdp */,
-    const double        /* time */,
-    const bool          /* overlapped */,
-    const bool          /* force_write_solution */)
-{
-  printf("WARNING ExtrudedDiscretization::writeSolution not yet implemented\n");
-  // throw NotYetImplemented("ExtrudedDiscretization::writeSolution");
-}
-
-void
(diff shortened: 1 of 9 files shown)


src/disc/Albany_ExtrudedDiscretization.cpp
```diff
@@ -86,61 +86,10 @@ ExtrudedDiscretization::setupMLCoords()
...
-    const bool               /* force_write_solution */)
-{
-  throw NotYetImplemented("ExtrudedDiscretization::writeSolutionMV");
-}
-
 void
 ExtrudedDiscretization::writeSolutionToMeshDatabase(
     const Thyra_Vector& /* soln */,
     const Teuchos::RCP<const Thyra_MultiVector>& /* soln_dxdp */,
-    const double /* time */,
```
mperego, line 143: Why are you removing time info here?

bartgol: Because the time var is only used when writing _to file_, to assign a label to the unlimited time dim (in netcdf lingo). When we write to the in-memory mesh database, there is no concept of time: the mesh only stores ONE copy of the solution field at any given moment...

bartgol: To clarify a bit better, saving the solution has 2 phases: saving to the in-memory mesh data structures, and writing to file. The time value is only needed in the latter.

mperego (approved):
It looks good to me. thanks!

PR #1123, opened 2025-05-07
Updates to omegah disc structures

bartgol:
This PR includes some work by me as well as @cwsmith regarding extending some omegah capability. I will add a list of mods once the PR is ready for review.

Edit: ok, all tests pass on my desktop, so this PR is in a reasonably good shape. Here's the list of changes:

- Consolidate omegah meshes into one: no more `OmegahOshMesh` and `OmegahBoxMesh`, just `OmegahGenericMesh`.
- Disc factory checks if "Method" is "Omegah" in order to build an `OmegahGenericMesh`. The mesh, inside, takes care of either load or build the mesh, based on the input param `Mesh Creation Method`.
- OmegahGenericMesh can read input fields from file using the recently extracted utilities that used to be in GenericSTKMeshStruct (not yet tested, will be tested in follow up PR).
- OmegahFieldAccessor can write to mesh from an input Thyra_MutliVector, and distinguishes adding and setting a field on mesh (the latter requires it was added already)
- Add support for 1d in omegah conn mgr
- Add support for edges shards<->omegah permutation handling during conn mgr setup.
- OmegahGenericMesh can correctly build node/side sets in 1d and 2d box meshes, using geometric classification. Support for 3d to be added soon.
- Add the 1d adv-diff adaptive test with an Omegah mesh
- Needed by the above: the disc method `checkForAdaptation` needs to be nonconst, since omegah disc needs to save the solution in the mesh first.
- Omegah disc can handle 2+ worksets (a few fixes to a few methods). This was needed since the 1d test starts with 10 elems, a value then used also for worksetSize in all evaluators; when adapt increase the number of elems, it can't add them to the 1st workset, since the evaluators don't get the memo about a changed workset, so omegah is forced to create additional worksets.
- Fixed a couple of bugs in state struct and manager.


cwsmith (changes requested):
Looks good.  Thank you.

I left a few minor comments below regarding debug prints and a couple functions used for the 'box' adapt test(s).

bartgol:
It looks like there are some warning left that make the [No-Warn](https://sems-cdash-son.sandia.gov/cdash/viewBuildError.php?buildid=498841) build fail. I'll fix them.

PR #1125, opened 2025-05-14
Modify StateInfoStruct handling

bartgol:
This PR modifies a few things regarding state info struct. I built the branch on top of #1123 , so that I don't have to solve many conflicts when that one is merged.

Summary of changes:

- Remove completely unused args SIS from a few disc/mesh methods signatures
- Treat Nodal and Elem states more simmetrically in the STK field container
- No need for the STKNodeFieldContainer structures, as we are not using them after creation. They were just adding the field to the mesh, but we can do that manually, like we do for Elem states.
- Use range-for loops in a couple of places, to make the code cleares (no need to track counters/iterators).

The reason for this PR (other than the fact that cleaning up unused code is always nice) is to make the logic cleaner without too many files indirections. This helps a lot when trying to implement new discretizations (like omegah or the extruded ones).

I'll ping someone for review once the Omegah upgrades PR is merged.


src/disc/stk/Albany_GenericSTKFieldContainer.cpp
```diff
@@ -91,107 +91,84 @@ addStateStructs(const Teuchos::RCP<StateInfoStruct>& sis)
...
       } else if (dim.size()==3) {
         // 2nd order tensor on cell
-        cell_tensor_states.push_back(& metaData->declare_field< double >(stk::topology::ELEMENT_RANK, st.name));
+        cell_tensor_states.push_back(& metaData->declare_field< double >(stk::topology::ELEMENT_RANK, st->name));
         stk::mesh::put_field_on_mesh(*cell_tensor_states.back(), metaData->universal_part(), dim[2], dim[1], nullptr);
-        set_output_role(*cell_tensor_states.back(),st.output);
+        set_output_role(*cell_tensor_states.back(),st->output);
       } else {
-        TEUCHOS_TEST_FOR_EXCEPTION (true, std::logic_error, "Error! Unexpected state rank.\n");
+        throw std::logic_error("Error! Unexpected state rank.\n");
```
mperego, line 112: why are you replacin `TEUCHOS_TEST_FOR_EXCEPTION` with `throw`?

src/Albany_StateInfoStruct.hpp
```diff
@@ -174,12 +172,8 @@ struct StateStruct
...
 };
 
-// New container class approach
-class StateInfoStruct : public std::vector<Teuchos::RCP<StateStruct>>
-{
-public:
-  NodeFieldContainer  nodal_field_container;
-};
+// Alias to a vector of state struct pointers
+using StateInfoStruct = std::vector<Teuchos::RCP<StateStruct>>;
```
mperego, line 176: nice!

mperego (approved):
Looks good, thanks for the cleanup/refactor.

src/disc/stk/Albany_GenericSTKFieldContainer.cpp
bartgol: I just find it silly to use a macro to just throw an exception. The macro is meant to wrap a condition check, which should throw the given exception type if the check fails. Hard-coding `false` is just a convoluted way to throw the exception. At this point, why not just throwing the exception by hand?

bartgol: If you dislike this, I can roll back the change.

mperego: Sounds reasonable. I guess one difference is that throw will throw a message for each process, right?

bartgol: does TEUCHOS_TEST_FOR_EXCEPTION only throw on rank0?

mperego: Uhm.. maybe not.. I got confused. Here's is what the macro does:
```
{ \
  const bool throw_exception = (throw_exception_test); \
  if(throw_exception) { \
    Teuchos::TestForException_incrThrowNumber(); \
    const int throwNumber = Teuchos::TestForException_getThrowNumber(); \
    std::ostringstream omsg; \
    omsg \
      << __FILE__ << ":" << __LINE__ << ":\n\n" \
      << "Throw number = " << throwNumber \
      << "\n\n" \
      << "Throw test that evaluated to true: "#throw_exception_test \
      << "\n\n" \
      << msg; \
    const std::string &omsgstr = omsg.str(); \
    TEUCHOS_STORE_STACKTRACE(); \
    Teuchos::TestForException_break(omsgstr, throwNumber); \
    throw Exception(omsgstr); \
  } \
}
```

bartgol: Yeah, I _think_ Teuchos simply wraps the message formatting and gives a point where one can break wtih gdb.

I'm going to merge as is, I can revert in the next PR I submit.

PR #1126, opened 2025-05-16
Adding AlbanyLaunchBounds to control Kokkos Launch Bounds configurations for GPU kernel optimization

OscarAntepara:
By adding AlbanyLaunchBounds, we can control the default behavior of Kokkos launchbounds for AMD GPUs. 
We found that by using  Kokkos::LaunchBounds<128,2> on AMD MI250X GPUs, we could improve performance of the StokesFOResid<Jacobian> kernel by 1.54x


jewatkins (approved):

mperego (approved):
Lgtm. Thanks!

PR #1127, opened 2025-05-22
Fix ISMIP tests mesh modification

bartgol:
I think the current way we modify the mesh in ISMIP tests is not right. This PR aims to fix (and simplify) the logic.

The current implementation, does the following on each element:

- check if any coord is 0 (along the dim where periodic mesh is requested). If none are 0, go to next elem
- If at least 1 coord is 0, check if any coord (along the periodic dim) is larger than the mesh width / 1.9. If not, go to the next elem
- At this point, loop over nodes, and if any node has coord 0 (along the periodic dim) set it equal to the mesh width (meaning the node wraps around). ALSO, subtract the factor x[d]*tan(alpha) to surface height, which makes surf_height match the value it would have gotten when computed (during transformMesh) with the right (meaning periodic) coordinates.

My concern is mostly with step 2: first, what's with 1.9? It's such an arbitrary number. Second, and most important, why flipping the coord (to the other end of the mesh) only if any coord is larger than L/1.9? It makes no sense. We should just set coord[d]=L whenever coord[d]=0.

@mperego @ikalash Maybe I'm missing something here, but I don't know what it is. This block of code is very old (8 years), so the rationale of it may have gotten lost in the cracks.

If we agree on the mods, I will go ahead and modify also the regression tests values. I wanted to check with you guys before doing that.

### DIFF
--- a/src/disc/stk/Albany_STKDiscretization.cpp
+++ b/src/disc/stk/Albany_STKDiscretization.cpp
@@ -1321,41 +1321,33 @@ STKDiscretization::computeWorksetInfo()
         }
         for (std::size_t i = 0; i < buckets[b]->size(); i++) {
           int  nodes_per_element = buckets[b]->num_nodes(i);
-          bool anyXeqZero        = false;
-          for (int j = 0; j < nodes_per_element; j++)
-            if (m_ws_elem_coords[b][i][j][d] == 0.0) anyXeqZero = true;
-          if (anyXeqZero) {
-            bool flipZeroToScale = false;
-            for (int j = 0; j < nodes_per_element; j++)
-              if (m_ws_elem_coords[b][i][j][d] > stkMeshStruct->PBCStruct.scale[d] / 1.9)
-                flipZeroToScale = true;
-            if (flipZeroToScale) {
-              for (int j = 0; j < nodes_per_element; j++) {
-                if (m_ws_elem_coords[b][i][j][d] == 0.0) {
-                  double* xleak = new double[stkMeshStruct->numDim];
-                  for (int k = 0; k < stkMeshStruct->numDim; k++)
-                    if (k == d)
-                      xleak[d] = stkMeshStruct->PBCStruct.scale[d];
-                    else
-                      xleak[k] = m_ws_elem_coords[b][i][j][k];
-                  std::string transformType = stkMeshStruct->transformType;
-                  double      alpha         = stkMeshStruct->felixAlpha;
-                  alpha *= M_PI / 180.;  // convert alpha, read in from
-                                       // ParameterList, to radians
-                  if ((transformType == "ISMIP-HOM Test A" ||
-                       transformType == "ISMIP-HOM Test B" ||
-                       transformType == "ISMIP-HOM Test C" ||
-                       transformType == "ISMIP-HOM Test D") &&
-                      d == 0) {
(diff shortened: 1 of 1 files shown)


bartgol:
I changed the title, since I realized that this really applies to all tests that use a periodic mesh.

The snippet of code that makes no sense to me is [this](https://github.com/sandialabs/Albany/blob/1212161b4cc0b9d891b6f06c89db5d6d9e296bd9/src/disc/stk/Albany_STKDiscretization.cpp#L1330).

Pinging @mcarlson801 and @jewatkins in case they have thoughts on this.

jewatkins:
that bit of code looks even older: https://github.com/sandialabs/Albany/commit/50c61f304c31b698686171349addabcc41eebaf8#diff-e898f8522a0fea888075605242d52625c6c7c0b9eb3b22cd2bf1c27d2c466a23

what you have looks fine.

jewatkins (approved):

bartgol:
Ok, since Jerry already seems to agree, I'm going to start fix all the regression test values then. But I will not merge until I hear another thumbs up from the other ppl I pinged, to make sure we're not overseeing something.

bartgol:
For some context: I have a couple of PRs that restructure how we handle states, to allow more seamless switch to omegah (e.g., allow to use omegah device pointers for states). In the process, I changed this bit of code to edit the _nodal_ state, and that started breaking stuff. Digging, I realized that the current master modifies elem states, but the odd check on coord[d]>1.9 was skipping the update in some elements.

bartgol:
Btw, I am not sure how the tests passed before. I saw this for the ISMIP test B:
```
2: Response 1: Solution Max Value
2: 
2:                     2.372857585175e+00
2: Response Test 1: 2.372857585175e+00 != 0.000000000000e+00 (rel 1.000000000000e-06 abs 1.000000000000e-10)
2: 
2: Sensitivities (1,0):
2: 
2:                     -8.996073943430e-05
2: 
2: Response 2: Solution Average
2: 
2:                     4.862116055279e+00
2: Response Test 2: 4.862116055279e+00 != 3.649614402659e+00 (rel 1.000000000000e-06 abs 1.000000000000e-10)
```

So the 2nd response was the sol max value, and it was supposed to be 0. Yet, the solution average was supposed to be 3? You can't have avg=3 if your max is 0... This makes me think that one (or both) of those responses are buggy...

tests/landIce/FO_ISMIP/input_fo_ismip-hom_testB.yaml
```diff
@@ -142,15 +142,15 @@ ANONYMOUS:
   Regression For Response 0:
     Relative Tolerance: 1.0e-06
     Absolute Tolerance: 1.0e-06
-    Test Value: 1.067962127873e+01
+    Test Value: 1.618727420764e+01
   Regression For Response 1:
     Relative Tolerance: 1.0e-06
     Absolute Tolerance: 1.0e-10
-    Test Value: 0.0
+    Test Value: 2.372857585175e+00
```
jewatkins, line 149: this seems like a pretty big change but your code seems more general and I'm not really following what the original intention of the code was for the ismip tests. maybe @ikalash or @mperego could explain.

bartgol: @jewatkins does my argument about the origin of the big regression test diff seem reasonable? If so, I can merge this PR. Otherwise, we need to dig more.

PR #1130, opened 2025-08-21
Updates to handling of states

bartgol:
Besides some cleanup here and there, the main goal of this PR is to shift the responsibility of handling states from the discretization to the field accessor. The main reason is to make the ExtrudedMesh implementation easier (follow up PR will come).


bartgol:
Tests pass on my workstation.

To recap, the main change is to move the handling of states from the disc to the field accessor (aka field container in STK). This required a few changes here and there.

mperego (approved):
Looks good to me. Thanks!

Issue #1129, opened 2025-06-10
Failing tests starting May 30, 2025


ikalash:
The following tests began failing on May 30: 

![Image](https://github.com/user-attachments/assets/a619bea3-cae0-4d7c-b1bd-aec502fbb12d)

Could @mperego or @bartgol please have a look when you have time?  It looks like the BC syntax changed, leading to an error when starting the problems.

ikalash:
I forgot to put the CDash link: https://sems-cdash-son.sandia.gov/cdash//index.php?project=Albany&date=2025-06-10 .

bartgol:
I noticed this when debugging one of my branches. I think I may have a commit to fix this somewhere, I have to check.

jewatkins:
@bartgol the same 10 blake tests are still failing since May and that PR has been pushed. It looks like there's some slight differences in some tests while others have issues with sidesets:
```
 Throw test that evaluated to true: meshSpecs[0]->sideSetMeshSpecs.find(ssName)==meshSpecs[0]->sideSetMeshSpecs.end()
 
 Error! Either the side set name is wrong or something went wrong while building the side mesh specs.
```
any idea why that might have happened?

jewatkins:
It looks like it wasn't your commits. they passed on 5/30 with commit a2404972c0bafc4f48376a008105cd868749cc7d

jewatkins:
This trilinos commit passed: [a0c3c90a901](https://github.com/trilinos/Trilinos/commit/a0c3c90a9012ecc0cff2d87f593c48fce25ffc78)
This trilinos commit failed: [e7e7e2b34ce295bf5e5f6ada193e5b88482b7fa9](https://github.com/trilinos/Trilinos/commit/e7e7e2b34ce295bf5e5f6ada193e5b88482b7fa9)

and there was a seacas update as well as a few muelu updates inbetween. I don't think we really use muelu in those tests so probably the seacas update.

bartgol:
I think it's great that seacas now honors the names, so we don't need to think about "where is the name coming from". E.g., when creating a STK2D mesh, we use uppercase, but if we save and restart, the ss names would change, which is annoying.

Given the situation, I think we can simply use the original case in our yaml files, and maybe be done?

jewatkins:
confirmed that is indeed the issue. Going through and changing all the names now.

mperego:
Thanks @jewatkins !

PR #1132, opened 2025-09-05
Fix nodeset and sideset names to be consistent with what's in the mesh

jewatkins:
This fixes https://github.com/sandialabs/Albany/issues/1129 .


mperego (approved):
Looks good to me. Thanks!

PR #1146, opened 2025-09-23
More code cleanup in disc classes.

bartgol:
A few mods, to avoid code duplication and/or increase readability:

- create disc/extruded class for extruded disc data structures (they are, or soon will be, growing in number)
- move side set cell/node numeration maps to base class
- move creation of kokkos side-sets views from regular sideset structs to base class (can be used by all discs)
- allow recycling existing dof managers: dof managers really don't care about the fields names, so recyle an existing one if 2 dofs have the same fe type, order, dim and are on same mesh part.Must pass also the state structs (including side set ones)


mperego (commented):
It looks good to me. 
Don't you need to modify the `MpasSTKMeshStruct` as well?
Thanks!

PR #1147, opened 2025-09-29
Mods to remove STK knowledge from state saving evaluators

bartgol:
In order to achieve the goal, I needed a few mods:

- Add a side indexer in the disc: much like the cell indexer that is stored in the dof manager, it maps global<->local ids for mesh sides
- Minor change for the WsLID struct: it took me a while to realize that the "LID" member was not the element LID, but rather the "LID within the workset". So I changed the struct and members names.
- Put MeshFieldAccessor in charge of saving states to the mesh. This entailed:
  - Adding a "transferElemStateToNodeState" method (the "bacward" version of transferNodeStatesToElemStates, but just for ONE state)
  - Add knowledge of num worksets to the Workset struct (so that SaveStateField can transfer to elem->node once, when all worksets are done)
  - Make SaveStateField (and the sideset version) copy PHX fields into mesh states, without using any STK specific stuff.


bartgol:
Verified one more time on my workstation that tests pass. Merging.

PR #1148, opened 2025-09-29
Add Teko/MueLu preconditioner for ALI FO Thermo Coupled problem

jewatkins:
- add creation of preconditioner inside app using Teko
 - add nullspace creation for Teko/muelu
 - remove some unused preconditioner code from previous implementations
 - add wet bed test
 
Currently this is hard-coded for the ALI FO Thermo coupled problem but it should be easily extensible to other problems. It's enabled with `Use Physics-Based Preconditioner: true` and Teko parameter list (see test problem). It will also add the rotation mode to the nullspace for the velocity block with:
```
    LandIce Rigid Body Modes For Preconditioner:
      Compute Constant Modes: true
      Compute Rotation Modes: true
```

### DIFF
--- a/src/Albany_Application.cpp
+++ b/src/Albany_Application.cpp
@@ -28,6 +28,11 @@
 #include "Teuchos_TimeMonitor.hpp"
 #include "Zoltan2_TpetraCrsColorer.hpp"
 
+#ifdef ALBANY_TEKO
+#include "Teko_BlockedTpetraOperator.hpp"
+#include "Teko_InverseLibrary.hpp"
+#endif
+
 #include <stdexcept>
 #include <string>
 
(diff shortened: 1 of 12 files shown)


jewatkins:
Actually, I should probably add a check for the problem just to make sure it's not used for something else.

mperego (approved):
It looks good to me. Thanks for adding this capability to Albany.

jewatkins:
> @jewatkins when you have time (before or after merging the PR), would you mind adding a README file or some sort of docs explaining how the teko specs work? I saw the yaml file for the Thermo-coupled problem, and the list is quite long. I assume you gained a bit of expertise in the structure of the Teko options, so maybe you can give a reasonable overview of the params structure?

yeah I can do that. most of the parameters are muelu parameters, carried over from before my time. There might be some muelu documentation. The Teko parameters are what stratimikos/teko would normally use to build the preconditioner but I extract them and use them in albany instead. So it makes sense to document what I actually use and how they are used.

PR #1150, opened 2025-10-07
Rework the AdvDiff test folder

bartgol:
- Add 1d and 2d tests (no adaptation)
- Ensure every stk test has an omegah counterpart
- Simplify problem specs
- Adjust regression values

There are now 6 tests:

- 1d no adapt, with stk
- 1d no adapt, with omegah
- 1d yes adapt, with stk
- 1d yes adapt, with omegah
- 2d no adapt, with stk
- 2d no adapt with omegah

The 1d tests are bit-for-bit equivalent when run with stk or omegah. For the 2d tests, we have small roundoff diff, mostly due to how the two libraries order elements, which ultimately cause the order of assembly to be slightly different. However, the diffs are small, and below the usual tol for our tests (rel=1e-4, abs=1e-8). E.g.

```
Solution Max Value:
stk:    6.825290995201e-02
omegah: 6.825293131870e-02
```


cwsmith (approved):
👍

PR #1154, opened 2025-10-23
Fix omegah adapt test

bartgol:
This PR fixes how the OmegahMeshFieldAccessor handles the aliasing of albany states with omegah tags during adaptation.

Since we cannot (as of now, at least) get a nonconst version of a tag's array out of the omegah mesh, in order to ensure albany and omegah use the same arrays, we must re-create the writable array, copy data from the new (adapted) tag (if existing) into the newly created array, then re-set the tag in the mesh. Once that is done, we can proceed to recreate the albany states as we did before.

Note: this approach is a bit clunky, and requires a new allocation+deep_copy at run time. It would be nice if we could avoid that, but it may require a modification in Omega_h, in order to expose a non-const handle to the tag array. This may or may not happen. For the time being, we can bite the bullet and do the alloc+copy, and we should re-assess whether this is indeed a noticeable performance hit or if it's negligible in the overall timestep workflow.

### DIFF
--- a/src/disc/omegah/Albany_OmegahMeshFieldAccessor.cpp
+++ b/src/disc/omegah/Albany_OmegahMeshFieldAccessor.cpp
@@ -20,7 +20,9 @@ addFieldOnMesh (const std::string& name,
       "Error! Tag '" + name + "' is already defined on the mesh.\n");
   Omega_h::Write<ST> f(m_mesh->nents(entityDim)*numComps,name);
   m_mesh->add_tag<ST>(entityDim,name,numComps,f,false);
-  m_tags[name] = f;
+  m_tags[name].array = f;
+  m_tags[name].ncomps = numComps;
+  m_tags[name].ent_dim = entityDim;
 }
 
 void OmegahMeshFieldAccessor::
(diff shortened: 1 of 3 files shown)


src/disc/omegah/Albany_OmegahMeshFieldAccessor.cpp
```diff
@@ -362,4 +361,22 @@ saveVector (const Thyra_Vector&  field_vector,
...
+void OmegahMeshFieldAccessor::reset_mesh_tags ()
+{
+  for (auto& [name, tag_handle] : m_tags) {
+    auto& array = tag_handle.array;
+    int dim     = tag_handle.ent_dim;
+    int ncmp    = tag_handle.ncomps;
+    if (m_mesh->has_tag(dim,name)) {
+      auto tag = m_mesh->get_tag<ST>(dim,name);
+      array = Omega_h::Write<ST>(tag->array().size(),name);
+      Kokkos::deep_copy(array.view(),tag->array().view());
```
cwsmith: I think we could simplify this a bit with:

```
array = Omega_h::Write<ST>(tag->array().view());
```

Note, this and the following `set_tag` call should just increase the reference count on the original array/view (i.e., no copies).

PR #1156, opened 2025-11-03
Add version of GIS unstruct test that uses the logically extruded mesh/discretizartion

bartgol:
It took me a while, but now I can make all tests pass on my workstation (except the nullspace utils tests, which also fail on master).

Here's a summary of changes/additions:

- add a `ExtrudedMeshFieldAccessor` class: this class relies on the basal mesh to store/load fields. It is also in charge of extruding/interpolating basal fields
- extruded mesh upgrades: store all node/side sets, store same 2d mesh also for the upperside, delay setting num horiz entities in LayeredMeshNumbering till after construction.
- fix some bugs in extruded conn mgr
- minor upgrades to DualDynRankView
- add `create_conn_mgr` method to disc classes (allows extruded disc to be agnostic to basal disc type)
- make LoadSideSetStateField indep of stk, and use only MeshFieldAccessor info (this required a small fix to how SideSetSTKMeshStruct computes its workset size, more on this below)
- remove WS specs from FO_GIS input files (more on this below)

Regarding workset size, I think there are fragile assumptions in the code. If we have only one workset, then everything is good. But with 2+ ws, there can be situations in which two 3d elements belong to the same 3d workset but have their boundary sides belonging to two different worksets. At least when the side mesh is the SideSetSTKMesh. That's b/c when we create the side mesh, we have no control (that I can tell) on how the elements are stuffed into the 2d mesh buckets. This discrepancy can cause issues when loading/saving sideset states, as we mix up a bit 2d states and 3d worksets.

Another issue is that with extruded meshes we should not freely set the 3d workset size, but rather hard-code it to be the product `ws_size_basal*num_layers`. Again, this is to ensure the same data locality for 2d and 3d quantities when loading/saving states.

### DIFF
--- a/tests/landIce/FO_GIS/input_fo_gis_beta_smb.yaml
+++ b/tests/landIce/FO_GIS/input_fo_gis_beta_smb.yaml
@@ -70,7 +70,6 @@ ANONYMOUS:
     Number Of Time Derivatives: 0
     Method: STKExtruded
     Exodus Output File Name: gis_beta_smb.exo
-    Workset Size: 2500
     NumLayers: 5
     Use Glimmer Spacing: true
     Thickness Field Name: ice_thickness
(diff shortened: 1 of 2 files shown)


jewatkins:
one benefit multiple worksets does provide is less memory usage. i.e. if you want to take a hit on perf. because you don't have much memory (maybe very large fad sizes?) then it's useful there. I'm not sure if @mperego has utilized that recently with optimization work.

PR #1157, opened 2025-11-06
mesh adaptation: add MeshFields SPR support

cwsmith:
This PR adds MeshFields as a new TPL that provides a superconvergent patch recovery implementation to compute a size field to drive mesh adaptation.  A 2d adaptive version of the 2d transient stk test `tests/corePDEs/AdvDiff/input_2d_stk.yaml` was added to exercise the functionality.

CMake flags to enable UVM support in Omega_h is included in this PR and was tested on Perlmutter GPU.

### DIFF
--- a/src/disc/omegah/Albany_OmegahDiscretization.cpp
+++ b/src/disc/omegah/Albany_OmegahDiscretization.cpp
@@ -2,6 +2,7 @@
 #include "Albany_OmegahUtils.hpp"
 #include "Albany_StringUtils.hpp"
 #include "Albany_ThyraUtils.hpp"
+#include "Albany_KokkosTypes.hpp" // PHX::Device::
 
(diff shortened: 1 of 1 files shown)


src/disc/omegah/Albany_OmegahDiscretization.cpp
```diff
@@ -5,11 +5,59 @@
...
+
+#ifdef ALBANY_MESHFIELDS
+#include <KokkosController.hpp>
+#include <MeshField.hpp>
+#include <MeshField_SPR_ErrorEstimator.hpp>
+#endif
 
 #include <Panzer_IntrepidFieldPattern.hpp>
 
+//FIXME!
```
bartgol: What's the FIXME about? Meaning, what is wrong with the code below that we need to keep in mind?

PR #1158, opened 2025-11-12
Some cleanup of the STK classes, mostly field containers

bartgol:
I was trying to make the logically extruded mesh/disc save the solution into the basal mesh, and ran into issues regarding wrong layouts (as the basal disc already defines a solution, with layout matching the extents of the basal mesh).

In order to be more free to add the solution field in the basal mesh, I had to cleanup and rework a few things, and ended up cleaning a bit of other stuff in the STK stack. Below is a sumamry of the main changes.

- Remove mesh transformation params from the abstract stk mesh struct. They are read from the plist and just stored, and later only used in the STK disc when transforming the mesh. Hence, the disc can simply access them from the plist if needed.
- Similar change for exo and output params. Move them from the mesh struct (where they are just stored) to the disc (which is the only customer of these params).
- Remove a few other unused params in the abstract stk mesh struct.
- Do not call setFieldData on the side discs from the disc factory. Instead, let the disc do it. This will help the extruded disc to inject different logic regarding the basal disc solution field.
- Move common impl of Oridnary and Multi stk field containers to the Generic one
- Remove 2nd constructor of Multi/Ordinary/Generic stk field container. The user can _add_ solution field metadata at a later stage, if need be.
- To avoid adding coord/proc_rank fields twice to the metadata, make Ordinary/Multi stk containers add these geo fields at construction only if requested (via bool flag)
- In Generic STK mesh, only create ordinary stk container for fields. Since this container is NOT used for the solution, the Multi one is pointless.

Tests pass on my workstation (except the known fail of NullSpace utils).

### DIFF
--- a/src/disc/stk/Albany_STKDiscretization.cpp
+++ b/src/disc/stk/Albany_STKDiscretization.cpp
@@ -81,6 +81,8 @@ STKDiscretization::STKDiscretization(
       auto ss_disc_params = Teuchos::sublist(ss_discretizations_params,ss_name);
       ss_disc_params->set("Number Of Time Derivatives",discParams->get<int>("Number Of Time Derivatives"));
       ss_disc_params->set("Sensitivity Method",discParams->get<std::string>("Sensitivity Method","None"));
+      ss_disc_params->set("Response Function Index", discParams->get<int>("Response Function Index"));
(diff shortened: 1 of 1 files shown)


mperego (approved):
It looks good to me.
I tried to run a MALI problem and it works fine.

PR #1160, opened 2025-11-20
Start working on gh CI build for PR testing

bartgol:
Implement a workflow for testing PRs on GitHub free runners.

Looking at the workflow run, it appears testing took ~45min, which seems like a reasonable amount. Perhaps we can make it a tad faster if we allow 4 MPI ranks in the test phase (I had it set to 2), although the bulk of the time was spent in the build phase...

The CI uses the [albany:gcc-12-serial](https://github.com/orgs/sandialabs/packages/container/package/albany) container, which I built from a recipe in [this](https://github.com/bartgol/albany-containers) repo. I can later move the recipes repo into the sandialabs org.

### DIFF
--- a/.github/workflows/ghci-gcc-serial.yml
+++ b/.github/workflows/ghci-gcc-serial.yml
@@ -45,16 +45,16 @@ jobs:
         run: |
           mkdir -p ${GITHUB_WORKSPACE}/albany/build && cd ${GITHUB_WORKSPACE}/albany/build
           cmake \
-            -D CMAKE_BUILD_TYPE:STRING=RELEASE                              \
-            -D CMAKE_INSTALL_PREFIX:PATH=${GITHUB_WORKSPACE}/albany/install \
+            -D CMAKE_BUILD_TYPE:STRING=RELEASE                                \
+            -D CMAKE_INSTALL_PREFIX:PATH=${GITHUB_WORKSPACE}/albany/install   \
             \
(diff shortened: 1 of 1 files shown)


bartgol:
Most tests fail b/c we assume a pre-decomposed mesh is avaialble for the number of ranks used. On gh free runners, I set the max number of ranks to 2, but I could prob set it to 4 and let the tests pass. Still, it may be worth removing this pointless assumption in the test folder.

Other failures seem to be due to failed regression values comparisons, like `landIce_FO_ISMIP_TestD`:
```
2025-12-06T03:38:48.4911336Z Sensitivities (2,0):
2025-12-06T03:38:48.4911519Z 
2025-12-06T03:38:48.4911651Z                     -6.688832971773e-02
2025-12-06T03:38:48.4912330Z Sensitivity (2,0), Test 0,0: -6.688832971773e-02 != -6.689022448775e-02 (rel 1.000000000000e-06 abs 1.000000000000e-08)
2025-12-06T03:38:48.4912898Z 
2025-12-06T03:38:48.4913104Z Main_Solve: MeanValue of final solution 8.276731681373e+00
2025-12-06T03:38:48.4913425Z 
2025-12-06T03:38:48.4913573Z Number of Comparisons Attempted: 4
2025-12-06T03:38:48.4913968Z Number of Failed Comparisons: 1
```
which is also a bit concerning/annoying if it boils down to a dependence on the number of ranks.

PR #1161, opened 2025-12-02
omegah 2d adapt: add two process test

cwsmith:
This PR supports execution of the 2d omegah adapt test with multiple MPI ranks.  Computing the SPR error estimator in parallel (distributed/partitioned mesh) requires one layer of ghost elements.

A two process ctest was added with modified response tests in the corresponding yaml file to account for the difference in solution due to parallel execution; meshfields SPR is not deterministic.  The results were checked in paraview and look good.


bartgol (approved):

bartgol:
@cwsmith I'll let you merge this, in case you decide to add some last minute change. If not, merge away. :)

PR #1171, opened 2025-12-17
Collection of Albany updates and fixes

mperego:
This PR includes:

1.  Added capability to interpolate elem layered fields into QuadPoint fields:
This is particularly useful for accurately evaluating the temperature on quadrature points for the MOLHO model. While doing this I slightly changed the way we interpolate the temperature for (P0) elem fields so that it is more consistent with the P0 discretization. Updated regressions values of a few tests correspondingly.
2.  Updates to the QuadraticLinearOperator Response:
    - added/fixed some comments
    - fixed bug when computing quadratic response when A=D
    - handled the case when the field p is the solution field.
    - added some printouts enable by verbose flag
3.  PyAlbany fixes:
    - Fixed some issues in the CMake logic (thanks @bartgol)
    - Fixed a few issues with the interface code and parameter handling.
    - Fixed input file missing MueLU specs.
4. LandIce: added the Elliptic Mock problem
    - Elliptic Mock problem provides a scalar elliptic PDE that mimics the kind of nonlinearity present in ice-sheet flow models
    and has a reaction term that mimics the basal traction term in ice-flow models
    - The same problem class can optionally provide the linearization above model at given nominal parameter and solution fields.
5. LandIce: updates to MPAS interface
   - Changed some defaults regarding quadrature rules.


bartgol (approved):
Looks good

PR #1178, opened 2026-01-08
Fix for CellTopologyData build error on pm-gpu

mcarlson801:
Looks like with the change to c++20, we started seeing build errors like

```/.../repos/Albany/src/landIce/problems/LandIce_ColumnCouplingTest.hpp(143): error: expected a ")"```

with the offending line:

```
p->set<Teuchos::RCP<const CellTopologyData> >("Cell Topology",Teuchos::rcp(new CellTopologyData(meshSpecs.ctd)));
```

It seems there is an issue with the implicit copy of `CellTopologyData` not compiling with the updated compilers on perlmutter. This PR fixes this by setting `shards::CellTopologyData` in the parameter lists and uses that to access `CellTopologyData` instead. Confirmed that this fixes the perlmutter issues and doesn't introduce any new failing tests. Running tests on other machines now but I figured I'd post this to see if anyone has any feedback in the mean time.

### DIFF
--- a/src/landIce/evaluators/LandIce_GatherVerticallyContractedSolution_Def.hpp
+++ b/src/landIce/evaluators/LandIce_GatherVerticallyContractedSolution_Def.hpp
@@ -63,11 +63,11 @@ GatherVerticallyContractedSolution(const Teuchos::ParameterList& p,
 
   this->addEvaluatedField(contractedSol);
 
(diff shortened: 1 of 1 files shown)


jewatkins (approved):
thanks!

bartgol (approved):

PR #1189, opened 2026-04-06
Add support for landice tests using omegah for the basal mesh

bartgol:
Most fixes are related to owned-vs-ghosted entities in omegah.

I still can't make the omegah and stk tests produce similar results. I know I can't expect bit-for-bit agreement, due to different ordering of entities inside the mesh, but the responses are wildly different, suggesting that there is still some areas of the code where things are not done correctly. I'll as a friend to take a look once the CI tests fail.


bartgol:
Mr. Claudio is now helping me with this. See [this](https://github.com/sandialabs/Albany/tasks/b10a624a-2620-4d1a-b78a-f0bd5b1e29e4?author=bartgol) session.

PR #1191, opened 2026-04-07
Promote computeField to a generic disc util

bartgol:
`computeField` already rejects non-nodal and layered fields, but both the STK and Omegah `load_math_expr` branches extracted/allocated coordinates before reaching that check — causing a segfault (STK) or out-of-bounds access (Omegah) for those unsupported cases.

## Changes

- **STK (`Albany_GenericSTKMeshStruct.cpp`)**: Add early `TEUCHOS_TEST_FOR_EXCEPTION` guards for `!nodal` and `layered` before the `stk::mesh::field_data` coordinate extraction loop, preventing a null-dereference when entities are elements.
- **Omegah (`Albany_OmegahGenericMesh.cpp`)**: Add the same guards before coordinate view allocation/copy, preventing out-of-bounds indexing into `m_coords_d` when `num_ents` would be an element count rather than a node count.
- **Omegah**: Remove unused `layout` type alias in the same branch.

### DIFF
--- a/src/disc/Albany_DiscretizationUtils.cpp
+++ b/src/disc/Albany_DiscretizationUtils.cpp
@@ -20,6 +20,11 @@
 #include "Intrepid2_HGRAD_WEDGE_C1_FEM.hpp"
 #include "Intrepid2_HGRAD_WEDGE_C2_FEM.hpp"
 
+// Expression reading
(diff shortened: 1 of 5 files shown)


bartgol:
@copilot can you add fixes for the comments you had?

PR #1193, opened 2026-04-13
Add GIS_Unstruct test that uses Omega_h for the basal mesh

bartgol:
This is a do-over of #1189, but that PR went through a lot, including being chopped into pieces to be integrated one at a time. This will hopefully be a cleaner one.


bartgol:
With a big help from bots, I was able to get the test to work. However, it still fails if the input fields are read from the mesh (resulting in NaN's). I decided to merge this (if approved and if tests pass) and try to figure out the culprit later. For now, we have to deal with the issue of having to load data from an ascii file at runtime.

bartgol:
I'm going to merge this, so we can move on to the issues that a FOThickness problem brings up.

PR #1195, opened 2026-04-22
Add omegah version of Humboldt thickness-coupled test

bartgol:
Summary:

- add a version of the test that uses the logically extruded mesh and stk on the basal mesh
- add a version of the test that uses the logically extruded mesh and omegah on the basal mesh (todo)


bartgol:
@copilot please address the failure in the last round of CI

bartgol:
@copilot The test I added does not give the same result as the original test (which uses STKExtruded discretization). The two should yield the same numbers though. I need you to investigate the possible source of differences, and push a fix. Then, keep an eye on the CI job, so you can see if your fix worked, and iterate.

bartgol:
@copilot The test still fails, and now it's not just a matter of different numbers, but an actual exception. Please address the error, then wait for the CI to complete to see if you fixed it, or if you need to iterate more.

bartgol:
Nope. SIdeSetSTK only works if the 3d mesh is STK, which it isn't here. I suppose I will have to debug myself.

bartgol:
@mperego I can't make the STKExtruded and Extruded test for fo_humboldt give bfb the same result for the linear solver iterations, even if I use RILUK (like the FO_GIS unstruct test, which _does_ give bfb the same linear solver iterations). However, the NOX iterations and the response are _exactly_ the same. I think this is acceptable, and may be due to some internal reordering of something (maybe sides?). I tried to debug with copilot, and he eventually told me that there is likely an element LID difference in the two discretizations, which is not per se a bug. I am leaning on just being fine with it, but if you disagree, i may make a point to change internal storage and make them match, though it may prove hard or impossible (due to the fact that we need to subview stuff for the states).

mperego:
@bartgol I'm not concerned about bfb. What's the difference? Have you checked the sensitivities as well to see if they are close to each other?

bartgol:
That's a good idea. I tried to turn on `Compute Sensitivities`, but the test has no parameters, so I may just do that at a later time.

However, I tried to run both tests (stk3d and stk logically extruded) with Column ordering, and even the linear solvers iterations residuals are bfb. Which reminded me that the logically extruded is FORCED to use a column ordering for local ids, as elements in a column _must_ be contiguous in order to view the 3d states as "vector" 2d states, so that we can use the basal field container to store them.

And, as I said, when we use Layer ordering, the linear solver residuals are slightly diff (something like 

```
  Iter   1, [ 1] :    1.370037e-12

vs

  Iter   1, [ 1] :    1.479166e-12
```
but the NOX residuals as well as the NOX step size is the same (well, the first 4 digits at least, which is what NOX prints). So I am relatively confident that this is just a matter related to the local ordering of the elements, impacting the order in which residuals are accumulated..

mperego (approved):
Looks good to me!

PR #1196, opened 2026-04-30
Add checks when importing ascii files

mperego:
This prevents hard to debug se.g faults occurring when loading ascii files with the wrong format.


mperego:
Merging. It's just a simple fix.

PR #1199, opened 2026-05-11
Restore libparmetis-dev package, but use login SHELL to enable multiverse repos

bartgol:
Turns out the right thing to do for fixing the build-containers workflow was to ensure the multiverse repos were enabled.

I got puzzled when running the container manually i was able to install libparmetis-dev, while the podman build cmd couldn't. Turns out that it's b/c running manually we get a _login_ shell, where some apt setup cmd are run, enabling the multiverse repo, which is where libparmetis-dev has been moved (it wasn't removed completely).

I also switched our build cmd to use `--format docker`, so that it used docker metadata format rather than OCI; the latter does not recognize the SHELL cmd in the dockerfile, which allows to specify the type of shell to use during the build (in our case, login shell). I like setting SHELL b/c the result of the podman build should be _the same_ as you would get if you podman run the base image and manually run all the RUN cmd from inside the base image. With a non-login shell, the two can yield slightly different outcomes.

Once this is merged, I will run again the build-containers job. Once done, the gcc-12-serial image will be slightly different, as it will contain the SHELL metadata and will be have a docker-format manifesto. Anyone using it as base image (e.g., for the mali image) and building with podman will get the warning

```
WARN[0155] SHELL is not supported for OCI image format, [/bin/bash -l -c] will be ignored. Must use `docker` format 
```

To get rid of it, simply add `--format docker` to the podman build line. Anyone using Docker to build (instead of podman) will not get the warning in the first place...

Edit: I have manually triggered a build-containers [job](https://github.com/sandialabs/Albany/actions/runs/25685753681/job/75408914120) on this branch, so we can merge only if it indeed fixes the workflow...

### DIFF
--- a/.github/workflows/build-containers.yml
+++ b/.github/workflows/build-containers.yml
@@ -34,7 +34,7 @@ jobs:
       - name: Build image
         run: |
           cd albany/source/tools/containers/gcc-serial
-          podman build -f Dockerfile -t albany:gcc-12-serial
+          podman build --format docker -f Dockerfile -t albany:gcc-12-serial
       - name: Tag image
         run: |
           podman tag albany:gcc-12-serial ghcr.io/sandialabs/albany:gcc-12-serial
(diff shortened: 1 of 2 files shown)


bartgol:
This was not right. Gemini told me something wrong, so I'm working on a different fix. Will reopen.

PR #1200, opened 2026-05-11
Fix ubuntu 24.04 as the base image

bartgol:
Newer versions have issues with parmetis package(s).

---

Testing: https://github.com/sandialabs/Albany/actions/runs/25702366807/job/75465075368


bartgol:
This seems to have worked. Merging.

PR #1208, opened 2026-09-15
LandIce: Add "Transition" Effective Pressure Type

matthewhoffman:
Adds a new `"Effective Pressure Type": "Transition"` option to LandIce::BasalFrictionCoefficient, a drop-in alternative to "Hydrostatic Computed At Nodes".  The formula computes N per node as a near-ocean, ocean-connected value that transitions over a bounded length scale to a prescribed inland minimum floatation fraction, based on height above flotation, then interpolates to quadrature points via the basis functions exactly like "Hydrostatic Computed At Nodes" does.

Three parameters are now exposed via YAML:
- "Minimum Fraction Overburden Pressure" (reused from the existing Hydrostatic types)
- "Length Scale Factor" (reused from the existing Hydrostatic types; 
- "Transition Height Above Flotation" (new, km, typical value 0.025); the region over which effective pressure is calculated exactly using the hydrostatic assumption before the transition region begins.

Changes:
- LandIce_BasalFrictionCoefficient.hpp: add TRANSITION to the EFFECTIVE_PRESSURE_TYPE enum, add the transition_h_ocean member, and declare a computeTransitionEffectivePressure() helper.
- LandIce_BasalFrictionCoefficient_Def.hpp: parse the new YAML option and validate parameters, extend the dependent-field setup (bed topography/thickness/physical constants) to cover the new type, add the TRANSITION case to the per-node/QP evaluation switch, and implement computeTransitionEffectivePressure() following the reference Python implementation.

No existing effective pressure types are altered.


src/landIce/evaluators/LandIce_BasalFrictionCoefficient_Def.hpp
```diff
@@ -740,6 +771,62 @@ operator() (const BasalFrictionCoefficient_Tag&, const int& cell) const {
...
+  // Ocean-connected effective-pressure fraction (guard against division
+  // by zero at ice-free cells: N will end up 0 there regardless, since N
+  // is proportional to ice_term below).
+  MeshScalarT q_ocean = (ice_term > 0.0) ?
+      MeshScalarT(KU::max(1.0 - ocean_term / ice_term, 0.0)) : MeshScalarT(0.0);
+
+  // Prescribed inland effective-pressure fraction: N/Pice_inland =
+  // 1 - overburden_fraction (same convention as Albany's HYDROSTATIC(_AT_NODES)
+  // types, where "Minimum Fraction Overburden Pressure" is subtracted from
+  // a retained-overburden fraction of 1, not applied directly).
```
mperego, line 797: maybe we should change this.. it seems non-intuitive that (1- overburden_fraction) is actually the overburden fraction..

src/landIce/evaluators/LandIce_BasalFrictionCoefficient_Def.hpp
```diff
@@ -740,6 +771,62 @@ operator() (const BasalFrictionCoefficient_Tag&, const int& cell) const {
...
+
+  MeshScalarT q;
+  if (pressure_smoothing_length_scale == 0.0) {
+    // Length scale of 0 disables the smooth transition (step function).
+    q = (height_above_flotation <= transition_h_ocean) ? q_near_ocean : q_inland;
+  } else {
+    MeshScalarT distance_into_transition =
+        KU::max(height_above_flotation - transition_h_ocean, 0.0);
+    MeshScalarT transition_q = q_inland - (q_inland - q_start) *
+        std::exp(-std::log(2.0) * distance_into_transition / pressure_smoothing_length_scale);
```
mperego: why log(2.0)? One can tweak the length scale to determine the exponential decay. But if you prefer it in "base" 2 instead of e, it's fine with me.

src/landIce/evaluators/LandIce_BasalFrictionCoefficient_Def.hpp
```diff
@@ -232,19 +234,35 @@ BasalFrictionCoefficient (const Teuchos::ParameterList& p,
...
-      TEUCHOS_TEST_FOR_EXCEPTION(pressure_smoothing_length_scale <= 0.0, Teuchos::Exceptions::InvalidParameter,
+      bool length_scale_must_be_strictly_positive = (effectivePressure_type != EFFECTIVE_PRESSURE_TYPE::TRANSITION);
+      TEUCHOS_TEST_FOR_EXCEPTION(length_scale_must_be_strictly_positive && pressure_smoothing_length_scale <= 0.0, Teuchos::Exceptions::InvalidParameter,
         std::endl << "Error in LandIce::BasalFrictionCoefficient:  \"Length Scale Factor\" should be positive\n");
+      TEUCHOS_TEST_FOR_EXCEPTION(!length_scale_must_be_strictly_positive && pressure_smoothing_length_scale < 0.0, Teuchos::Exceptions::InvalidParameter,
+        std::endl << "Error in LandIce::BasalFrictionCoefficient:  \"Length Scale Factor\" should be non-negative (0 disables the smooth transition, using a step function instead) when Effective Pressure Type is Transition\n");
+    }
+
+    if(effectivePressure_type == EFFECTIVE_PRESSURE_TYPE::TRANSITION) {
+      transition_h_ocean = beta_list.get<double>("Transition Height Above Flotation", 0.025);
```
mperego: I prefer not to provide hidden default parameters. You could change this to 
transition_h_ocean = beta_list.get<double>("Transition Height Above Flotation")
which will throw if the parameter is not provided.

src/landIce/evaluators/LandIce_BasalFrictionCoefficient_Def.hpp
```diff
@@ -740,6 +771,62 @@ operator() (const BasalFrictionCoefficient_Tag&, const int& cell) const {
...
+  // Ocean-connected value at the end of the fixed near-ocean region.
+  MeshScalarT q_start = (ice_term > 0.0) ?
+      MeshScalarT(rho_w * transition_h_ocean / ice_term) : MeshScalarT(0.0);
+  // Guarantee the transition begins at or below the prescribed inland
+  // value, and keep the near-grounding-line branch bounded as well.
+  q_start = KU::min(q_start, q_inland);
+  MeshScalarT q_near_ocean = KU::min(q_ocean, q_inland);
+
+  MeshScalarT q;
+  if (pressure_smoothing_length_scale == 0.0) {
```
mperego: This would create numerical issues with the integration (a la GLP). I'd rather use a small length scale if I want a sharp transition.

mperego (changes requested):
It  looks good. Thanks! We would need to provide at least a regression test though. I can help you with that.

src/landIce/evaluators/LandIce_BasalFrictionCoefficient_Def.hpp
matthewhoffman: Yeah, I get confused if it's the effective ice pressure overburden fraction or the water pressure overburden fraction.  So are you suggesting we change it for both this new parameterization and Downs & Johnson?  I don't think we need to maintain any backward compatibility since we have never operationally used DJ-N.  Should we change the name for both?

src/landIce/evaluators/LandIce_BasalFrictionCoefficient_Def.hpp
matthewhoffman: No reason.  I'll change it to base e.

src/landIce/evaluators/LandIce_BasalFrictionCoefficient_Def.hpp
matthewhoffman: Sounds good - updated in next commit.

src/landIce/evaluators/LandIce_BasalFrictionCoefficient_Def.hpp
matthewhoffman: I agree that we don't need to support this.  I'll remove in next commit.

matthewhoffman:
@mperego, I pushed a commit fixing the final 3 comments.  I didn't address the first one yet until we decide how best to change it.

src/landIce/evaluators/LandIce_BasalFrictionCoefficient_Def.hpp
matthewhoffman: I think my preference would be to call it "Maximum flotation fraction" where flotation fraction means Pw/Pice.  From a subglacial hydrology perspective, that feels more intuitive to me.  But given this parameterization is to calculate effective pressure, if you think it makes more sense to write it in terms of effective pressure, I can be convinced!

mperego: > Yeah, I get confused if it's the effective ice pressure overburden fraction or the water pressure overburden fraction. So are you suggesting we change it for both this new parameterization and Downs & Johnson? I don't think we need to maintain any backward compatibility since we have never operationally used DJ-N. Should we change the name for both?

yes, I think it's best. we need to track where it is used in the tests and fix it there too.

src/landIce/evaluators/LandIce_BasalFrictionCoefficient_Def.hpp
mperego: sorry, you need to specify the type:
`transition_h_ocean = beta_list.get<double>("Transition Height Above Flotation");`

src/landIce/evaluators/LandIce_BasalFrictionCoefficient_Def.hpp
mperego: > I think my preference would be to call it "Maximum flotation fraction" where flotation fraction means Pw/Pice. From a subglacial hydrology perspective, that feels more intuitive to me. But given this parameterization is to calculate effective pressure, if you think it makes more sense to write it in terms of effective pressure, I can be convinced!

Fine with me.

matthewhoffman: @mperego , I made the change in 05c14855d.  Note that I realized my proposed "Maximum flotation fraction" doesn't really make sense because flotation fraction (Pw/Pice) near the GL is 1.0.  So I changed it to "Minimum flotation fraction".  If you prefer a different name (e.g. "Inland flotation fraction", "Flotation fraction limit"), feel free to suggest something different or just update the PR directly.

mperego (approved):
Thanks for the changes. 
I added a regression test, and fixed minor issue with validation of the parameter list.

PR #1210, opened 2026-09-16
Add optional thickness-dependent cutoff to Regularized Coulomb friction law

matthewhoffman:
This PR introduces an optional thickness transition to the existing Regularized Coulomb (RC) basal friction law, allowing thin ice to recover a velocity power-law response while preserving the standard RC behavior for thick ice.

Motivation:
In the effective-pressure-dependent regularized Coulomb friction law, thin grounded glaciers can be forced into the Coulomb-limited regime due to small absolute effective pressure, even when basal water pressure is a small fraction of overburden. This can produce unrealistically large velocities in steep, thin glaciers (e.g., small land-terminating glaciers on the Antarctic Peninsula) that must be balanced primarily by membrane stresses. The thickness cutoff allows basal traction to regain a velocity-power-law asymptote in thin ice.

Implementation:
The modification introduces a smooth, dimensionless thickness switch SH(H) that multiplies only the velocity term in the RC denominator:

```
  beta_RC = mu * N * |u|^(q-1) / [SH(H)*|u| + L*A*N^n]^q
```

For thick ice (H >= upper bound), SH=1 recovers the exact existing RC law. For thin ice (H <= lower bound), SH=0 removes the velocity term from the denominator, yielding the RC power-law asymptote where traction varies as |u|^q. A quintic smoothstep provides a smooth transition between bounds.

New YAML parameters (all optional, default off):
- Use Regularized Coulomb Thickness Cutoff: false
- Regularized Coulomb Thickness Lower Bound: 0.075  [km]
- Regularized Coulomb Thickness Upper Bound: 0.150  [km]

The feature is backward compatible (disabled by default), adds no overhead when disabled, and is limited to the REGULARIZED_COULOMB friction type. Field dependencies are registered conditionally to avoid unnecessary data access. The implementation follows existing Albany conventions for Phalanx evaluators, Kokkos kernels, and parameter validation.


src/landIce/evaluators/LandIce_BasalFrictionCoefficient_Def.hpp
```diff
@@ -172,6 +178,23 @@ BasalFrictionCoefficient (const Teuchos::ParameterList& p,
         std::endl << "Error in LandIce::BasalFrictionCoefficient:  \"" << flowRateType << "\" is not a valid parameter for Is not a valid parameter for Flow Rate Type\n");
     }
 
+    // Parse optional thickness cutoff for Regularized Coulomb
+    use_rc_thickness_cutoff = beta_list.get<bool>("Use Regularized Coulomb Thickness Cutoff", false);
+    if (use_rc_thickness_cutoff) {
+      rc_thickness_lower = beta_list.get<double>("Regularized Coulomb Thickness Lower Bound", 0.075);
```
mperego, line 184: I typically prefer avoiding hidden default values.

mperego (changes requested):
It looks good. We just need to add a regression test that uses this feature.

PR #1211, opened 2026-09-18
Fix missing break statements in transferElemStateToNodeState switch

matthewhoffman:
The switch(rank) block in transferElemStateToNodeState() was missing break statements after the case 3 (vector) and case 4 (tensor) bodies, causing execution to always fall through into the default case and throw std::runtime_error("Unsupported rank for state ..."). This meant ANY node-rank vector or tensor state saved via SaveSideSetStateField (e.g. a side-set 'Velocity' field with Field Usage: Output) would unconditionally crash, regardless of whether its rank was actually valid (2, 3, or 4). Scalar (rank 2) states were unaffected since that case already had a break.


mperego (approved):
Thanks @matthewhoffman !!

PR #1212, opened 2026-09-18
Few fixes

mperego:
- Fix depth-averaged velocity for MOLHO model, used when computing flux divergence terms
    When using the MOLHO model, the depth-averaged velocity is equal to 1/5 bed_velocity + 4/5 surface_velocity, not the average of the two, which would be correct for a linear velocity   
    Fixed regression value of FO_Thermo_Humboldt_depthInt_fluxDiv accordingly

- Disable unsafe option to use layer-averaged temperature for viscosity, when using MOLHO model
  Averaging the temperature in each element (piece-wise constant approximation) when computing the viscosity is used to match MPAS temperature discretization. However, MOLHO has only one layer and this approach results in using constant temperature in a column of ice, which is a bad approximation.
- fix scatter/gather of 2d field
  When doing a gather of scatter of the 2d field living at layers interfaces, we need to decide whether to use the layer above the field or below the filed to get 3d info. Here we make the choice, consistent across the evaluators, to pick the layer below the field, except when the field is at the bed, in which case we use the layer above it. While it is not necessary to use the same approach in each evaluator, it makes it easier to spot bugs. 
 Fixed a couple of related bugs.


bartgol (approved):
Looks good. I'm assuming "depth integrated" will _always_ mean "one layer with two 4th degree basis functions in the z direction"?

mperego:
> Looks good. I'm assuming "depth integrated" will _always_ mean "one layer with two 4th degree basis functions in the z direction"?

yes.. we should change the name at some point and call it MOLHO. Thanks!

mperego:
> Btw, what does "MOLHO" stand for?

MOno Layer Higher Order  model
https://tc.copernicus.org/articles/16/179/2022/

